PR #20, opened 2016-03-04
Fixed quiet mode

SethDusek:
I couldn't seem to figure out why my images were coming out as corrupted even though they were fine. Turns out oxipng was still printing information about the image such as the reduction even though it was in --quiet mode and --stdout, and that broke the image.

Anyways this does some additional checking for opts.verbosity before it prints stuff

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,7 +45,7 @@ pub struct Options {
 
 pub fn optimize(filepath: &Path, opts: &Options) -> Result<(), String> {
     // Decode PNG from file
-    println!("Processing: {}", filepath.to_str().unwrap());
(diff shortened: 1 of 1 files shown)


shssoichiro:
Thanks for the pull request! I'll get this merged right away.

shssoichiro:
Do you think it would be necessary/useful to enable `quiet` mode automatically whenever `stdout` is selected?

SethDusek:
I think so, yes, since most people would probably use -q so the only thing in stdout is the image itself and nothing more so they don't have to parse it and remove the text. So I think maybe you should enable quiet mode when stdout is also enabled

CounterPillow:
Alternatively the program could output status messages to stderr instead of stdout, like curl does.

shssoichiro:
This is true. Maybe that would be a better option.

PR #25, opened 2016-03-05
Update README.md

corppneq:
> According to http://optipng.sourceforge.net/history.txt "Version 0.7.5 from 24-mar-2014" is the latest optipng version.

https://www.reddit.com/r/linux/comments/491uf5/oxipng_png_optimizer_written_in_rust/

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ More advanced options can be found by running `oxipng -h`.
 ## History
 
 Oxipng began as a completely rewrite of the OptiPNG project,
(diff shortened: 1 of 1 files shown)


shssoichiro:
Interesting. The latest Hg commit is from 2013-12-29. I had assumed it to be newer than 0.7.5, and the history.txt implies such as well. It looks like 0.7.5 was released after the latest changes to Hg, without incorporating the Hg changes.

Either way, it seems March 2014 is more accurate.

PR #40, opened 2016-05-04
Add support for optimizing in-memory PNGs.

tobz:
In keeping with being able to use oxipng as a library, there should be a way to alter an image already in-memory.  This allows end users working with in-memory representations -- say, external web servers providing optimization services -- to easily take advantage of oxipng.


tobz:
So, this adds two new public functions, one to `lib` and one to `PngData`, which both allow being loaded from an existing buffer.

Things I'd love deep thoughts on:
- `[u8]` vs `Vec<u8>`? (still a Rust newbie, can't extol the differences, honestly)
- naming of functions

I still need to write a test for operating in in-memory mode.

shssoichiro:
Thanks for the pull request!

I think this is a great addition, and I think the function naming is fine. As for the `Vec` question, it is generally better for a function to take `&[u8]` as an argument rather than a `Vec<u8>`. There are several reasons why, but you can research Rust's concepts of ownership and borrowing to delve a bit deeper. The function can return a `Vec<u8>` with the new data.

This will also need to bump the minimum Rust version, which I'm fine with doing, since `as_slice` wasn't stable in 1.6.0. You can just change the minimum version from 1.6.0 to 1.7.0 in `.travis.yml` and `README.md`.

shssoichiro:
Actually, we can probably avoid bumping the minimum version by changing `byte_data.as_slice()` to `&byte_data`. These should be equivalent if the

tobz:
Sweet!  Also, you beat me to it: I ended up doing what you described and just switched `&byte_data`, so 1.6 should work. :)

So yeah, with the `[u8]` vs `Vec<u8>`, it seems like returning `[u8]` or `&[u8]` isn't trivially possible anyways? (issues with not implementing `Sized` and lack of lifetime modifier) so I stuck with returning `Vec<u8>`.  I'll switch to at least taking in `&[u8]` for the input.

I also added simple tests for both of the public API methods -- `optimize` and `optimize_from_memory` -- that should serve as a basic guarantee that the main entry points don't change out from under people, and that my changes actually work.

shssoichiro:
Yes, although it's possible to return a `&[u8]`, which can be more performant in some cases, it's usually more effort than it's worth.

src/lib.rs
```diff
@@ -134,21 +134,148 @@ impl Default for Options {
...
+    let file_original_size = data.len() as usize;
+    let mut png = match png::PngData::from_slice(&data, opts.fix_errors) {
+        Ok(x) => x,
+        Err(x) => return Err(x),
+    };
+
+    // Run the optimizer on the decoded PNG.
+    let optimized_output = optimize_png(&mut png, file_original_size, opts);
+
+    if file_original_size <= optimized_output.len() && !opts.force && opts.interlace.is_none() {
```
shssoichiro: This looks like a duplicate of the block at line 151.

shssoichiro:
I left one nit. Everything else looks fine. If you fix that, then I'll merge this after the CI tests complete.

Thanks!

tobz:
Schweet.  Do you prefer I squash the commits, or will you do it at merge time? Or do you even bother squashing commits? :)

shssoichiro:
You can squash if you feel comfortable with it, otherwise it's fine leaving it as is. I don't have a strong preference. :)

I also realize that the duplicate block was actually fine as-is. I thought it would print out the message twice, but `optimize` and `optimize_from_memory` are actually separate functions. It's been one of those days. But it's not bad having that `if` check DRYed out, so there's no need to revert that change.

tobz:
Ah, right on WRT the duplicate block.

All squashed, ready to merge once the tests pass. :D

shssoichiro:
Awesome. Thanks!

tobz:
Wooo!  No problem. :D

PR #67, opened 2017-06-18
Make the error module public

SethDusek:
Previously the error module and the PngError type were not public, which made users of your library unable to export it. This change makes the error module public, and reexports PngError


CounterPillow:
you might want to cherry-pick that commit to a new branch instead of sending in a PR with the random stuff in your master branch

SethDusek:
My bad, there's some stuff from about a year back. Creating a new one

PR #86, opened 2017-09-12
handle_optimization: skip files that do not end in .png

matthiaskrgr:
This should help avoiding a lot of unneccessary IO.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -258,6 +258,20 @@ fn handle_optimization(inputs: Vec<PathBuf>, opts: &Options) {
                 eprintln!("{} is a directory, skipping", input.display());
             }
             continue;
+        } else { // input is a file
+            let ext = input.extension();
+            match ext {
+                Some(ext) => {
+                    if ext != "png" { // not .png
+                        eprintln!("{} is not .png, skipping", input.display());
+                        continue;
+                    }
+                }
+                _ => { // no file extension
+                    eprintln!("{} is not .png, no file extension, skipping", input.display());
+                    continue;
+                }
+            }
         }
         if let Some(ref out_dir) = current_opts.out_dir {
             current_opts.out_file = out_dir.join(input.file_name().unwrap());


CounterPillow:
This is dumb. It should check the magic number instead of the file extension, it's perfectly valid to name your files whatever you want.

matthiaskrgr:
Would making the Overviewer write to ````.tmp.png```` avoid issues?

CounterPillow:
Not merging this garbage already avoids the issue.

matthiaskrgr:
Well, feel free to provide a better solution in another pull request.

CounterPillow:
I already provided a better solution for you: Don't feed files that aren't PNGs to oxipng. That is literally all you have to do. There is no PR to make here. Depending on file extensions is a distinctly Windows-esque thing to do.

It is worth noting that neither optipng nor zopflipng do this either.

Argon-:
I fail to get the problem in the first place. The usage example calls oxipng with `*.png` and that's obviously the right solution for only processing files ending in .png?

agrif:
For a concrete example of when it is possible, in a totally reasonable workflow, to have a PNG without any extension, see the [Apple guide on icon file names](https://developer.apple.com/library/content/qa/qa1686/_index.html). In particular, `iTunesArtwork` and `iTunesArtwork@2x`, which are noted:

> **Important:** iTunesArtwork icon images should be in png format, but name them without the .png extension.

If you wish to only process files ending in `.png`, you can use a `*.png` glob on the command line, or your favorite combination of `find` and `xargs`.

matthiaskrgr:
My use case is ````oxipng -r <dir with subdirs containing images>````.
Lets say you have a git repo of a game with source files and sounds and images, it will be sufficient to just ````oxipng -r .```` in the repo root instead of ```` find . | grep "\.png$" | xargs oxipng```` which will still fail if one of the filenames happens to contain a whitespace.

Also note that ````*.png```` only works for images directly in the current directory unfortunately.

CounterPillow:
So you want to break existing workflows because it saves you some typing?

CounterPillow:
Also pro tip: `find . -name '*.png' -exec oxipng {} \;`

Argon-:
I'm sorry but there's no reason to _hard code_ something like file suffixes (which are, as repeatedly stated, totally optional and arbitrary) and all just because your shell fu is not high enough.

And of course `*.png` only works for files in the current directory, that's exactly what it's supposed to do. When you want to include other directories then there are other globbings/patters available in every decent shell. Aliases and functions provide for additional comfort.

shssoichiro:
After reading through the discussion, I think I agree that oxipng shouldn't restrict processing based on file extension. I do understand the issue, which is that oxipng loads each file entirely before determining if it is a PNG file or not. I would be willing to accept a PR that optimizes this to only load enough bytes to read the header before either returning an `Err` or continuing to load the rest of the file. In the meantime, the other users have provided workarounds for the issue that hopefully should work for you.

CounterPillow:
I believe that reading just a few bytes would not be a sensible optimisation anyway, since I'd assume the OS reads the entire block no matter how much you read.

shssoichiro:
It depends on the size of the file and probably the filesystem as well. Regardless of filesystem though, the file would need to be 10+ MB before there was any noticeable difference.

Issue #85, opened 2017-09-12
don't load every file completely


matthiaskrgr:
It looks like oxipng reads the entire file in order to determine if it is a png or not.
When being lazy and optimizing a git repo via  ````oxipng . -r````, oxipng apparently also reads all source files or object files and even the .git repo pack into ram  which can mean several gigabyte of useless IO.

Maybe we can read only the header of a file?

PR #87, opened 2017-09-26
Check header of file before loading it into memory

emielbeinema:
This pull request seeks to provide a fix to #85 by checking only the first bytes of the file for the PNG header before loading the entire file into memory.
As this is my first PR for a Rust project, I'd be delighted to receive some feedback.

### DIFF
--- a/src/png.rs
+++ b/src/png.rs
@@ -11,7 +11,7 @@ use itertools::Itertools;
 use reduction::*;
 use std::collections::{HashMap, HashSet};
 use std::fs::File;
-use std::io::Read;
+use std::io::{Read, Seek, SeekFrom};
 use std::iter::Iterator;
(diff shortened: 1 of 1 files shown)


shssoichiro:
Thanks! This looks good. I just left one comment that should be easy to fix.

emielbeinema:
Thanks for the feedback! I can't seem to find the comment you mention, though. Have I overlooked something?

shssoichiro:
Sorry, I guess I had to hit "Submit Review". I'm not use to Github's way of doing PR comments, so that was probably my fault.

emielbeinema:
Thanks again! I followed your suggestion and it does look cleaner :)

shssoichiro:
Thanks!

Issue #83, opened 2017-09-11
crash: index 54 out of range for slice of length 48


matthiaskrgr:
file: https://github.com/FLIF-hub/benchmarks/blob/master/test_animations/random-gifs/SteamEngine.png

````
Processing: ./test_animations/random-gifs/SteamEngine.png
    320x240 pixels, PNG format
    8 bits/pixel, 16 colors in palette
    IDAT size = 2761 bytes
    File size = 38713 bytes
Reducing image to 4 bits/pixel, 15 colors in palette
Trying: 8 combinations
Found better combination:
    zc = 9  zm = 9  zs = 0  f = 0        2457 bytes
    IDAT size = 2457 bytes (304 bytes decrease)
    file size = 3868 bytes (34845 bytes = 90.01% decrease)
thread 'main' panicked at 'index 54 out of range for slice of length 48', /checkout/src/libcore/slice/mod.rs:748:4
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:60
             at /checkout/src/libstd/panicking.rs:381
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:397
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:577
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:538
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:522
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:498
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:71
   9: core::slice::slice_index_len_fail
             at /checkout/src/libcore/slice/mod.rs:748
  10: png::decoder::expand_paletted
             at /checkout/src/libcore/slice/mod.rs:879
             at /checkout/src/libcore/slice/mod.rs:730
             at /checkout/src/liballoc/vec.rs:1575
             at /home/matthias/.cargo/registry/src/github.com-1ecc6299db9ec823/png-0.9.0/src/decoder/mod.rs:455
[...]

Issue #79, opened 2017-08-07
APNG not supported


javiergutierrezchamorro:
With old OptiPNG I could use -protect acTL,fcTL,fdAT to keep APNG animations. But seems it is not possible with OxiPNG.

Find attached example.

Will you support it?

![animated_png_example_bouncing_beach_ball](https://user-images.githubusercontent.com/4672971/29039372-58f088f0-7bab-11e7-9afc-5de33a4ded54.png)

shssoichiro:
I do plan to implement support for APNG files in Oxipng, however at the moment it is low priority so I can't give a timeframe on when it will be implemented.

javiergutierrezchamorro:
Thanks.
Once implemented, I will include it in FileOptimizer replacing OptiPNG:
http://nikkhokkho.sourceforge.net/static.php?page=FileOptimizer

TPS:
_N.B.: 1st work on this is @ https://github.com/shssoichiro/oxipng/commit/db0131c47e09754777bd642313a0e37a5a910ca3_

shssoichiro:
I just wanted to reassure everyone that I am still working on this issue. I have had less time lately to focus on open source, but I have not abandoned this project.

javiergutierrezchamorro:
Glasd to hear @shssoichiro

shssoichiro:
Unfortunately this is soft-blocked on https://github.com/PistonDevelopers/image/issues/195. The code in oxipng can still be written, but there would be no sanity check on optimized images to make sure oxipng hasn't corrupted them, and no automated tests. I don't feel comfortable delivering this functionality until its correctness can be verified.

AlexApps99:
APNG support was added in image-rs/image#1250, so it's now possible to resume development of this feature.

javiergutierrezchamorro:
Just tried with latest 5.0.0 and reports as not supported yet.

![image](https://user-images.githubusercontent.com/4672971/131451924-b8a5234e-5229-4e46-9035-ee2ffc0dffe4.png)

javiergutierrezchamorro:
Any progress on this?

TPS:
@shssoichiro, @andrews05:  Sorry for tagging y'all directly 🙇🏾‍♂️, but minimal support seems so close.  If y'all could spare a few minutes & let us know whether this is a reasonable idea, we'd appreciate it.:

[...]

PR #88, opened 2017-09-30
Return an error when an APNG header is encountered

emielbeinema:
This should "fix" #83 until support for APNG is implemented (#79).
When we encounter the acTL-header, we stop processing the file, to prevent errors to occur.

The image for the test was taken from https://commons.wikimedia.org/wiki/File:Animated_PNG_example_bouncing_beach_ball.png and is in the public domain.


shssoichiro:
Looks great! Thanks for this contribution!

PR #103, opened 2018-06-01
Option to use Cloudflare's zlib for compression

kornelski:
There's a zlib fork optimized for x86-64 with SSE 4.2. For most workloads it's 2-3× faster than miniz, and it improves compression by ~1%. This PR adds `--features=cfzlib` to use it.

```diff
  name                        miniz ns/iter  cfzlib ns/iter  diff ns/iter   diff %  speedup
+ deflate_16_bits_strategy_0  45,920,423     20,495,977       -25,424,446  -55.37%   x 2.24
+ deflate_16_bits_strategy_1  44,222,845     18,983,512       -25,239,333  -57.07%   x 2.33
+ deflate_16_bits_strategy_2  19,427,583     8,535,854        -10,891,729  -56.06%   x 2.28
+ deflate_16_bits_strategy_3  24,503,836     9,132,294        -15,371,542  -62.73%   x 2.68
- deflate_1_bits_strategy_0   85,105         115,958               30,853   36.25%   x 0.73
- deflate_1_bits_strategy_1   87,566         118,008               30,442   34.76%   x 0.74
+ deflate_1_bits_strategy_2   395,498        119,599             -275,899  -69.76%   x 3.31
+ deflate_1_bits_strategy_3   79,729         21,454               -58,275  -73.09%   x 3.72
- deflate_2_bits_strategy_0   226,508        215,926              -10,582   -4.67%   x 1.05
- deflate_2_bits_strategy_1   170,883        215,592               44,709   26.16%   x 0.79
+ deflate_2_bits_strategy_2   765,287        235,921             -529,366  -69.17%   x 3.24
+ deflate_2_bits_strategy_3   139,843        43,825               -96,018  -68.66%   x 3.19
- deflate_4_bits_strategy_0   346,245        642,137              295,892   85.46%   x 0.54
- deflate_4_bits_strategy_1   337,300        732,665              395,365  117.21%   x 0.46
+ deflate_4_bits_strategy_2   1,435,181      475,612             -959,569  -66.86%   x 3.02
+ deflate_4_bits_strategy_3   264,397        82,781              -181,616  -68.69%   x 3.19
+ deflate_8_bits_strategy_0   31,314,571     14,158,232       -17,156,339  -54.79%   x 2.21
+ deflate_8_bits_strategy_1   35,105,558     14,265,609       -20,839,949  -59.36%   x 2.46
+ deflate_8_bits_strategy_2   9,819,240      4,303,855         -5,515,385  -56.17%   x 2.28
[...]

PR #104, opened 2018-06-01
Bail early

kornelski:
Added on top of #103

* Abort deflate early when compressed size exceeds best known size so far. This reduces CPU usage by 10-15%
* Run alpha reductions in parallel, because why not?
* Error struct has changed, so that's technically a breaking change

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ matrix:
       env: TARGET=x86_64-apple-darwin
       cache: cargo
     - os: linux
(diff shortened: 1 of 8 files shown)


kornelski:
I've rebased it

shssoichiro:
This looks good to me. I'm not thrilled about the breaking change, but I agree the enum is a better way of handling the errors. I might take it a step further and implement `failure` at some point before releasing 2.0.

Thanks!

PR #105, opened 2018-06-01
allow listing more than 2 filters (issue 101)

magicgoose:
https://github.com/shssoichiro/oxipng/issues/101

The only function that's edited is `parse_numeric_range_opts`.

The `regex` crate is not used anymore in main.rs, so if it's not used in other places, it probably can be dropped from dependencies. (It was only used in `parse_numeric_range_opts` and it seems it didn't really simplify the code that much…)  
Also this is my first pull request with Rust code ever, so I could write something silly, please check carefully.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,7 +10,6 @@
 extern crate clap;
 extern crate glob;
 extern crate oxipng;
-extern crate regex;
 
 use clap::{App, Arg, ArgMatches};
 use glob::glob;
@@ -18,7 +17,6 @@ use oxipng::AlphaOptim;
 use oxipng::Deflaters;
 use oxipng::Headers;
 use oxipng::{Options, PngError};
-use regex::Regex;
 use std::collections::HashSet;
 use std::fs::DirBuilder;
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


magicgoose:
Hmm actually it also affects `compression` and `strategies` parameters, but I guess that's a good thing.

src/main.rs
```diff
@@ -441,40 +439,41 @@ fn parse_numeric_range_opts(
...
-        ).as_ref(),
-    ).unwrap();
+    let error_message = "Not a valid input";
     let mut items = HashSet::new();
 
-    if one_item.is_match(input) {
-        items.insert(input.parse::<u8>().unwrap());
-        return Ok(items);
+    // one value
+    if let Some(one_value) = input.parse::<u8>().ok() {
```
shssoichiro: You can write this more efficiently as `if let Ok(one_value) = input.parse::<u8>()`

src/main.rs
```diff
@@ -441,40 +439,41 @@ fn parse_numeric_range_opts(
...
         }
+    }
 
-        match &captures[2] {
-            "," => {
-                items.insert(first);
-                items.insert(second);
+    // a list ("A,B[,…]")
+    for value in input.split(',') {
+        if let Some(value_int) = value.parse::<u8>().ok() {
```
shssoichiro: Same as above (`Ok(value_int)` to avoid the `.ok()` call)

src/main.rs
```diff
@@ -441,40 +439,41 @@ fn parse_numeric_range_opts(
...
     max_value: u8,
 ) -> Result<HashSet<u8>, String> {
-    let one_item = Regex::new(format!(r"^[{}-{}]$", min_value, max_value).as_ref()).unwrap();
-    let multiple_items = Regex::new(
-        format!(
-            r"^([{}-{}])(,|-)([{}-{}])$",
-            min_value, max_value, min_value, max_value
-        ).as_ref(),
-    ).unwrap();
+    let error_message = "Not a valid input";
```
shssoichiro: This can be a `const`.

src/main.rs
```diff
@@ -441,40 +439,41 @@ fn parse_numeric_range_opts(
...
+        let second_opt = range_values[1].parse::<u8>().ok();
+        if let (Some(first), Some(second)) = (first_opt, second_opt) {
+            if min_value <= first && first < second && second <= max_value {
+                for i in first..second + 1 {
+                    items.insert(i);
+                }
+                return Ok(items);
+            }
         }
+    }
```
shssoichiro: This will silently fail and use the default option if someone puts in e.g. `1-` or `1-2-3` or `1-A`.

src/main.rs
```diff
@@ -441,40 +439,41 @@ fn parse_numeric_range_opts(
...
-        let first = captures[1].parse::<u8>().unwrap();
-        let second = captures[3].parse::<u8>().unwrap();
-        if first >= second {
-            return Err("Not a valid input".to_owned());
+    // a range ("A-B")
+    let range_values = input.split('-').collect::<Vec<&str>>();
+    if range_values.len() == 2 {
+        let first_opt = range_values[0].parse::<u8>().ok();
+        let second_opt = range_values[1].parse::<u8>().ok();
+        if let (Some(first), Some(second)) = (first_opt, second_opt) {
```
shssoichiro: Use `Ok(first), Ok(second)` here as well.

magicgoose:
@shssoichiro thanks for the review, I made the changes now.

src/main.rs
```diff
@@ -10,15 +10,13 @@
 extern crate clap;
 extern crate glob;
 extern crate oxipng;
-extern crate regex;
```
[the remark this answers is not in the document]
magicgoose: GitHub still shows this change as not "outdated" yet, unlike others, and I'm not sure if this is normal, but I removed `regex` from the `Cargo.toml` file too.

shssoichiro:
Thanks!

PR #110, opened 2018-07-05
Enum for stdout writing flag

kornelski:
* It's more idiomatic to have one enum than two/three inter-dependent struct fields.
* Backup is not performed when writing to stdout, since it's not necessary. In an unfortunate case of a bash footgun `oxipng foo.png --stdout > foo.png` it wouldn't be able to help anyway.
* Changed `perform_backup()` to use `output_path` rather than `input path`.
* Added cause to error descriptions.
* Removed `out_dir` and clone+mutate workarounds from `Options` in favor of separate, explicit arguments. This makes options purely immutable configuration shared for all files.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -45,22 +45,28 @@ mod reduction;
 mod atomicmin;
 
 #[derive(Clone, Debug)]
+pub enum OutFile {
+    /// Path(None) means same as input
+    Path(Option<PathBuf>),
+    StdOut,
+}
+
+impl OutFile {
+    pub fn path(&self) -> Option<&Path> {
+        match *self {
+            OutFile::Path(Some(ref p)) => Some(p.as_path()),
+            _ => None,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
 /// Options controlling the output of the `optimize` function
 pub struct Options {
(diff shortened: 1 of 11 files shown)


kornelski:
There's no support for reading from stdin, so I'm planning to add a similar enum for stdin.

src/lib.rs
```diff
@@ -289,65 +292,52 @@ pub fn optimize(input_path: &Path, opts: &Options) -> Result<(), PngError> {
...
+                    "Unable to write to file {}: {}", output_path.display(), err
+                )))?;
             if opts.preserve_attrs {
                 copy_permissions(input_path, &out_file, opts.verbosity);
             }
-
             let mut buffer = BufWriter::new(out_file);
-            match buffer.write_all(&optimized_output) {
-                Ok(_) => if opts.verbosity.is_some() {
-                    eprintln!("Output: {}", output_path.display());
```
shssoichiro: It looks like this log line is being removed, when it probably shouldn't be.

kornelski:
I've bumped it to 1.24

kornelski:
I've reinstanted the output line and rebased.

shssoichiro:
Thanks!

PR #111, opened 2018-07-05
Avoid double glob processing on unix

kornelski:
On unix-like platforms globs are expanded for all arguments before the executable is called, so if user runs:

```sh
oxipng foo*
```

the shell actually runs expanded args:

```rust
["oxipng", "foobar", "foobaz"]
```

which makes processing of glob paths inside the program unnecessary. Also, if there was a file called `foo*` (it's a legal name in Linux), unix tools wouldn't know how to specify just that file, since normal quoting wouldn't be enough:

```sh
oxipng "foo*"
```

because the un-expanded value passed to the program would still be expanded internally. So on unix doing nothing is the best way to support globs :)

Windows is different in that regard. It doesn't expand globs in the shell, and requires programs to handle globs internally. So to get uniform cross-platform support I've used the [wild](https://crates.rs/crates/wild)card crate which adds glob arguments on Windows.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -363,12 +363,12 @@ dependencies = [
  "clippy 0.0.195 (registry+https://github.com/rust-lang/crates.io-index)",
  "cloudflare-zlib-sys 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "crc 1.8.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


TPS:
@kornelski Should the rest of the references to `glob` be stripped, also, or are they somehow required by `wild` to be left as-is in `oxipng`?

kornelski:
@TPS I've removed the unnecessary entry from `Cargo.toml`. The `glob` crate is now conditionally included by the `wild` crate.

shssoichiro:
Cool, thanks. This even looks cleaner.

PR #112, opened 2018-07-05
Make Rayon an optional dependency, because WASM

myfreeweb:
On WebAssembly, Rayon does not work: the thread pool fails to initialize since threads are stubbed out.

With this and some changes to `miniz_oxide` (for now, described in https://github.com/Frommi/miniz_oxide/issues/26), I've been able to run oxipng in the browser :)

wayland-screenshot-2018-07-05_16-33-27-fs8


shssoichiro:
Thanks! Looks good, I'll go ahead and merge.

PR #115, opened 2018-07-10
Limit wall clock time spent on optimization trials

kornelski:
For automated optimization tools it's problematic when large images take very long time to optimize, so they have to have some time limit. 

Simply killing the process when it takes too long is wasteful. It's better to keep existing work done and exit gracefully.

This PR adds a coarse-grained limit for maximum time spent on trying potential optimizations. For simplicity, it doesn't try to finish exactly within the given timeout, merely stops adding new work after a deadline.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -22,6 +22,8 @@ use std::collections::{HashMap, HashSet};
 use std::fs::{copy, File};
 use std::io::{stdout, BufWriter, Write};
 use std::path::{Path, PathBuf};
+use std::time::{Duration, Instant};
+use std::sync::Mutex;
 use atomicmin::AtomicMin;
 
 pub use colors::AlphaOptim;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -158,6 +160,10 @@ pub struct Options {
     ///
     /// Default: 1.5x CPU cores, rounded down
     pub threads: usize,
+
+    /// Maximum amount of time to spend on optimizations.
+    /// Further potential optimizations are skiped if the timeout is exceeded.
```
shssoichiro: Typo in "skipped".

src/lib.rs
```diff
@@ -675,11 +699,50 @@ fn perform_reductions(png: &mut png::PngData, opts: &Options) -> bool {
...
 
     reduction_occurred
 }
 
+
+/// Keep track of processing timeout
+struct Deadline {
+    start: Instant,
+    timeout: Option<Duration>,
+    print_message: Mutex<bool>,
```
shssoichiro: Could this be accomplished with an AtomicBool?

kornelski: Probably, but the implementation would be less obvious. This path is very cold, so it wouldn't make any difference to performance. I'd rather not complicate it.

PR #122, opened 2018-07-16
Reduce allocations

kornelski:
ScanLine iterator can borrow lines instead of copying them

### DIFF
--- a/src/reduction/alpha.rs
+++ b/src/reduction/alpha.rs
@@ -1,12 +1,14 @@
 use png::PngData;
 
-pub fn reduce_alpha_channel(png: &mut PngData, channels: usize) -> Option<Vec<u8>> {
-    let byte_depth: u8 = png.ihdr_data.bit_depth.as_u8() >> 3;
-    let bpp: usize = channels * byte_depth as usize;
-    let colored_bytes = bpp - byte_depth as usize;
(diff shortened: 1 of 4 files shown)


src/reduction/alpha.rs
```diff
@@ -1,24 +1,35 @@
...
-    let bpp: usize = bpp_factor * byte_depth as usize;
+    let bpp: usize = channels * byte_depth as usize;
     let colored_bytes = bpp - byte_depth as usize;
     for line in png.scan_lines() {
-        reduced.push(line.filter);
-        for (i, byte) in line.data.iter().enumerate() {
+        for (i, &byte) in line.data.iter().enumerate() {
             if i % bpp >= colored_bytes {
-                if *byte != 255 {
+                if byte != 255 {
```
shssoichiro: I'm not sure that removing the need to allocate in the case that the alpha channel can't be eliminated is worth the additional cost of iterating over the image twice. In the no optimization case, the allocated `reduced` Vec will be dropped at the end of this function regardless.

kornelski: I've removed 64-bit modulo used in the loop, which made iteration essentially free, so now it's faster in both case when it reduces and when it doesn't.

Issue #126, opened 2018-07-21
cfzlib processing many files causes system to OOM


shssoichiro:
When the `cfzlib` feature is enabled, and many input files are passed into oxipng at once, oxipng will consume massive amounts of memory and cause the system to run out of memory.

Need to determine if the memory leak is from cfzlib upstream or from oxipng's usage of it. Marking as lower priority because it is behind a feature flag.

PR #132, opened 2018-08-04
Cleanup on error

kornelski:
That's probably the memory leak you've noticed


shssoichiro:
Yes, this did fix #126

PR #135, opened 2018-09-14
Chunk name as 4 bytes

kornelski:
Avoids unnecessary small temporary allocations and pointers larger than the value pointed to.

### DIFF
--- a/src/headers.rs
+++ b/src/headers.rs
@@ -49,7 +49,7 @@ pub fn file_header_is_valid(bytes: &[u8]) -> bool {
 
 #[derive(Debug, Clone, Copy)]
 pub struct RawHeader<'a> {
-    pub name: &'a [u8],
+    pub name: [u8; 4],
     pub data: &'a [u8],
 }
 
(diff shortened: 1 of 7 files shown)


src/headers.rs
```diff
@@ -47,17 +47,11 @@ pub fn file_header_is_valid(bytes: &[u8]) -> bool {
...
-    pub name: &'a [u8],
-    pub data: &'a [u8],
-}
-
 pub fn parse_next_header<'a>(
     byte_data: &'a [u8],
     byte_offset: &mut usize,
     fix_errors: bool,
-) -> PngResult<Option<RawHeader<'a>>> {
+) -> PngResult<Option<([u8; 4], &'a [u8])>> {
```
shssoichiro: I disagree with removing the RawHeader struct in favor of a tuple. I don't feel that it's necessary in order to clean up everything else, and it makes the code harder to understand. If the type of `name` needs to be changed within `RawHeader`, I'm good with that.

PR #143, opened 2018-11-13
Identify and drop useless sRGB profiles

kornelski:
libpng has a [list of profiles](https://github.com/glennrp/libpng/blob/a294c1bcb65b44d0875ecd28a628f39f1253dce1/png.c#L2257) it considers sRGB and/or "bad", and this makes pngcrush drop these profiles.

I've added similar logic that replaces these sRGB-ish profiles with an equivalent `sRGB` chunk.


shssoichiro:
Actually, this seems to be breaking the `strip_headers_safe` test. Could you verify if the test needs to be updated, or if there is a code issue that needs to be resolved?

Issue #152, opened 2018-11-30
Framework for trying different color modes


kornelski:
I'm working on implementing #74 

The problem is that the main loop which tries different compression modes/filters assumes there's a single PNG file, and only IDAT differs.

However, palette sorting makes palette + IDAT dependent, so the compression trials can't just pick a smaller IDAT, they have to also set the right palette to match.

The alpha filtering trials get away with this, because they don't change the image mode, but I'm still not entirely happy with alpha trials running compression trials "internally" as part of reduction, rather than being integrated with the main loop of trials (e.g. the side effect is that verbose mode doesn't show alpha mode compressions tested, and the winning IDAT is recompressed later with the same settings).

So I think some refactoring is needed. That's also related to #145

I could just compress with default settings as part of the palette reduction, same as alpha reductions do it, but that feels like copying&pasting a lot of code, and will perform redundant compressions.

So I'm thinking about introducing an object that will explicitly manage compression trials, which could be given to alpha and palette reductions to add their trials to the queue, so that the progress is properly displayed and the best IDAT is saved.

WDYT?

shssoichiro:
On one hand, this makes sense. My concern, however, and the reason I implemented alpha reductions as separate compression tries, is to avoid an exponential increase in number of trials. For example, a full alpha optimization will currently run 6 trials, one for each alpha setting, and after choosing the best, then runs through the normal number of trials (e.g. 48), for a total of 48 + 6 trials. If we aren't careful, we could end up running 48 * 6 trials, and explode the time oxipng takes. If we can avoid that problem, I'm onboard with this refactor.

kornelski:
I've also found a case where reducing an image to 4 bits compresses better than 2-bit image. That would need compression-based check, too.

So my current thinking is:

1. Pick one fast and reasonably representative compression setting, e.g. f=0 zs=0 zx=6 (perhaps level depending on `-o` setting)
2. Immediately recompress IDAT with that setting, and keep it (or at least make note of the size)
3. After each reduction, compress with that setting and keep the reduction only if it compresses same or better.

Unfortunately, this means order of reductions matters, but checking of multiple permutations of reductions would be too expensive.

The step 2 has to exist, since comparing against compressed IDAT already in the file will give false negatives, e.g. zopfli-compressed bad color mode may seem better than good color mode that hasn't been compressed hard enough yet.

andrews05:
[...]

PR #154, opened 2018-12-04
Non-self-mutable reductions

kornelski:
Instead of immediately, mutably modifying `self`, reductions return a struct with the changed fields.

For now, this is just work in progress, but in the future it'll allow two things:

* Run some reductions in parallel
* Apply reductions only after verifying they're beneficial — #152

### DIFF
--- a/src/png/mod.rs
+++ b/src/png/mod.rs
@@ -328,14 +328,15 @@ impl PngData {
 
     /// Attempt to reduce the bit depth of the image
     /// Returns true if the bit depth was reduced, false otherwise
-    pub fn reduce_bit_depth(&mut self) -> bool {
+    #[must_use]
+    pub fn reduce_bit_depth(&self) -> Option<ReducedPng> {
         if self.ihdr_data.bit_depth != BitDepth::Sixteen {
             if self.ihdr_data.color_type == ColorType::Indexed
                 || self.ihdr_data.color_type == ColorType::Grayscale
             {
                 return reduce_bit_depth_8_or_less(self);
             }
-            return false;
+            return None;
         }
 
         // Reduce from 16 to 8 bits per channel per pixel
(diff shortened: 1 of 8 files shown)


src/png/mod.rs
```diff
@@ -490,26 +376,35 @@ impl PngData {
         // Go down one step at a time
         // Maybe not the most efficient, but it's safe
         if self.ihdr_data.color_type == ColorType::RGBA {
-            if reduce_rgba_to_grayscale_alpha(self) || reduce_rgba_to_rgb(self) {
+            if reduce_rgba_to_grayscale_alpha(self) {
```
shssoichiro: Are this and the later `reduce_rgb_to_grayscale` intended to be left as is? Is that part of the aforementioned WIP?

src/reduction/mod.rs
```diff
@@ -1,3 +1,163 @@
...
+    /// replace if Some
+    pub palette: Option<Vec<RGBA8>>,
+    /// replace if Some
+    pub transparency_pixel: Option<Vec<u8>>,
+    /// replace if Some, delete if None
+    pub aux_headers: HashMap<[u8; 4], Option<Vec<u8>>>,
+}
+
+/// Attempt to reduce the number of colors in the palette
+/// Returns true if the palette was reduced, false otherwise
```
shssoichiro: The second line of this comment is outdated. Could you remove it please?

src/png/mod.rs
kornelski: That's the WIP part. I'll convert other ones too.

src/reduction/mod.rs
kornelski: Changed

kornelski:
I've finished changing other reductions

kornelski:
The CI failure is due to rust nightly/rustup not working.

PR #158, opened 2019-01-14
Make reductions return a new uncompressed image

### LINKED ISSUE #145: Dual purpose and implicit state in PngData
`PngData` has both `idat_data` and `raw_data`, which are related, but used in different contexts and handled as independent fields.

This is problematic, because:

* It's possible to modify `raw_data` without invalidating `idat_data`. Currently the code has to keep track of that "manually" through a flag passed around. That feels fragile.

* Copying of `PngData` copies both of them, even when it's just for trials/reductions which are going to use only one of them.

* Alpha filter trials get `filtered_image` version of the idat that needs to be recompressed better later, but there's no way to store the filtered version, so later recompression does filtering again (that's very minor though).

I'm not entirely sure how this should be handled, but there must be a better way. Some ideas:

* Have separate CompressedPng and UncompressedPng objects, with only either raw or idat, and try to use them throughout the code accordingly.
* Have one body as enum `Compressed`/`Uncompressed`/`Both`
* Have getters/setters on PngData that clear `idat_data` when `raw_data` is set or borrowed mutably.

kornelski:
Continues making reductions immutable, and partially fixes #145


kornelski:
Ouch, dropping worse-compressing reductions has broken tests that expected these reductions

kornelski:
The test failure on Windows is surprising. I've tested on my Windows machine, with both rayon and without, and it worked fine.

I think it could happen if multiple reductions compress to the exact same size, and then the (non-deterministically) last one to compress wins. But a tie in compression size seems quite unlikely.

shssoichiro:
It might make sense to make the compression deterministic, e.g. we should prefer lower bit-depth if there's otherwise a tie.

kornelski:
Another source of difference was zlib implementation. On very small files (~15 bytes long) file sizes from miniz and cfzlib differ by 1 byte, and cause a different depth to be chosen.

kornelski:
In the end I've skipped problematic test with miniz. It's an unusual file that compresses over 100KB of data to 890 bytes, and the difference between various modes is just 20 bytes, so it's a very close call.

kornelski:
Finally ready :)

PR #160, opened 2019-01-26
Timeout

kornelski:
(requires #158)

Short timeouts didn't work well with large images, because the timeout was checked rarely.

This adds more checks of `deadline.passed()`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -341,6 +341,8 @@ pub fn optimize(input: &InFile, output: &OutFile, opts: &Options) -> PngResult<(
         eprintln!("Processing: {}", input);
     }
 
+    let deadline = Arc::new(Deadline::new(opts.timeout, opts.verbosity.is_some()));
+
     let in_data = match *input {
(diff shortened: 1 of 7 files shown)


kornelski:
Rebased

shssoichiro:
Sorry, one more from merging #162. :)

src/lib.rs
```diff
@@ -341,6 +341,8 @@ pub fn optimize(input: &InFile, output: &OutFile, opts: &Options) -> PngResult<(
         eprintln!("Processing: {}", input);
     }
 
+    let deadline = Arc::new(Deadline::new(opts.timeout, opts.verbosity.is_some()));
```
shssoichiro: Does it make more sense to instantiate the Deadline (here and on 443) inside `optimize_png` to avoid the duplication?

kornelski: It starts counting time from when it's instantiated, and I wanted it to also include time it takes to load the image.

PR #163, opened 2019-01-28
Avoid mixing Rayon with Mutex

kornelski:
Previous loop had Mutex inside rayon's for_each, and [this was dangerous](https://github.com/rayon-rs/rayon/issues/592).

This reorganizes the loop to be single threaded, fed by multiple async tasks. This makes evaluation even more async than before.


shssoichiro:
Looks like this is dependent on #160, so I will wait until that one is ready before reviewing this.

PR #171, opened 2019-06-10
Fixes for grey scale reduction

fhanau:
If I understand the color conversion code in [lodepng](https://github.com/lvandeve/lodepng/blob/master/lodepng.cpp#L2936) correctly, the RGB values that are represented by 2-bit and 4-bit greyscale in bit depth reduction are wrong.
With the current implementation, oxipng corrupts a proprietary file because it thinks it can be reduced to 2-bit greyscale and quits when it detects that the result is corrupted.


shssoichiro:
I assume since this is a proprietary file you wouldn't be able to make a test case out of it. But this seems reasonable. Thanks!

fhanau:
Here's a synthetic test file created using `convert -size 4x4 canvas:'#303030' -type GrayScale issue-171.png` 
issue-171

PR #172, opened 2019-06-10
Fix info message and typos

fhanau:
Fixes a few typos discovered by codespell and a verbose message that has a wrong number of colors since the palette object stores one entry per color.


shssoichiro:
Thanks!

PR #173, opened 2019-06-10
Make order of chunks deterministic by sorting chunks

fhanau:
This patch makes the order of chunks deterministic, which is helpful when debugging and not removing metadata.
A BTreeMap is created from the HashMap instead of replacing the HashMap entirely since BTreeMap lacks the retain method.


shssoichiro:
I'd prefer to make this work with an in-place BTreeMap, to avoid the extra cloning. I have a commit that should do this, so I'll push it up.

shssoichiro:
Addressed in 5b1121c

PR #177, opened 2019-06-18
Enable link-time optimization for release builds

Calinou:
This creates faster and smaller binaries, at the cost of longer build times.

## Benchmark

- **OS:** Fedora 30
- **CPU:** i7-6700K @ 4.4 GHz
- **Compiler:** rustc 1.35.0

This comparison was made by running `oxipng -Zs` on all images from [this repository](https://github.com/minetest-mods/moreblocks) (while they were already optimized).

## Before

```text
Benchmark #1: oxipng -Zs **/*.png
  Time (mean ± σ):      1.576 s ±  0.048 s    [User: 2.486 s, System: 0.238 s]
  Range (min … max):    1.518 s …  1.654 s    10 runs
```

## After

```text
Benchmark #1: oxipng -Zs **/*.png
  Time (mean ± σ):      1.495 s ±  0.030 s    [User: 2.365 s, System: 0.232 s]
  Range (min … max):    1.429 s …  1.530 s    10 runs
```

It's not huge, but there's a noticeable speedup :slightly_smiling_face:

The release binary now weighs only 1.4 MB after stripping (compared to 1.7 MB before).


shssoichiro:
Thanks, especially for doing the benchmarks on this! Seems like it's worth the build time cost.

PR #180, opened 2019-06-21
Preallocate memory in reduced_alpha_to_up

SethDusek:
For this [image](https://proxy.duckduckgo.com/iu/?u=https%3A%2F%2Fwww.pngarts.com%2Ffiles%2F4%2FSpace-Rocket-PNG-Image-Transparent-Background.png&f=1) when using oxipng -a, valgrind reports 20 less allocs/frees

Original: 
>==8410==   total heap usage: 5,305,656 allocs, 5,305,577 frees, 9,187,337,048 bytes allocated

With this patch:
>==8977==   total heap usage: 5,305,636 allocs, 5,305,557 frees, 9,186,624,996 bytes allocated

It's a pretty low-hanging fruit optimization, which is maybe why I picked it up :P. Ideally though oxipng should use some kind of memory pool data structure to allocate memory and allow stages of the optimizer to reuse it instead of allocating. There are tons of new vecs created from function to function. This patch could also be optimized a bit more maybe, by also accounting for pixel size when allocating.

Also I wasn't quite able to run benchmark tests on this. Seemingly the benchmark tests fail to compile due to missing functions? Is this problem reproducible for other people?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,6 +37,7 @@ zopfli = "^0.4.0"
 miniz_oxide = "0.2.0"
 rgb = "0.8.11"
 
+
 [dependencies.rayon]
 optional = true
 version = "^1.0.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,7 +37,6 @@ zopfli = "^0.4.0"
 miniz_oxide = "0.2.0"
 rgb = "0.8.11"
 
-
 [dependencies.rayon]
 optional = true
 version = "^1.0.0"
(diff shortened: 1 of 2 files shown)


SethDusek:
Allocations reduced even more by also adding a filter byte in the count,

shssoichiro:
Thanks! I have not done well with keeping the benchmark tests running, so they probably are broken. I would like to add them to the CI pipeline to keep them from breaking, but some of them are quite slow, so that hasn't been an option.

PR #182, opened 2019-07-16
Interlacing fixes

fhanau:
Oxipng currently produces corrupt output when deinterlacing files with width or height <= 2.


fhanau:
Since I'm on the subject: Why not deinterlace images by default? It is a feature that is rarely useful AFAIK and reduces file size a lot on affected files.

shssoichiro:
The decision was for parity with Optipng, originally. Optipng preserves interlacing state of the input image. There's no reason to keep that parity, but I don't know how many people would be surprised (or care?) if their interlaced images suddenly became deinterlaced. It is probably useful to deinterlace by default, but I'd like to receive some more input on the topic first.

PR #184, opened 2019-07-18
Implement palette sorting

fhanau:
Sorting is based on ascending alpha channel value and descending luminosity. I only encountered very small gains when additionally using popularity based sorting alternatively, so I think this is a sensible default. It should be easy to implement additional approaches to sorting when desired.
I tried using the evaluation system to try both compression with and without palette sorting, but that resulted in smaller in compression gains (perhaps something wrong with the evaluation code), so enabling this by default is worth considering.


fhanau:
I added sorting for files that have to be reduced to use the palette color type.

PR #185, opened 2019-07-19
Use heuristics for filter selection

fhanau:
The filter strategy is evaluated heuristically by trying all filters at compression level 1 and only using full compression with the best result. In this patch, it is enabled by default on level 2. While compression is a little bit worse since the heuristic is sometimes wrong, performance improves by around a third in my tests. Alternatively, all filters could be tested heuristically, which is only slightly slower and produces better filtering than the current approach for level 2.

This represents a notable deviation from what optipng does. If you think that oxipng's behaviour should mirror optipng, it's probably best to only enable this if requested through the command line.
The same approach can be used to determine the best strategy heuristically (possibly with a higher compression level than 1, which helped with choosing the right one in my tests).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -540,6 +540,8 @@ fn optimize_png(
                 (*f, png.raw.filter_image(*f))
             })
             .collect();
+
+        //Use heuristics to choose the best filter strategy.
         if opts.use_heuristics && filter.len() > 1 {
             let best_size_heuristics = AtomicMin::new(None);
             let mut heuristic_results: Vec<TrialOptions> = Vec::with_capacity(filter.len());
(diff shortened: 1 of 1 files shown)


fhanau:
Updated to use similar heuristics for zlib strategy.

shssoichiro:
In this case I think I'd want to prefer the strategy that gives the best compression. With the reduction (other PR) that is fine to be fast, because it's likely to be wrong on only a very small percentage of images. Since the filter trials are more likely to be wrong, and can already be controlled by a CLI setting, I think it makes sense to keep the current approach here.

fhanau:
There is a chance of these heuristics getting the right filter wrong, but it only worsens compression by a small amount in my tests. When using the heuristics in addition to using `-o2 -f0-5`, both speed and compression are better compared to using `-o2` without the heuristic. Since it will worsen compression on some files, it is only enabled for mode 2 with this patch, offering a space/speed tradeoff not possible through the current command line interface. It would be great if you or someone else could benchmark this code on a set of images to see if my results can be reproduced.

PR #187, opened 2019-08-01
Don't apply any alpha optimizations if they are disabled

### LINKED ISSUE #164: Truly lossless by default?
Glad to see that a replacement for optipng exists, but a few things in the usage info have me concerned that oxipng is only perceptually lossless and might trash alpha pixels by default. If so is there any option I may use to prevent any alpha optimizations at all? In my use case, optimizing alpha will lead to black outlines around pixels. Otherwise I cannot safely use this tool.

SethDusek:
This should fix #164

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -285,9 +285,7 @@ impl Default for Options {
         for i in 0..4 {
             strategies.insert(i);
         }
-        let mut alphas = HashSet::new();
-        alphas.insert(colors::AlphaOptim::NoOp);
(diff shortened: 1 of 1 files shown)


shssoichiro:
Given the discussion around how alpha reduction may not be lossless in some cases, I think it may be better to never enable alpha reduction with a preset, i.e. only enable if the user passes the `--alpha` flag. It's probably also worth adding something like "This may not be lossless for some applications." to the help text for the `--alpha` setting.

PR #188, opened 2019-08-05
Use fast heuristic for reduction evaluation

fhanau:
This improves runtime by 8% on my benchmark. Surprisingly, it only causes output files to be a fraction of a percent larger. I feel like this compression loss definitely _should_ be larger. It might just be an issue with the images I use for testing, but it seems to me that something is going wrong in the evaluation code. I also think that evaluation is not that impactful on many images since several reductions are rarely not worth doing.

Note that I have not refactored this code to fix warnings since it is a significant change and I want to hear your opinion first.

### DIFF
--- a/src/evaluate.rs
+++ b/src/evaluate.rs
@@ -86,7 +86,8 @@ impl Evaluator {
         // but results will be collected via the message queue
         let eval_send = self.eval_send.clone();
         rayon::spawn(move || {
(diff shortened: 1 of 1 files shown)


shssoichiro:
Yeah, I think this is a good change. The reduction evaluation should be as fast as possible while still being reasonably accurate.

PR #191, opened 2019-08-12
Reduce alloc in filtering/unfiltering by reusing buffers

SethDusek:


shssoichiro:
Thanks! Time difference is in the noise, but peak memory usage looks to be down about 5%.

SethDusek:
> Thanks! Time difference is in the noise, but peak memory usage looks to be down about 5%.

I was meaning to talk about that but got kinda caught up. It seems that perf is reporting more "page faults" with this patch as compared to before. I'm not sure why this is, but it'll take a deeper investigation most probably

PR #193, opened 2019-08-23
Palette sorting

kornelski:
Fixes #184

### DIFF
--- a/src/reduction/mod.rs
+++ b/src/reduction/mod.rs
@@ -61,9 +61,24 @@ pub fn reduced_palette(png: &PngImage) -> Option<PngImage> {
             }
         }
 
+        let mut used_enumerated : Vec<(usize, &bool)>= used.iter().enumerate().collect();
+        used_enumerated.sort_by(|a, b| {
+            //Sort by ascending alpha and descending luma.
+            let color_val = |palette : &Vec<rgb::RGBA<u8>>, i| {
+                let color = palette.get(i).cloned()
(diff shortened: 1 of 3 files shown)


src/reduction/mod.rs
```diff
@@ -61,9 +61,24 @@ pub fn reduced_palette(png: &PngImage) -> Option<PngImage> {
             }
         }
 
+        let mut used_enumerated : Vec<(usize, &bool)>= used.iter().enumerate().collect();
+        used_enumerated.sort_by(|a, b| {
```
shssoichiro: You can use itertools to make this a little more concise.

src/reduction/mod.rs
```diff
@@ -61,9 +61,24 @@ pub fn reduced_palette(png: &PngImage) -> Option<PngImage> {
...
+            //Sort by ascending alpha and descending luma.
+            let color_val = |palette : &Vec<rgb::RGBA<u8>>, i| {
+                let color = palette.get(i).cloned()
+                    .unwrap_or_else(|| RGBA8::new(0, 0, 0, 255));
+                let mut result = (color.a as i32) << 18;
+                result -= (color.r as i32) * 299;
+                result -= (color.g as i32) * 587;
+                result -= (color.b as i32) * 114;
```
shssoichiro: Could you expand your comment to indicate where these constants are coming from?

fhanau: RGB luma values are derived from [here](https://www.w3.org/TR/AERT/#color-contrast). Unfortunately, I have never used itertools before so I wouldn't know what to change.

PR #194, opened 2019-08-23
Allow usage on wasm32-unknown-unknown target

### LINKED ISSUE #157: WebAssembly version?
Wondering if anyone has / is planning to do a [WebAssembly](https://rustwasm.github.io/book/) version for NodeJS/browser usage?

RReverser:
This PR allows to compile and use OxiPNG on `wasm32-unknown-unknown` target, thus fixes #157.

The changes are described in commit messages, but, as a short summary, two primary fixes are:
 - Avoiding call to `Instant::now()` in `Deadline::new` when no `Timeout` is set. This is required because `Instant::now()` panics on `wasm32-unknown-unknown` (as it tries to retrieve time, for which there is no generic Wasm syscall implementation), and so allows to initialise OxiPNG with default options (no timeout) while still allowing to use timeout on other platforms.
 - Trying all the combinations on the main thread when `feature = "parallel"` is unset. Threads proposal for Wasm is not finalised and will take time to be ready in mainstream implementation, so this is also important for OxiPNG to work on Wasm target.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -784,19 +784,25 @@ fn perform_reductions(
     try_alpha_reductions(png, &opts.alphas, eval);
 }
 
+struct DeadlineImp {
+    start: Instant,
+    timeout: Duration,
+    print_message: AtomicBool,
+}
+
 /// Keep track of processing timeout
 pub(crate) struct Deadline {
-    start: Instant,
-    timeout: Option<Duration>,
-    print_message: AtomicBool,
+    imp: Option<DeadlineImp>,
 }
 
 impl Deadline {
     pub fn new(timeout: Option<Duration>, verbose: bool) -> Self {
         Self {
-            start: Instant::now(),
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -779,31 +779,37 @@ fn perform_reductions(
     try_alpha_reductions(png, &opts.alphas, eval);
 }
 
-/// Keep track of processing timeout
-pub(crate) struct Deadline {
+struct DeadlineImp {
```
shssoichiro: Does Imp = Implementation? I think `Impl` would be a more easily understood abbreviation. Alternatively it's common in Rust to name it something like `DeadlineInner`.

RReverser: I found `imp` pretty common in Rust land, because you can't use `impl` as a field name, and it's better to have consistency between type and field here. `Inner` works too I guess.

Issue #198, opened 2020-03-10
Non-deterministic output


RReverser:
Trying to run e.g.

```bash
cargo run -- -o 6 -P tests/files/grayscale_alpha_16_reduce_alpha_right.png --alpha
```

on current master (39ff7cfd5c8403af82df097f7a227ebac370254a) seems to produce non-deterministic results.

Sometimes I'm getting

```
Processing: tests/files/grayscale_alpha_16_reduce_alpha_right.png
    200x152 pixels, PNG format
    2x16 bits/pixel, GrayscaleAlpha
    IDAT size = 32322 bytes
    File size = 32400 bytes
Trying: 216 combinations
Found better combination:
    zc = 9  zs = 1  f = 5        20295 bytes
    IDAT size = 20295 bytes (12027 bytes decrease)
    file size = 20373 bytes (12027 bytes = 37.12% decrease)
Running in pretend mode, no output
```

and sometimes (more frequently) a worse result

```
Processing: tests/files/grayscale_alpha_16_reduce_alpha_right.png
    200x152 pixels, PNG format
    2x16 bits/pixel, GrayscaleAlpha
    IDAT size = 32322 bytes
    File size = 32400 bytes
Trying: 216 combinations
Found better combination:
    zc = 9  zs = 1  f = 5        20335 bytes
    IDAT size = 20335 bytes (11987 bytes decrease)
    file size = 20413 bytes (11987 bytes = 37.00% decrease)
Running in pretend mode, no output
```

Originally posted as comments on https://github.com/shssoichiro/oxipng/commit/239ca81db7b45b3ac731a8d8b2746e760500d0a9#commitcomment-37751319.

RReverser:
Verbose logs don't reveal anything obvious - for same combinations the "good" result just always outputs better sizes than the "bad" result for same combinations.

RReverser:
Can reproduce even with `-t 1` (single thread), so either that option is ignored somewhere, or at least it's not a thread race, which makes it even weirder...

RReverser:
Looks like it's a result of different evaluation result "winning" in different cases due to different order.

RReverser:
Looks like if I remove the "bias" code in the evaluator, then I consistently get the better results.

It seems to make sense, given that we always want to compare actual lengths independently from order, but maybe I don't understand the role of that code there in the first place?

RReverser:
[...]

PR #199, opened 2020-03-10
Remove the concept of size bias

kornelski:
#198


RReverser:
I'm now more confused than before :) From your explanation in https://github.com/shssoichiro/oxipng/issues/198#issuecomment-597316339, it seems that `bias` is generally a useful concept and should result in a better compression in most cases, and it's just non-determinism that has to be fixed?

kornelski:
I think it's a useful concept in the abstract, but this implementation has caused more trouble than it's worth, so let's simplify. A better one can be added later.

kornelski:
@shssoichiro I think it's good to merge this one

shssoichiro:
Agreed, Sorry for delay, been busy with IRL. ^^;

PR #200, opened 2020-03-11
Respect deadline in miniz

RReverser:


shssoichiro:
Could you clarify the motivation behind this change? It seems like this shouldn't help significantly, because we're already checking the deadline after each trial. Would we want to kill in-progress trials?

RReverser:
> Would we want to kill in-progress trials?

Yeah, this is for consistency with same existing functionality in cloudflare-zlib: https://github.com/shssoichiro/oxipng/blob/121558bf5cb3883296a74b09b40b70bbd64930a2/src/deflate/cfzlib.rs#L34-L36

cloudflare-zlib is a faster library than miniz, so miniz supposedly needs such functionality even more on platforms where the former is not supported.

However, I'm personally not using `deadline` option actively, so I don't have very strong opinion, just a difference in functionality of different libraries I've noticed while looking at code.

PR #202, opened 2020-03-16
Improve performance consistency

RReverser:
Switch from HashMap / HashSet to IndexMap / IndexSet for consistent iteration order of various options and, as a result, more predictable performance.


Cargo.toml
```diff
@@ -36,6 +36,8 @@ num_cpus = "^1.0.0"
 zopfli = "^0.4.0"
 miniz_oxide = "0.3"
 rgb = "0.8.11"
+vec_map = "0.8.1"
```
shssoichiro, line 39: I don't see this dependency used anywhere. Can it be removed?

RReverser: Oops, yes, that was a temporary experiment. Good catch.

RReverser:
@shssoichiro I suppose it's obvious, but just in case - note that this changed types of some public-facing options, so it will probably require bump to 2.4.0.

PR #203, opened 2020-03-16
Add libdeflater as an option

RReverser:
libdeflater is a Rust wrapper around [libdeflate](https://github.com/ebiggers/libdeflate) - an alternative heavily optimised library for deflate/zlib/gzip compression and decompression that is intended for situations where upper bounds of the output are well-known.

In my benchmarks on test files in the repo it has shown to be usually both faster and providing better compressed output than cloudflare-zlib, but in some cases showing the opposite, so rather than swapping defaults, it's currently provided as another option,
similarly to zopfli.

Since it's not strictly better in all cases, I'm not providing median numbers, but you can check distribution histograms for time and size differences here (all using `oxipng -o 6 -t 6 -P`): https://docs.google.com/spreadsheets/d/1WOKgeYZBhLkQvMGAC36snN4azilElzOFhx63RJu0EZY/edit?usp=sharing

---

# Highlights


RReverser:
Note: this was using #199 and #202 to make numbers more reproducible.

RReverser:
> If it's faster on average, might be worth making it the default in the future...

Maybe. For start I just wanted to experiment with such integration, get numbers and see whether you or someone else would be interested :)

RReverser:
Just FYI: this has now made it to squoosh.app as the default compressor to replace OptiPNG.

PR #210, opened 2020-04-15
Move Zlib-specific options under Zlib compressor

RReverser:
Make it statically obvious that they're not compatible with other compressors.


shssoichiro:
Seems fine from a code standpoint. Could you please run rustfmt to make the builds happy?

RReverser:
> Noting this is a breaking API change, but a reasonable one

Yeah some other PRs were breaking changes as well, so I suppose they can be grouped together.

PR #218, opened 2020-04-17
Switch to standard logging library

### LINKED ISSUE #217: Switch to standard logging approach
Right now, `lib.rs` takes care of logging by printing all messages to `eprintln` when configured to do so via own `verbosity` option.

It might be better to switch to a [standard logging facade](https://github.com/rust-lang/log) compatible with rest of the ecosystem.

This way, users of the library could configure logging from a single place with any [logging library](https://github.com/rust-lang/log#in-executables) of their preference. This would also allow them to [configure](https://docs.rs/env_logger/0.7.1/env_logger/#enabling-logging) verbosity from a single place, filter logs [based on library](https://docs.rs/env_logger/0.7.1/env_logger/#filtering-results), redirect any logs to the desired file, use logger that is specific to some targets like [WebAssembly](https://docs.rs/console_log/0.1.2/console_log/) or [Android](https://docs.rs/android_log/0.1.3/android_log/), or even compiling debug logs away to no-op in case they don't want it to be present in the final artifact.

RReverser:
This allows to configure or compile away logging in the library from a single place in Rust apps.

For the CLI side, the usage and output remained the same, except it's now colour-coded.

Fixes #217.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -405,7 +405,7 @@ pub fn optimize_from_memory(data: &[u8], opts: &Options) -> PngResult<Vec<u8>> {
     let optimized_output = optimize_png(&mut png, data, opts, deadline)?;
 
     if is_fully_optimized(original_size, optimized_output.len(), opts) {
-        warn!("Image already optimized");
+        info!("Image already optimized");
         Ok(data.to_vec())
     } else {
         Ok(optimized_output)
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -27,6 +27,7 @@ use crate::png::PngImage;
 use crate::reduction::*;
 use crc::crc32;
 use image::{DynamicImage, GenericImageView, ImageFormat, Pixel};
+use log::{debug, error, info, warn};
 use rayon::prelude::*;
 use std::fmt;
 use std::fs::{copy, File};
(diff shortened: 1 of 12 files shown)


RReverser:
(fixing tests)

RReverser:
Fixed tests, also expanded `verbose_mode` test to verify that we produce expected verbose logs.

RReverser:
> For the CLI side, the usage and output remained the same, except it's now colour-coded.

To show an example - this is what I mean by colour-coded (specific colours depend on the terminal theme, of course):
image

src/lib.rs
```diff
@@ -419,7 +405,7 @@ pub fn optimize_from_memory(data: &[u8], opts: &Options) -> PngResult<Vec<u8>> {
     let optimized_output = optimize_png(&mut png, data, opts, deadline)?;
 
     if is_fully_optimized(original_size, optimized_output.len(), opts) {
-        eprintln!("Image already optimized");
+        warn!("Image already optimized");
```
shssoichiro, line 408:
```suggestion
        info!("Image already optimized");
```

This probably makes more sense at `info`

RReverser: Yeah, I was on the fence because I thought that someone might want it even if they set logging level to exclude `info` (to `warn`), but makes sense.

src/lib.rs
```diff
@@ -695,7 +670,7 @@ fn optimize_png(
         }
     }
 
-    eprintln!(
+    error!(
```
RReverser, line 673: @shssoichiro By the way, I'm not sure whether it's worth keeping this log message if it's duplicate in an actual error too (and will be printed e.g. by CLI app twice). Thoughts?

shssoichiro: I'm not entirely sure. But _theoretically_ this should never show up, and if it does, we'll definitely want the user to be aware of it. So I think it's fine to leave it.

Issue #221, opened 2020-04-19
Is preset 4 deprecated or incorrect?


RReverser:
Currently `--help` says:

```
Optimization levels:
    -o 0  =>  --zc 3 --nz                  (0 or 1 trials)
    -o 1  =>  --zc 9                       (1 trial, determined heuristically)
    -o 2  =>  --zc 9 --zs 0-3 -f 0,5       (8 trials)
    -o 3  =>  --zc 9 --zs 0-3 -f 0-5       (24 trials)
    -o 4  =>  --zc 9 --zs 0-3 -f 0-5 -a    (24 trials + 6 alpha trials)
    -o 5  =>  --zc 3-9 --zs 0-3 -f 0-5 -a  (96 trials + 6 alpha trials)
    -o 6  =>  --zc 1-9 --zs 0-3 -f 0-5 -a  (180 trials + 6 alpha trials)
```

However, actual code for preset 4 is equivalent to preset 3 (no alpha optimisation):

```
    fn apply_preset_3(mut self) -> Self {
        for i in 1..5 {
            self.filter.insert(i);
        }
        self
    }

    fn apply_preset_4(self) -> Self {
        self.apply_preset_3()
    }
```

I suppose this is the outcome of a decision not to perform alpha optimisations by default, made in https://github.com/shssoichiro/oxipng/pull/187.

Which is fine, but makes state of this preset fairly confusing. Should it be removed at this point?

Also, docs for higher opt levels should probably be updated as well to exclude `-a`?

TPS:
Maybe for backwards compatibility for existing users (especially scripting & such)?

RReverser:
Maybe, but given the planned major release, it might be worth finally removing it.

And, even before that, the docs likely weren't updated by accident.

shssoichiro:
> I suppose this is the outcome of a decision not to perform alpha optimisations by default, made in #187.

> Maybe for backwards compatibility for existing users (especially scripting & such)?

Both of these assumptions are correct. It may make sense to remove it, but my concern is for CLI users who are more likely to not expect a config option to be removed. For this reason, I suggest if we remove level 4 and squash the levels to 0-5, the CLI should allow level 6 and set it to the same as level 5. (It might even make sense to allow any preset level as input, and clamp to a max of level 5? This is what GCC does if you pass e.g. `-O9` although I'm mixed on whether this is the best approach.)

TPS:
[...]

PR #224, opened 2020-04-20
Add "max" level alias; various level-related warnings

RReverser:
- Show a deprecation warning for `Options::from_preset(4)` / `-o 4` as per #221. Can be still removed later, but for now at least avoids confusion.
 - Show warning for usage of levels 4 and above for Zopfli and Libdeflate. This is not immediately obvious to users, but such levels end up iterating over same parameters as level 3 for these compressors, since there are no fine-tuned compressor options.
 - Adds `Options::max_compression` and `-o max` to always refer to best compression as discussed in #221.
 - Warns about usage of levels that don't exist yet. Previously these would silently downgrade to `-o 6`, but I think it's worth showing a warning now that we have explicit `max_compression` method / option.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -233,8 +233,8 @@ fn main() {
     -o 2   =>  --zc 9 --zs 0-3 -f 0,5       (8 trials with zlib or 2 trials with other compressors)
     -o 3   =>  --zc 9 --zs 0-3 -f 0-5       (24 trials with zlib or 6 trials with other compressors)
     -o 4   =>                               (deprecated; same as `-o 3`)
-    -o 5   =>  --zc 3-9 --zs 0-3 -f 0-5     (96 trials with zlib; same as `-o 3` for other compressors)
-    -o 6   =>  --zc 1-9 --zs 0-3 -f 0-5     (180 trials with zlib; same as -o 3` for other compressors)
+    -o 5   =>  --zc 3-9 --zs 0-3 -f 0-5     (168 trials with zlib; same as `-o 3` for other compressors)
+    -o 6   =>  --zc 1-9 --zs 0-3 -f 0-5     (216 trials with zlib; same as `-o 3` for other compressors)
     -o max =>                               (stable alias for the max compression)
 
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -227,13 +228,14 @@ fn main() {
...
-    -o 3  =>  --zc 9 --zs 0-3 -f 0-5       (24 trials)
-    -o 4  =>  --zc 9 --zs 0-3 -f 0-5 -a    (24 trials + 6 alpha trials)
-    -o 5  =>  --zc 3-9 --zs 0-3 -f 0-5 -a  (96 trials + 6 alpha trials)
-    -o 6  =>  --zc 1-9 --zs 0-3 -f 0-5 -a  (180 trials + 6 alpha trials)
+    -o 0   =>  --zc 3 --nz                  (0 or 1 trials)
+    -o 1   =>  --zc 9                       (1 trial, determined heuristically)
+    -o 2   =>  --zc 9 --zs 0-3 -f 0,5       (8 trials with zlib or 2 trials with other compressors)
+    -o 3   =>  --zc 9 --zs 0-3 -f 0-5       (24 trials with zlib or 6 trials with other compressors)
+    -o 4   =>                               (deprecated; same as `-o 3`)
+    -o 5   =>  --zc 3-9 --zs 0-3 -f 0-5     (96 trials with zlib; same as `-o 3` for other compressors)
```
RReverser, line 236: @shssoichiro Actually, wait, I'm not sure that 96 is / was correct here: `(9 - 3 + 1) * (3 - 0 + 1) * (5 - 0 + 1) = 168`.
And the one below: `(9 - 1 + 1) * (3 - 0 + 1) * (5 - 0 + 1) = 216`.

shssoichiro: I think you're right. Maybe this was right at some point in the past and then the settings got changed? Or my math was just wrong to begin with. Since you're in this code already, would you mind updating that? Thanks!

RReverser: Done.

PR #226, opened 2020-05-03
Bump

kornelski:
How about making a release?


shssoichiro:
Cool! I was waiting for the breaking changes to settle. I'll have to write up a Changelog, but I'll get that prepared here soon.

shssoichiro:
Closing this as we actually need to bump to 3.0 due to some breaking API changes. I'll get those uploaded today. Thanks!

PR #230, opened 2020-05-15
Add release description

Luni-4:
- Add release description
- Fix typo

Thanks in advance for your review! :)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -66,10 +66,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - name: Get the version
+      - name: Get version and release description
         id: tagName
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -85,7 +86,8 @@ jobs:
       - name: Create a release
         uses: softprops/action-gh-release@v1
         with:
-          name: ${{ steps.tagName.outputs.tag }}
+          name: v${{ steps.tagName.outputs.tag }}
+          body_path: CHANGELOG.txt
```
shssoichiro, line 90: I assume this automatically parses the Changelog for the version number and copies the release notes? That's neat.

Luni-4: `CHANGELOG.txt` contains only the release notes, while the version is obtained using `cargo pkgid`

Issue #134, opened 2018-09-11
Distribute Linux and macOS binaries


Calinou:
It would be helpful to have Linux and macOS binaries on the Releases page, so that users can get started with oxipng faster, especially on slower machines such as the Raspberry Pi. This would also make it possible to add oxipng to [Homebrew](https://brew.sh/) for easy installation on macOS.

I used [trust](https://github.com/japaric/trust) to cross-compile to various platforms/architectures in [one of my crates](https://github.com/Calinou/lagraph) and it worked well, so I would recommend it.

jansol:
Homebrew is shipping oxipng already. Linux binaries are generally better left to distro package managers as well. OTOH a static musl binary could come in handy at times.

TPS:
`OxiPNG` has been [added to the TUR](https://github.com/termux-user-repository/tur/pull/2134) yesterday for convenience & semi-official support! 🎉

PR #233, opened 2020-05-31
Build more release binaries

lunaryorn:
Include a static Linux binary with musl and a macOS build.  Also add readme, license and changelog to the release archives.  Closes GH-134.

I copied and adapted the workflow from one of my crates but I didn't test it; I just don't know how to do that with a forked repo.

I don't have access to macOS, but I'm sure it works, given that Windows also works.  I tested a musl build locally, but I didn't add it to the CI workflow.  Should I?

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -7,20 +7,22 @@ on:
 
 jobs:
 
-  create-binaries-windows:
+  create-binaries:
 
     strategy:
       matrix:
-        conf:
-          - x86_64-pc-windows-msvc
-          - i686-pc-windows-msvc
+        os: [ubuntu-latest, macOS-latest, windows-latest]
         include:
-          - conf: x86_64-pc-windows-msvc
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+          - os: macOS-latest
+            target: x86_64-apple-darwin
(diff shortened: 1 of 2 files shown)


shssoichiro:
Thank you, this is great!

If we're building release binaries for MacOS, I think it makes sense to also run the per-commit CI steps against MacOS. Just testing stable rust should be fine. As for testing the release, unfortunately I'll just have to push a tag to test it. I'm okay with doing that after this is merged, and can do a fixup commit if needed.

lunaryorn:
I'll add musl and macOS to the CI builds 👍

lunaryorn:
@shssoichiro I added Musl and macOS to the test build.  Since I didn't want to copy the whole build instructions I took the liberty to merge the windows and unix builds into a single parametrized build first :innocent:  But I see now that this broke your expected checks, since builds now have different names.  I'm sorry, if you like I can revert the change and add the new builds as copy of the Windows build.

shssoichiro:
No problem with changing the names, I think it makes sense to use the more verbose names. I just have to edit a repo setting to fix that.

lunaryorn:
You're welcome 🙏 Thank you for reviewing and merging this change, and for your work on this tool.

If these changes cause any trouble, please do ping me and I'll try my best to fix.

PR #258, opened 2020-07-13
Update readme for latest version

jakearchibald:
Although it feels like this should be automated.


shssoichiro:
Thanks! You're right, this definitely should be automated...

PR #318, opened 2020-09-28
Make `rayon` truly optional

Anhil-111:
This reduces the amount of dependencies of a project that directly depends only on `oxipng` by 26 (with `default-features = false`).


RReverser:
Oh, good call! I wonder if the resulting binary size is any different as a result - I'd assume all those unused functions wouldn't be linked anyway?

Anhil-111:
It's slightly smaller: 386.9 KiB vs 384.4 KiB (targeting `wasm32-unknown-unknown`, exposing a single function that calls `oxipng::optimize_from_memory` with its sole parameter and `oxipng::Options::max_compression()`).

shssoichiro (approved):

PR #319, opened 2020-09-28
Make `libdeflater` and `zopfli` optional

Anhil-111:
Had some issues with `oxipng` being a dependency when compiling to `wasm32-unknown-unknown` due to `libdeflater-sys`, found out that there's currently no way to get rid of it completely.


RReverser:
> Had some issues with `oxipng` being a dependency when compiling to `wasm32-unknown-unknown` due to `libdeflater-sys`, found out that there's currently no way to get rid of it completely.

What issues did you have? In fact, we use OxiPNG specifically with libdeflater in Squoosh. I suspect you might've run into Clang not having a stdlib for that target - if so, this has been already fixed upstream, I just need to integrate that fix into `libdeflater`. Meanwhile, please try to use Clang from [wasi-sdk](https://github.com/WebAssembly/wasi-sdk) instead of the system-wide one.

RReverser:
I wonder if it would be possible to get `miniz_oxide` optional as well? Then one could choose a single compressor to build the library with, and exclude all others.

Anhil-111:
> I suspect you might've run into Clang not having a stdlib for that target

Yep, that was the issue. Compiling it with Clang from wasi-sdk does work. I guess `libdeflater` does not make use of syscalls so using WASI build-tools shouldn't be a problem when targeting browsers, is that correct?

> I wonder if it would be possible to get `miniz_oxide` optional as well?

It should be possible. There's a bunch of options exclusive to `miniz_oxide`, making it optional would be a little bit more nuanced than what I did in this PR.

RReverser:
> I guess `libdeflater` does not make use of syscalls so using WASI build-tools shouldn't be a problem when targeting browsers, is that correct?

Yeah, that's right and the way we use it. It only needs stdlib for malloc, memcpy and few other self-hosted functions. In fact, you can even build it in a way that integrates with Rust's allocator API, and then actual allocation routines get deduplicated as well. You can find a shim example here: https://github.com/GoogleChromeLabs/squoosh/blob/dev/codecs/oxipng/src/malloc_shim.rs

Upstream there is now also a way to build libdeflate in a freestanding mode (no reliance on stdlib whatsoever), but it still needs a bit of work to integrate into the Rust wrapper (I have it half-done or so). Either way, the result should be the ~same.

RReverser:
@Anhil-111 UPD: PR for libdeflater is up, in case you'd want to try it out: https://github.com/adamkewley/libdeflater/pull/8

Anhil-111:
Just had a chance to test your branch, it seems to compile just fine, thanks.

shssoichiro:
Actually wanting this for myself also, to reduce the number of dependencies for a large project I have that depends on oxipng.

shssoichiro (approved):

PR #327, opened 2020-10-06
Switch to crossbeam-channel + rayon::spawn

RReverser:
So this PR might look a bit weird, but hopefully still acceptable. There are two things going on here:

1. A custom alternative to `std::thread::spawn` based on `rayon` + `crossbeam-channel`. See the comment on the function for motivation and more details, but TL;DR is that it makes possible to run OxiPNG with threads even on a Wasm target! (in engines that support Wasm threads). Working variant of Squoosh.app that makes use of this patch and runs OxiPNG+threads here: https://multithread--squoosh.netlify.app/editor
2. Since the part 1 requires `crossbeam-channel` anyway, I switched to it from `std::sync::mspc`, too, for the better performance as well as API improvements. For example, before this, evaluator had to use `sync_channel(4)` - a blocking synchronous channel with a magical constant maximum of 4 - because the async version of `std::sync::mpsc::Sender` can't be shared between threads (it doesn't implement `Sync` trait). However, with crossbeam it's not a problem, and it becomes possible to use asynchronous `unbounded` channel instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -164,12 +164,22 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
+version = "0.4.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b153fe7cbef478c567df0f972e02e6d736db11affe43dfc9c56a9374d1adfb87"
+dependencies = [
+ "crossbeam-utils 0.7.2",
+ "maybe-uninit",
+]
+
+[[package]]
+name = "crossbeam-channel"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dca26ee1f8d361640700bde38b2c37d8c22b3ce2d360e1fc1c74ea4b0aa7d775"
 dependencies = [
  "cfg-if 1.0.0",
- "crossbeam-utils",
+ "crossbeam-utils 0.8.0",
 ]
 
 [[package]]
@@ -180,7 +190,7 @@ checksum = "94af6efb46fef72616855b036a624cf27ba656ffc9be1b9a3c931cfc7749a9a9"
 dependencies = [
  "cfg-if 1.0.0",
  "crossbeam-epoch",
- "crossbeam-utils",
+ "crossbeam-utils 0.8.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


TPS:
@RReverser Strangely, _all_ of the associated tests have been running >3½ _hours_ (as of this note).  Such leads me to think _something_ isn't quite right.  Maybe check on that?

RReverser:
Yeah that does look odd. I'll look into it.

RReverser:
Ugh. Looks like some racy issue, I can't even reproduce it most of the time but only sometimes.

shssoichiro:
Coincidentally, I just was involved with [a PR that had a similar issue](https://github.com/rust-av/av-metrics/pull/168) on another project. The solution there was to have one additional thread available in the pool, to avoid a deadlock. I'm not entirely certain if this is the _same_ cause, but it's an idea to look at.

RReverser:
> I'm not entirely certain if this is the same cause, but it's an idea to look at.

Ah, that might be very likely. It crossed my mind at some point, but _seemed_ to work regardless, so I didn't bother.

RReverser:
Sounds a lot like https://github.com/rayon-rs/rayon/issues/743, too. I wish Rayon exposed a way to actually spawn a thread.

RReverser:
Okay, setting `RAYON_NUM_THREADS=1` helps a lot with the reproduction (everything is stuck immediately).

RReverser:
Okay, I think I found the most reliable solution: just avoid `std::thread::spawn` altogether. Since each `try_image` already spawns its own Rayon task and the caller had to already block on `.get_result()` until the evaluation thread has finished, on practice this doesn't sacrifice performance anyway.

In fact, adding logging shows that, when there is a sufficient number of threads, most of the time `unbounded` channel doesn't even buffer any messages up, because `.get_result()` is called before the threads even started execution. When there is only 1 thread for the entire pool, then it will buffer up all the candidates and compare them serially. This will require more memory than before the change, but it's an edge case and given that there is explicitly only 1 Rayon thread given to the app, I think it's an expected behaviour and an acceptable compromise.

WDYT?

RReverser:
By the way, testing with `RAYON_NUM_THREADS=1` also revealed that `verbose_mode` test gets stuck even on master 😬 Not strictly in scope of this issue, but probably worth fixing too.

RReverser:
Fixed `verbose_mode` test by doing `current_num_threads() + 1` instead (unlike with the library itself, we're in control of the rayon pool here, so just went with the simpler fix).

RReverser:
@shssoichiro Ping - any thoughts on this?

shssoichiro:
This seems good to me. Would you mind rebasing?

shssoichiro (approved):

RReverser:
Thanks for rebasing. It was on my list after seeing your comment, but a bit further away after other tasks.

PR #328, opened 2020-10-07
Use the new stabilised fetch_min

RReverser:
A minor simplification using the corresponding new standard library function.

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -19,7 +19,7 @@ jobs:
         - x86_64-apple-darwin
         toolchain:
         # Minimum stable
-        - "1.41.0"
+        - "1.45.0"
         - stable
         - beta
         - nightly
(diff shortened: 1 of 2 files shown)


RReverser:
Ah right, the Github action uses Rust 1.41. @shssoichiro is this an intended MSRV you'd like to keep? If so, feel free to close this PR.

shssoichiro:
I haven't had an official policy on minimum Rust version, I've simply been bumping it at needed. Although at the very least I've been bumping the minor version of oxipng when the Rust minimum version is bumped. I'm not sure if that is the best approach either--maybe a major version bump is better?

RReverser:
TBH I'm not sure. I guess minor is fine, since it already doesn't count as semver-range-compatible.

Personally, I've been rather lax with Rust version in some libraries and have been bumping only patch when using new features and not considering it a breaking change as long as it's available on stable channel, but then I know some more popular libs try to stick to a specific MSRV very strictly and I can see their point as well.

I guess I'm not the best person to ask 😅

shssoichiro (approved):

PR #360, opened 2020-12-11
initial implementation of timestamp preservation #166

### LINKED ISSUE #166: Option to preserve modification dates
I'd like to start off by saying I highly appreciate what this software offers, but there is this one feature I miss.

In jpegoptim you can specify [--preserve](https://www.mankier.com/1/jpegoptim#--preserve) to not change the modification time of the file being optimized.
Could a similar feature be implemented into oxipng? Would make at least one user happy. ^^

moschroe:
Implements preservation of input file timestamps (access and modification time). Fixes #166.

Added one dependency (`filetime`) to provide cross-plattform support for modifying timestamps.


TPS:
@moschroe In #318 & #319, there was a large amount of work done to make most dependencies optional.  Is it possible to do the same for `filetime`?

moschroe:
took the liberty of enabling the feature by default and making the help text for `-p` explicit, because I think it _is_ an important distinction to make

src/lib.rs
```diff
@@ -324,9 +324,30 @@ pub fn optimize(input: &InFile, output: &OutFile, opts: &Options) -> PngResult<(
...
+    // grab metadata before even opening input file to preserve atime
+    let opt_metadata_preserved;
     let in_data = match *input {
-        InFile::Path(ref input_path) => PngData::read_file(input_path)?,
+        InFile::Path(ref input_path) => {
+            if opts.preserve_attrs {
+                opt_metadata_preserved = input_path
+                    .metadata()
+                    .map_err(|err| {
+                        // TODO: Fail if input and output file are the same and metadata cannot be preserved?
```
shssoichiro, line 335: This is a very good question. My gut tells me that we should fail if the user wants to preserve metadata and they are unable to. Then the user can remove the preserve metadata flag if this is acceptable.

src/lib.rs
```diff
@@ -382,20 +403,27 @@ pub fn optimize(input: &InFile, output: &OutFile, opts: &Options) -> PngResult<(
...
-                ))
-            })?;
+            buffer
+                .write_all(&optimized_output)
+                // flush BufWriter so IO errors don't get swallowed silently on close() by drop!
+                .and_then(|()| buffer.flush())
+                .map_err(|e| {
+                    PngError::new(&format!(
+                        "Unable to write to {}: {}",
+                        output_path.display(),
```
shssoichiro, line 418: I think you don't need to explicitly specify `display()` here--by default, when passing an argument to `format` as a `{}` parameter, it will call the Display impl.

src/lib.rs
moschroe: implemented the more rigorous failure mode in 0001346, please review :)

shssoichiro:
Sorry that I lost track of this for such a long time. The contribution is appreciated!

PR #369, opened 2021-01-09
Make cloudflare-zlib dependency optional

AlexTMjugador:
Although the Cloudflare fork of zlib may introduce performance improvements, there are issues building it with MinGW toolchains on Windows, as evidenced by this [StackOverflow question](https://stackoverflow.com/questions/61170244/gcc-exe-fail-when-building-cloudflare-zlib-sys-rs).

However, it is not always possible to switch to the MSVC toolchain on Windows, because that may break other native C libraries which have trouble with MSVC instead, and just compiling to a 32-bit target can get pretty messy too.

[I've opened an issue](https://gitlab.com/kornelski/cloudflare-zlib-sys/-/issues/1) to get this error fixed upstream, in the cloudflare-zlib-sys crate, but several months have passed with no answer.

Therefore, make the Cloudflare zlib dependency optional, so users will have to explicitly opt in for it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -59,7 +59,8 @@ default-features = false
 features = ["png"]
 version = "0.23"
 
-[target.'cfg(any(target_arch = "x86_64", target_arch = "aarch64"))'.dependencies.cloudflare-zlib]
+[dependencies.cloudflare-zlib]
+optional = true
 features = ["arm-always"]
 version = "^0.2.2"
 
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -18,7 +18,7 @@ pub mod cfzlib;
 #[cfg(not(feature = "cloudflare-zlib"))]
 pub mod cfzlib {
     pub fn is_supported() -> bool {
-        return false;
+        false
     }
 }
 
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -12,10 +12,10 @@ mod deflater;
 #[cfg(feature = "libdeflater")]
 pub use deflater::deflate as libdeflater_deflate;
 
-#[cfg(any(target_arch = "x86_64", target_arch = "aarch64"))]
+#[cfg(feature = "cloudflare-zlib")]
 pub mod cfzlib;
 
-#[cfg(not(any(target_arch = "x86_64", target_arch = "aarch64")))]
+#[cfg(not(feature = "cloudflare-zlib"))]
 pub mod cfzlib {
     pub fn is_supported() -> bool {
         return false;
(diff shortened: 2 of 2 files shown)


kattjevfel:
Why opt-in if it's only broken on one OS with one config?

AlexTMjugador:
Because AFAIK there is no nice way to tell Cargo to enable that feature by default on only certain platforms. Moreover, in case the cloudflare-zlib-sys crate gets updated to support more platforms, end users will inmediately be able to use that improvement once OxiPNG updates its dependency on it, with minimal maintenance needed by OxiPNG for that.

AlexTMjugador:
Good news: a nice way to implement platform-specific features got stabilized recently, and is underway to the next stable rust version. See https://doc.rust-lang.org/nightly/cargo/reference/features.html#feature-resolver-version-2 and https://github.com/rust-lang/cargo/pull/8997. This means that it should be possible to improve this PR soon in a way that it automatically excludes that feature for problematic platforms, but is this actually in scope for OxiPNG? To be honest, it seems like I'm really the only person that cares about the mess that is combining OxiPNG with other random C libraries that do not play nicely with the MSVC toolchain on WIndows.

shssoichiro:
As it is, I don't like that this PR disables cloudflare-zlib by default. However, now that Rust 1.50 is out, I'd like to see what we can achieve with that new approach. Or is that feature not out until 1.51?

AlexTMjugador:
It'll land on 1.51, as seen [here](https://github.com/rust-lang/rust/pull/81917/files#diff-4a00447b47c36ebd3ded77321deff288e424f4605496ee1199fc9953a20f4473R81). Let's wait until 1.51 is released for this PR to continue.

AlexTMjugador:
I've been looking into this lately and, instead of working around cloudflare-zlib portability problems by adding features to the crate or bumping MSRV and using platform-specific hacks, maybe it's a better idea to drop it altogether in favor of libz-sys with zlib-ng.

[zlib-ng](https://github.com/zlib-ng/zlib-ng#readme) is yet another fork of zlib that integrates both Cloudflare and Intel patches to zlib. Performance wise, it [trades blows with the Cloudflare fork](https://github.com/zlib-ng/zlib-ng/issues/42#issuecomment-675293711), and even [wins](https://github.com/Byron/gitoxide/issues/1#issuecomment-672626465) depending on the benchmark, but it has the advantage that is more portable. I've recently contributed zlib-ng support for the spng crate and it provided a 1,85x speedup for image decoding compared to stock zlib, and it's [allegedly faster than miniz_oxide anyway](https://internals.rust-lang.org/t/psa-libz-sys-flate2-and-git2-now-support-the-high-performance-zlib-ng/12924/3).

Would investigating the replacement of cloudflare-zlib with zlib-ng be a good thing for OxiPNG? I think it looks promising.

AlexTMjugador:
Days ago, version 0.3.0 of `cloudflare-zlib-sys` was released, which includes [these upstream changes](https://github.com/cloudflare/zlib/pull/25) that fix compilation issues on Windows MinGW platforms. It's enough to run `cargo update` to make OxiPNG use this new version. This is a nice fix for the underlying issue, so the problem this PR tried to solve doesn't exist anymore (at least for this platform), and I'm closing it.

PR #377, opened 2021-01-15
Avoid asserts panicking on broken files

kornelski:
This rejects malformed files gracefully instead of panicking in random places.


shssoichiro (approved):
Looks good, thanks!

PR #385, opened 2021-02-04
Simplified unfilter code

kornelski:


shssoichiro:
I think this looks good, but it's failing on rustfmt. Would you mind running `cargo fmt` and pushing? Thanks!

kornelski:
No, sorry. rustfmt makes code looks worse. I don't think it's a good tool, and I really don't like when it ruins CI because it's too inflexible and has bad heuristics.

PR #400, opened 2021-05-05
Optimize unfiltering

kornelski:
The previous implementation could also panic on truncated data. This one is more resilient.


kornelski:
@shssoichiro ping

shssoichiro:
I like the optimization changes, I'm going to give feedback on the clippy lints PR first since it looks like this is dependent on that one.

shssoichiro (approved):

PR #401, opened 2021-05-07
Fix clippy lints

kornelski:


shssoichiro:
I'm a little confused as to what is being addressed here--if I run cargo clippy, I only see a couple of issues. I do understand the use of `#[cold]` on the error handling functions, but I'm not sure the benefit of adding `#[must_use]` everywhere. My understanding of `#[must_use]` is that it's useful for structures like `Result` and `Iterator` which are lazy and do nothing unless they are used (i.e. your code will do unexpected things if you call it and don't use the result). I don't think that's the case for these instances.

kornelski:
It addresses the fact that your CI has started using a newer version of clippy, which started complaining about more things that are already present in the code on the master branch, so currently all pull requests are rejected because of pre-existing clippy issues.

shssoichiro:
Interesting, I just bumped to the latest rust nightly and I'm still not seeing the must_use lints. Is [this](https://rust-lang.github.io/rust-clippy/master/#must_use_candidate) the one you're referring to? It looks like it's set to off by default (enabled at "pedantic" level) and may have a lot of false positives? (I do like the `#[cold]` additions, they seem to be in reasonable places as well.)

kornelski:
Ok, I may have gone overboard with the pedantic setting.

Clippy in CI complains about this:


> GitHub Actions / test (x86_64-unknown-linux-gnu, stable)
> src/rayon.rs#L13
> manual implementation of `Option::map`

shssoichiro:
Would you mind pulling out the `#[must_use]` changes? I'm fine with everything else that's in this PR. Thanks!

kornelski:
done

shssoichiro (approved):

PR #408, opened 2021-06-25
update dependency

remilauzier:


shssoichiro:
Thanks, however I'll normally handle this immediately before pushing a new release. Which I'm going to do shortly.

There's also no need to update the Cargo.toml unless there's a breaking version update, just running `cargo update` will update the Cargo.lock file with any non-breaking dependency updates.

PR #409, opened 2021-06-25
Add option to skip grayscale reduction of RGB(A) images

AlexTMjugador:
For RGB(A) images that contain gray colors, this reduction can achieve significant space savings. However, in the absence of gamma correction data, some PNG decoders assume more exotic color spaces for grayscale images instead of the ubiquitous sRGB. This results in gamma miscorrection, and for the end user this means that colors will look wrong, like "washed-out". For example, Java's ImageIO class, which is popular in the JVM world to read PNG files, uses rather unconventional defaults, as explained in this StackOverflow question: https://stackoverflow.com/questions/31312645/java-imageio-grayscale-png-issue

Gamma miscorrection problems aside, OxiPNG currently tries hard to reduce RGB(A) images to grayscale, because it expects that reduction to be quite effective. However, in some cases, OxiPNG generates smaller PNG files when reducing grasycale RGB(A) images to paletted color than actual grayscale color. For example, let's say that `~/gray.png` is a 256x256 RGBA image entirely filled with (119, 119, 119, 255) pixels. OxiPNG, by default, reduces this image to grayscale and achieves a 68.23% decrease:

```
$ cargo build --release && target/release/oxipng -omax --out ~/out.png ~/gray.png
Processing: /home/user/gray.png
    256x256 pixels, PNG format
    4x8 bits/pixel, RGBA
    IDAT size = 604 bytes
    File size = 661 bytes
Reducing image to 1x4 bits/pixel, Grayscale
Trying: 144 combinations
Found better combination:
    zc = 6  zs = 0  f = 0        153 bytes
    IDAT size = 153 bytes (451 bytes decrease)
    file size = 210 bytes (451 bytes = 68.23% decrease)
Output: /home/user/out.png
```

However, if the `--ng` option that this commit adds is used to skip the grayscale reduction step, OxiPNG reduces to a single color palette instead, which is much more efficient, achieving a 84.42% decrease:

```
$ cargo build --release && target/release/oxipng -omax --ng --out ~/out.png ~/gray.png
Processing: /home/user/gray.png
    256x256 pixels, PNG format
    4x8 bits/pixel, RGBA
    IDAT size = 604 bytes
    File size = 661 bytes
Reducing image to 1 bits/pixel, 1 colors in palette
Trying: 144 combinations
Found better combination:
    zc = 3  zs = 3  f = 0        31 bytes
    IDAT size = 31 bytes (573 bytes decrease)
    file size = 103 bytes (558 bytes = 84.42% decrease)
Output: /home/user/out.png
```

While OxiPNG should arguably be made smarter to better handle these cases, adding an option to manually skip that grayscale reduction can't hurt. In fact, it may even help users achieving the most out of current versions of OxiPNG, and developers reasoning about what makes a grayscale-like RGB(A) image compress better with a color palette.

Due to the reasons stated above, this adds a simple "grayscale_reduction" option to the Options struct, and a "no-grayscale-reduction" command line switch, that makes OxiPNG not try this problematic grayscale reduction on RGB(A) images.

### DIFF
--- a/benches/reductions.rs
+++ b/benches/reductions.rs
@@ -80,7 +80,7 @@ fn reductions_rgba_to_rgb_16(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from("tests/files/rgba_16_should_be_rgb_16.png"));
     let png = PngData::new(&input, false).unwrap();
 
-    b.iter(|| reduce_color_type(&png.raw));
+    b.iter(|| reduce_color_type(&png.raw, true));
 }
 
 #[bench]
@@ -88,7 +88,7 @@ fn reductions_rgba_to_rgb_8(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from("tests/files/rgba_8_should_be_rgb_8.png"));
     let png = PngData::new(&input, false).unwrap();
 
-    b.iter(|| reduce_color_type(&png.raw));
+    b.iter(|| reduce_color_type(&png.raw, true));
 }
 
 #[bench]
@@ -98,7 +98,7 @@ fn reductions_rgba_to_grayscale_alpha_16(b: &mut Bencher) {
     ));
     let png = PngData::new(&input, false).unwrap();
 
-    b.iter(|| reduce_color_type(&png.raw));
+    b.iter(|| reduce_color_type(&png.raw, true));
 }
 
 #[bench]
@@ -108,7 +108,7 @@ fn reductions_rgba_to_grayscale_alpha_8(b: &mut Bencher) {
     ));
     let png = PngData::new(&input, false).unwrap();
 
-    b.iter(|| reduce_color_type(&png.raw));
+    b.iter(|| reduce_color_type(&png.raw, true));
 }
 
 #[bench]
@@ -118,7 +118,7 @@ fn reductions_rgba_to_grayscale_16(b: &mut Bencher) {
     ));
     let png = PngData::new(&input, false).unwrap();
 
-    b.iter(|| reduce_color_type(&png.raw));
+    b.iter(|| reduce_color_type(&png.raw, true));
 }
 
 #[bench]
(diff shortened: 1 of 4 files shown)


AlexTMjugador:
A quick note: this PR introduces a major change, because code that instantiates the `Options` struct directly by providing values for every field will no longer compile. To mitigate the impact of these changes in the future, I can add a `#[non_exhaustive]` trait to the struct, which forces to default-initialize fields even if values for all fields are provided, and to account for extra fields in match expressions, in code that uses that struct from another crate. Let me know if that's a desirable change!

AlexTMjugador:
On a related note, I guess I could look into trying both grayscale and palette color reductions in the `reduce_color_type` method that this PR modifies, as explained in https://github.com/shssoichiro/oxipng/issues/195#issuecomment-622968270, to let OxiPNG pick the one that results in smaller IDAT and PLTE chunks. That would hurt performance a bit, but also improve compression by a large margin in some cases.

But for now let's see if there's some interest on the changes that this PR introduces, or if someone has some better ideas :)

shssoichiro:
Thanks! Given that there is already a "no-palette-reduction" I think it makes sense to add this as well. I think having an automated check for whether palette or grayscale is better is also a good idea, but I'm going to merge this PR and the automated check can be added later.

andrews05:
@AlexTMjugador Should the `--ng` option also prevent conversion *from* grayscale? (e.g. to indexed)

AlexTMjugador:
> Should the `--ng` option also prevent conversion _from_ grayscale? (e.g. to indexed, which has been added since this PR)

Yes, the grayscale to indexed color conversion should be avoided to not trigger the weird default color profiles assumed by Java's `ImageIO`. Alternatively, a sRGB chunk could be added, but my expectation is that for small images the chunk overhead would be greater than the savings achieved by other color formats.

PR #413, opened 2021-07-26
Fix some warnings

remilauzier:


src/lib.rs
```diff
@@ -1,18 +1,28 @@
...
-#![allow(clippy::cognitive_complexity)]
-#![allow(clippy::upper_case_acronyms)]
+#![deny(
+    rust_2018_idioms,
+    rust_2018_compatibility,
+    future_incompatible,
+    unused,
+    nonstandard_style
+)]
+#![deny(clippy::all, clippy::pedantic)]
```
shssoichiro, line 10: I'm personally not a big fan of globally enabling `clippy::pedantic`, given that there are several lints in that group that either have a lot of false positives or are generally just not very useful. I see that you've set some of them to `allow`, but I'd prefer to instead pick out the lints from that group that we like and specifically enable those.

shssoichiro: Re: `deny(clippy::all)`: I'm not against setting specific lints to `deny` that we really, never ever want to be allowed, but even though CI will fail if there are clippy warnings (IIRC), the reason I wouldn't want to `deny(clippy::all)` is because this will likely break compilation when new versions of Clippy are released with new lints in the `all` group, including for downstream packages which use oxipng as a library. I want old versions to still be compilable, even if they may emit warnings.

shssoichiro (changes requested):
Thanks! I've left some feedback, I'm open to discussion if you think there's a compelling reason to proceed this way.

remilauzier:
I think I have refactored it better this time. Some Clippy warnings could be denied, especially those concerning floats, but you should know better than me, so I put them all in warning for now. I think I have put those that matter, but feel free to correct me.

remilauzier:
The msrv configuration in clippy.toml should reduce the risk of breaking thing since it deactivate a few things that arrive later in time. And by default Clippy tried to avoid breaking public API too.

PR #425, opened 2021-11-15
apply_preset_5 and apply_preset_6 to set correct compression levels.

### LINKED ISSUE #424: -o 3 gives better compression than -o 6/max
I'm running version 5.0.0 on Windows. Consider this file [here](https://purepng.com/public/uploads/large/purepng.com-pinapplepinapplepen-pinapple-apple-penfruitvegetableheathlyfoodcliparttropical-331522772242aurac.png), which I have downloaded and renamed to `pineapple.png`.

Running `oxipng -o 3 -P .\pineapple.png` gives:

```
Processing: .\pineapple.png
    1379x2957 pixels, PNG format
    4x8 bits/pixel, RGBA
    IDAT size = 6484198 bytes
    File size = 6486971 bytes
Trying: 24 combinations
Found better combination:
    zc = 9  zs = 0  f = 3        6429034 bytes
    IDAT size = 6429034 bytes (55164 bytes decrease)
    file size = 6431807 bytes (55164 bytes = 0.85% decrease)
Running in pretend mode, no output
``` 

Running `oxipng -o 6 -P .\pineapple.png` gives:

```
Processing: .\pineapple.png
    1379x2957 pixels, PNG format
    4x8 bits/pixel, RGBA
    IDAT size = 6484198 bytes
    File size = 6486971 bytes
Trying: 144 combinations
Found better combination:
    zc = 8  zs = 0  f = 3        6440096 bytes
    IDAT size = 6440096 bytes (44102 bytes decrease)
    file size = 6442869 bytes (44102 bytes = 0.68% decrease)
Running in pretend mode, no output
``` 

As you can see, `-o 3` achieves a size reduction of 0.85%, while `-o 6` only gets to 0.68%, even if the expectation is that it should at least be as good as `-o 3`.

Checking the verbose output of the two runs with `-v`, I noticed that `-o 6` does not try combinations with zc = 9, help if the help page says it should. `-0 3` uses zc = 9 exclusively and that seems to give it an edge.

elnigno:
The for loop in apply_preset_5 was not including 9.
apply_preset_6 now builds now calls apply_preset_4 instead of
apply_preset_5, and adds all compression levels from 1 to 9.

I have not written unit tests for this, yet.

fix #424

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -263,7 +263,7 @@ impl Options {
     fn apply_preset_5(mut self) -> Self {
         if let Deflaters::Zlib { compression, .. } = &mut self.deflate {
             compression.clear();
-            for i in 3..9 {
+            for i in 3..10 {
                 compression.insert(i);
             }
         }
@@ -273,11 +273,11 @@ impl Options {
     fn apply_preset_6(mut self) -> Self {
         if let Deflaters::Zlib { compression, .. } = &mut self.deflate {
             compression.clear();
-            for i in 1..3 {
+            for i in 1..10 {
                 compression.insert(i);
             }
         }
-        self.apply_preset_5()
+        self.apply_preset_4()
     }
 }
 


shssoichiro (approved):
Thanks, good find!

PR #434, opened 2022-02-01
Bump image crate to 0.24

paolobarbolini:
At least when running locally one of the tests broke. Maybe it's a `png` crate regression, since it updated from 0.16 to 0.17?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,12 +189,11 @@ dependencies = [
 
 [[package]]
 name = "deflate"
-version = "0.8.6"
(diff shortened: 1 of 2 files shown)


shssoichiro:
Odd that one of the tests is breaking, and still seems to be... especially since they pass on my local machine. Thanks for bringing this up, I'll see if I can figure out what's going on with this test... I'll merge this as well after the test issue is resolved (bumping crates is typically not a bad idea).

shssoichiro:
Superseded by #445

PR #439, opened 2022-05-02
Add check option

carbotaniuman:
This allows users to make sure Oxipng can parse a PNG file before commiting to an optimization.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,6 +131,10 @@ pub struct Options {
     ///
     /// Default: `false`
     pub fix_errors: bool,
+    /// Don't actually run any optimizations, just parse the PNG file.
+    ///
+    /// Default: `false`
+    pub check: bool,
     /// Don't actually write any output, just calculate the best results.
     ///
     /// Default: `false`
@@ -299,6 +303,7 @@ impl Default for Options {
 
         Options {
             backup: false,
+            check: false,
             pretend: false,
             fix_errors: false,
             force: false,
@@ -362,6 +367,11 @@ pub fn optimize(input: &InFile, output: &OutFile, opts: &Options) -> PngResult<(
     };
 
     let mut png = PngData::from_slice(&in_data, opts.fix_errors)?;
+
+    if opts.check {
+        info!("Running in check mode, not optimizing");
+        return Ok(());
+    }
 
     // Run the optimizer on the decoded PNG.
     let mut optimized_output = optimize_png(&mut png, &in_data, opts, deadline)?;
--- a/src/main.rs
+++ b/src/main.rs
@@ -83,6 +83,10 @@ fn main() {
             .help("Preserve file attributes if possible")
             .short('p')
             .long("preserve"))
+        .arg(Arg::new("check")
+            .help("Do not run any optimization passes")
+            .short('c')
+            .long("check"))
         .arg(Arg::new("pretend")
             .help("Do not write any files, only calculate compression gains")
             .short('P')
@@ -413,6 +417,10 @@ fn parse_opts_into_struct(
 
     if matches.is_present("fix") {
         opts.fix_errors = true;
+    }
+
+    if matches.is_present("check") {
+        opts.check = true;
     }
 
     if matches.is_present("pretend") {


Tristan971:
We'd like that very much; currently using --pretend -o0 as a validation step (proper optimization is executed asynchronously later) and it's still multiple seconds at times on particularly unfriendly files on the synchronous path, which is causing timeouts for our users.

Here is [one such file for reference](https://user-images.githubusercontent.com/875533/166309761-236f4370-b2ec-410c-bf5f-531dbcf8717b.png), seems to happen mostly on very high-definition B&W artwork.

shssoichiro (approved):
Thanks!

andrews05:
Hi @carbotaniuman @Tristan971, we've recently been discussing this option and whether we really need to keep it, as it seems very unusual. Could you explain a little more why you need to check that oxipng can parse the file beforehand?

Note that you actually can perform an "equivalent" check-oriented run with `--pretend --nx --nz -i keep`, which will disable all optimisations and transformations.

Tristan971:
Our use-case is to process PNG files uploaded by users asynchronously, but provide invalid file detection synchronously (so the user can correct their file).

But for the synchronous part, we must be fast to evaluate validity, as it's part of handling a user's HTTP request. And said request can contain more than 1 file, with each file potentially quite big, leading to unacceptable response times if we actually try any optimisation while handling it.

At the time this issue/PR were raised, `--pretend` alone was the best we knew of, and still around 10x slower (compared to `--check`), hence this original issue/PR.

--check:
```
$ hyperfine -n 10 'oxipng --check vertical.png'
Benchmark 1: 10
  Time (mean ± σ):      23.3 ms ±   0.5 ms    [User: 16.3 ms, System: 3.0 ms]
  Range (min … max):    22.3 ms …  26.7 ms    109 runs
```

--pretend:
```
$ hyperfine -n 10 'oxipng --pretend vertical.png'
Benchmark 1: 10
  Time (mean ± σ):      1.126 s ±  0.003 s    [User: 1.368 s, System: 0.023 s]
  Range (min … max):    1.122 s …  1.132 s    10 runs
```

--pretend --nx --nz -i 0: (no -i keep in the version I have locally, and -i1 is slow, so I bench'd -i0):
```
$ hyperfine -n 10 'oxipng --pretend --nx --nz -i 0 vertical.png'
Benchmark 1: 10
  Time (mean ± σ):      26.2 ms ±   0.5 ms    [User: 16.5 ms, System: 5.1 ms]
  Range (min … max):    25.4 ms …  28.4 ms    105 runs

```

So with that new combination of args we would be okay without `--check` anymore.

andrews05:
Thanks for the info!
Apologies, `-i keep` is new and hasn't found its way into a release yet - in the current v8 release you should omit the `-i` flag altogether but in the next release you will need `-i keep`.

I'm still curious though, what kind of invalid files do you encounter? Are they broken/damaged pngs, or simply not pngs at all? Or are they pngs that appear valid but oxipng doesn't handle for some reason?

Tristan971:
> I'm still curious though, what kind of invalid files do you encounter? Are they broken/damaged pngs, or simply not pngs at all?

We do magic bytes checking before feeding anything to oxipng, so at the very least all such cases are PNG-looking files. Just broken one (and the occasional maliciously crafter file that has the PNG bytes but isn't a PNG file, but that's much rarer).

> Or are they pngs that appear valid but oxipng doesn't handle for some reason?

So far, we haven't found a situation where oxipng didn't support a file that was valid, so no worries on that end. And we process around 350k PNG files per month.

andrews05:
Cool, good to know. We'll probably go ahead and remove `--check` then. The other alternative for this would be to use a separate tool such as `pngcheck`.

Tristan971:
Gotcha. We'd prefer to not have to introduce an extra tool, but if it is the decision of oxipng that effort-less passes are out of scope, we'll do so.

carbotaniuman:
For context, the reluctance to use another tool is partially due to `pngcheck` being written in C, and also due to a potential mismatch between what OxiPNG supports and what `pngcheck` might consider valid, due to the complexity of the PNG standard and possible implementation divergence. It's not the end of the world, but I'd prefer at least the workaround being kept.

andrews05:
For sure, there's no intention of removing those other flags that disable everything. In fact if you find these aren't actually disabling everything, let us know as it's probably an oversight.

andrews05:
Check has now been removed in the v9 release. Note that `--nx` now infers `-i keep`, so only need `--pretend --nx --nz` for a fully non-optimising run.

Tristan971:
Thanks for letting us know 👍

ace-dent:
@andrews05 - sorry to necro post, can I check, does `--pretend --nx --nz` still validate *all* the same aspects of a png file? (i.e. deflate stream, filters, chunk CRC's, etc). ... or is there some advantage to using `-o0`?

andrews05:
Yup, it will still do a full run of parsing the png and even reconstructing a new one, it just won't do any transformations to the IDAT (other than combining multiple IDATs into one).

ace-dent:
@andrews05 - Awesome. Thanks! 🙏

PR #441, opened 2022-07-02
Update stderrlog dependency to v0.5.1

flother:
Oxipng has an optional dependency on [stderrlog](https://github.com/cardoe/stderrlog-rs), which itself has a dependency on [thread_local](https://github.com/Amanieu/thread_local-rs/issues/33). Versions of thread_local < 1.1.4 have a [vulnerability caused by a data race in `Iter` and `IterMut`](https://rustsec.org/advisories/RUSTSEC-2022-0006.html). Updating to stderrlog v0.5.1 means that Oxipng will use a patched version of thread_local.


shssoichiro:
Thanks!

PR #445, opened 2022-08-19
Update and optimize dependencies

AlexTMjugador:
These changes update the dependencies to their latest versions, fixing some known issues that prevented doing so in the first place (see https://github.com/shssoichiro/oxipng/pull/434 and https://github.com/shssoichiro/oxipng/pull/437#issuecomment-1220968867).

In addition, the direct dependency on byteorder was dropped in favor of stdlib functions that have been stabilized for some time in Rust, and the transitive dependency on chrono, pulled by stderrlog, was also dropped, which had been affected by security issues and improperly maintained in the past:

- https://github.com/cardoe/stderrlog-rs/issues/31
- https://www.reddit.com/r/rust/comments/ts84n4/chrono_or_time_03/

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -19,7 +19,7 @@ jobs:
           - x86_64-apple-darwin
         toolchain:
           # Minimum stable
-          - "1.56.1"
+          - "1.57.0"
           - stable
           - beta
           - nightly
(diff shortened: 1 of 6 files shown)


AlexTMjugador:
I'd like to mention that the checks the GitHub PR UI marks as expected do no longer exist due to the MSRV bump, so they won't ever be completed. Every check that exists has passed, though.

shssoichiro (approved):
Thanks for solving the image dep issue! And the rest of the cleanup. :slightly_smiling_face:

PR #446, opened 2022-09-03
Add initial support for changing Zopfli iterations 

AlexTMjugador:
PR https://github.com/shssoichiro/oxipng/pull/445 did some dependency updates, which included using the latest zopfli version. The latest version of this crate exposes new options in its API that allow users to choose the desired number of Zopfli compression iterations, which may greatly affect execution time. In fact, other optimizers such as zopflipng dynamically select this number depending on the input file size (see: https://github.com/shssoichiro/oxipng/issues/414).

As a first step towards making OxiPNG deal with Zopfli better, let's add the necessary options for libraries to be able to choose the number of iterations. This number is still fixed to 15 as before when using the CLI.

This PR is meant to be merged after https://github.com/shssoichiro/oxipng/pull/445.

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -19,7 +19,7 @@ jobs:
           - x86_64-apple-darwin
         toolchain:
           # Minimum stable
-          - "1.56.1"
+          - "1.57.0"
           - stable
           - beta
           - nightly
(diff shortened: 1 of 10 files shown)


shssoichiro (approved):
Thanks!

PR #450, opened 2022-09-07
Fix #449 by bumping dependencies. Update changelog and prepare for 6.0.1 release

### LINKED ISSUE #449: Zopfli compression in 6.0.0 results in a corrupted image in some cases
Using the newly-released 6.0.0 version on Windows 11 (64-bit) on the attached PNG results in the following:

`> oxipng -Z -o max cover.png`

```powershell
Processing: cover.png
    2981x2981 pixels, PNG format
    3x8 bits/pixel, RGB
    IDAT size = 6087921 bytes
    File size = 6088031 bytes
Trying: 6 combinations
Found better combination:
    zc = 0  zs = 0  f = 4        5897004 bytes
    IDAT size = 5897004 bytes (190917 bytes decrease)
    file size = 5897114 bytes (190917 bytes = 3.14% decrease)
The resulting image is corrupted and will not be outputted.
This is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues
The resulting image is corrupted
```

As per instructions, here I am. 

The problematic image is available at https://user-images.githubusercontent.com/1986890/188520158-84a1a201-a63f-4884-b880-19980932551b.png

FWIW, I had just successfully compressed this image (same command) with `oxipng` 5.0.1 a few days ago.  

The `-o max` flag doesn't make a difference, either (same error), but I don't get the error in normal mode (it can't achieve better compression and doesn't even try, I guess).

AlexTMjugador:
The 0.7.0 version of the `zopfli` crate has a regression that could cause invalid DEFLATE streams to be generated. The cause of that regression has been identified and corrected, and the fix was tested. A new patch version of that crate was published with the fix.

OxiPNG needs that fix to generate proper PNGs when the Zopfli mode is enabled, so let's update dependencies again. I've modified the changelog and bumped the OxiPNG version to 6.0.1 to make it as easy to deploy as possible :tada:

Fixes #449.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-### Next
+### Version 6.0.1
 
 - [Bugfix] Fix an issue where Zopfli mode could generate corrupt images
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+### Next
+
+- [Bugfix] Fix an issue where Zopfli mode could generate corrupt images
+
 ### Version 6.0.0
 
 - [Breaking] Bump minimum Rust version to 1.57.0
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -185,15 +185,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "deflate"
-version = "1.0.0"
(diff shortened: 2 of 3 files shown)


shssoichiro (approved):
Thanks! Easy is great. :slightly_smiling_face:

PR #453, opened 2022-10-07
Avoid printing output bytes on decompress error

### LINKED ISSUE #452: OxiPNG outputs the whole file in error message on decompression failure
Using 6.0.1

```
$ oxipng src/test/resources/samples/12.rawkuma.com.png 2>&1 | head --bytes=256
Processing: src/test/resources/samples/12.rawkuma.com.png
Error on decompress: DecompressError { status: Adler32Mismatch, output: [0, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 25%
```

Obviously truncated here, it does go on and on in practice

Sample image: [sample.png](https://user-images.githubusercontent.com/875533/192359609-bbaeab2b-8f0f-484c-8f8c-908bd33f3913.png)

edit: by the looks of it, it's the fault of miniz_oxide that the entire file is in the error message, but I suppose oxipng could/should filter out the output field with a custom display? idk.

Tristan971:
Hi, this is (one of many) ways to fix #452.

This seemed the least intrusive as we can't specifically override the Display/Debug implementations without creating our own type and mapping to it first.

Also, I chose to preserve part of the "decompressed so far" information, but just the byte count since it's mostly interesting in special cases (0, 1, some power of 2, etc).

Unfortunately, this means that one does lose information like a non-specific number of bytes all being 0 for example. I'm not sure how to reconcile that without making more intrusive changes (returning a struct rather than a formatted string, and handling the formatting somewhere with access to the current verbosity level perhaps).

### DIFF
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -27,8 +27,13 @@ pub mod cfzlib {
 
 /// Decompress a data stream using the DEFLATE algorithm
 pub fn inflate(data: &[u8]) -> PngResult<Vec<u8>> {
-    miniz_oxide::inflate::decompress_to_vec_zlib(data)
-        .map_err(|e| PngError::new(&format!("Error on decompress: {:?}", e)))
+    miniz_oxide::inflate::decompress_to_vec_zlib(data).map_err(|e| {
+        PngError::new(&format!(
+            "Error on decompress: {:?} (after {:?} decompressed bytes)",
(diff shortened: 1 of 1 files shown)


shssoichiro:
Thanks, looks good!

PR #455, opened 2022-10-21
Allow setting libdeflate compression level

### LINKED ISSUE #454: Allow setting libdeflate compression level
Using the `-D` option to switch to libdeflate appears to default to the highest compression level (12). This is relatively slow for very little gain over some lower levels (say, 10). It would be helpful to be able to set the level with the same `--zc` option as zlib.

As a side note, I ran a benchmark of `-o 1 -D` vs `-o 3`. The result was (coincidentally) identical total savings over my test set, but the libdeflate one performed over twice as fast. Some experiments may be warranted, but I think with a level of 10 it could be both smaller and faster than zlib, potentially replacing it as the default compressor.

andrews05:
This PR enables the `--zc` option with range 1-12 when using libdeflater (`-D`). When `-D` is specified without `--zc`, the level will default to 12 (the same level it was fixed to previously).
Libdeflater is also updated to v0.11.0.

Additionally this fixes a minor issue where a non-optimal result may be achieved with libdeflate due to lack of slack space (see [libdeflate.h#L92](https://github.com/ebiggers/libdeflate/blob/7eb94ca1fa8c72663938d0fa05427e8f68c0cc87/libdeflate.h#L92)).

Thoughts on implementation:
It may be better to make the level an optional value for the `-D` argument rather than reusing `--zc`, since they have different ranges and the levels in the presets currently only apply to zlib.

Thoughts for future:
Libdeflater can outclass zlib in both speed and ratio. It might be nice to just remove zlib and make libdeflater the default.

Closes #454.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -281,18 +281,18 @@ checksum = "8371e4e5341c3a96db127eb2465ac681ced4c433e01dd0e938adbef26ba93ba5"
 
 [[package]]
 name = "libdeflate-sys"
-version = "0.10.0"
+version = "0.11.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43afa5b192ff058426ba20a4f35c290ef402478d6045ac934ac15aa947a3898d"
+checksum = "cb6784b6b84b67d71b4307963d456a9c7c29f9b47c658f533e598de369e34277"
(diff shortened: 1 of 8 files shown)


andrews05:
Finally got there! Only took half a dozen failures 😅

PR #456, opened 2022-10-22
Fix level 0

andrews05:
Preset level 0 was previously using the default (preset 2) filters and strategies, instead of what it says in the help. This PR fixes it by extending from preset 1.


shssoichiro:
Thanks!

Issue #167, opened 2019-02-17
Corrupted output using large (24000x24000) grayscale image


potentcocktail:
I'm trying to optimize a large (24000x24000) grayscale image.  I use oxipng all the time for images of this size, so I was surprised when I got the following output:

andy@sandy:~$ oxipng -s -Z --out output.png "f=000000 t=0.0.png"
Processing: f=000000 t=0.0.png
    24000x24000 pixels, PNG format
    1x8 bits/pixel, Grayscale
    IDAT size = 32341177 bytes
    File size = 32353257 bytes
Trying: 2 combinations
Found better combination:
    zc = 0  zs = 0  f = 5        14173342 bytes
    IDAT size = 14173342 bytes (18167835 bytes decrease)
    file size = 14173473 bytes (18179784 bytes = 56.19% decrease)
The resulting image is corrupted and will not be outputted.
This is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues
The resulting image is corrupted

I also tried with -o 4 instead of Zopfli. Same result:

andy@sandy:~$ oxipng -o 4 -s --out output.png "f=000000 t=0.0.png"
Processing: f=000000 t=0.0.png
    24000x24000 pixels, PNG format
    1x8 bits/pixel, Grayscale
    IDAT size = 32341177 bytes
    File size = 32353257 bytes
Trying: 24 combinations
Found better combination:
    zc = 9  zs = 0  f = 1        15640968 bytes
    IDAT size = 15640968 bytes (16700209 bytes decrease)
    file size = 15641099 bytes (16712158 bytes = 51.66% decrease)
The resulting image is corrupted and will not be outputted.
This is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues
The resulting image is corrupted

I can't attach the image here, so here's a Google Drive link to it: https://drive.google.com/open?id=1zcV6uZ3mRtlf9mzDtIamPM3glm59z4tP

Edit: I'm using oxipng version 2.2.0 on Debian Linux:
andy@sandy:~$ uname -a
Linux sandy 3.16.0-7-amd64 #1 SMP Debian 3.16.59-1 (2018-10-03) x86_64 GNU/Linux

shssoichiro:
Thanks for reporting this! It might be related to #168. I'll leave both open until I can verify both are fixed.

PR #457, opened 2022-10-24
Full switch to Libdeflater

### LINKED ISSUE #422: Support forwarding `freestanding` feature to `libdeflater` for WebAssembly
I'm not sure if this is already possible, but I couldn't find a way to enable the [`freestanding` feature of libdeflater](https://github.com/adamkewley/libdeflater#compile-time-features) apart from adding it as a dependency next to oxipng and setting that feature (which is relying on Cargo deduplicating the two instances of `libdeflater` which isn't guaranteed?).

```
oxipng = {version = "5.0.0", features = ["libdeflater", "zopfli"]}
libdeflater = {version = "0.7", features = ["freestanding"]}
```

So maybe oxipng should expose a `freestanding` feature that  is then forwarded to its `libdeflater` dependency?

This is needed to build for the wasm32-unknown-unknown target.

### LINKED ISSUE #454: Allow setting libdeflate compression level

andrews05:
Libdeflate has improved (since first added in #203) to the point where it almost always compresses significantly better than even cloudflare-zlib, and while slower at its highest levels it generally ends up faster in oxipng due to not needing to test 4 different "strategies". It definitely seems worth fully switching to libdeflate instead of zlib, as benchmarks below will show.

This PR is fairly large and includes the following changes:
- Changes all compression to libdeflater, adjusting default options as appropriate (see below).
- Changes all decompression to libdeflater. This is significantly faster but has little impact on total speed.
- Changes crc32 checksums to libdeflater's implementation. As above, this is also significantly faster but still has little impact on total speed.
- Removes dependencies on cloudflare-zlib, miniz-oxide, and crc.
- Removes redundant `-D`, `--zs` and `--zw` options.
- ~~Changes evaluation to only check filter 0 instead of both 0 and 5. This is faster and made zero difference to output in the files I tested.~~
- Adds `freestanding` feature to expose `libdeflater/freestanding`. (Closes #422).

The new optimisation levels are as follows, including a reinstated level 4:
```
-o 0   =>  --zc 5 --nz          (0 or 1 trials)
-o 1   =>  --zc 10              (1 trial, determined heuristically)
-o 2   =>  --zc 11 -f 0,5       (2 trials)
-o 3   =>  --zc 11 -f 0-5       (6 trials)
-o 4   =>  --zc 12 -f 0-5       (6 trials; same as `-o 3` for zopfli)
-o 5   =>  --zc 9-12 -f 0-5     (24 trials; same as `-o 3` for zopfli)
-o 6   =>  --zc 1-12 -f 0-5     (72 trials; same as `-o 3` for zopfli)
```
I benchmarked against the files in the test folder, totalling 23511320 bytes (excluding the >30MB `issue-167.png` which takes forever and skews the results).

Level | Zlib size | Libdeflate size | Zlib time (s) | Libdeflate time (s) | Zlib -t1 | Libdeflate -t1
-- | -- | -- | -- | -- | -- | --
-o0 | 21166902 | 20867510 | 10.32 | 8.86 | 14.67 | 8.98
-o1 | 19989869 | 19033143 | 29.88 | 27.04 | 35.44 | 27.88
-o2 | 19606473 | 18604674 | 31.12 | 41.37 | 66.92 | 63.42
-o3 | 19369736 | 18310670 | 52.82 | 51.78 | 202.67 | 190.62
-o4 |   | 18168526 |   | 84.42 |   | 335.27
-o5 | 19369729 | 18168521 | 93.82 | 152.48 | 415.52 | 646.27
-o6 | 19369729 | 18168521 | 99.07 | 167.83 |   |  
-o3 -Z | 18042595 | | >2.5 hours | | |

Caveats:
- As libdeflate is not a streaming library, we sometimes have to guess output buffer sizes. For IDAT the decompressed size is known in advance, but for iCCP I've just made a rough estimate of the max size based on some samples. It should be generous enough but the worst that can happen is it will fail to decompress and thus won't be able to replace it with a rendering intent.
- Again due to the non-streaming nature, there is no opportunity to abort early if a timeout has been specified. Not sure how much of a problem this is.


andrews05:
@shssoichiro Running `cargo fmt` will reformat all the args in main.rs (see the output of the failed test). Are you happy for this to happen?

shssoichiro:
That's interesting. I wonder if that's the result of a Rust version bump. But yes, that's fine, please let that happen.

andrews05:
I reverted the change to eval filters since after #458 landed it turns out filter 5 does have an impact. There are improvements to be found here which I'm continuing to explore, but they don't belong in this PR.

P.S. Do you mind if I remove/replace issue-167.png? It's super annoying when running tests 😅

shssoichiro:
What's the problem with issue-167.png exactly?

shssoichiro:
Ah I see. I checked on what #167 is and unfortunately the issue is related to the file being large. So I don't think there's a way around it. :sweat_smile:

PR #458, opened 2022-10-28
Fix directional alpha reductions

andrews05:
The alpha up and down modes were not working correctly, producing the same output as alpha black. This was also causing a nasty performance issue with alpha down.
I've fixed both of these modes and tested on one of the included files:
`oxipng -o0 -a issue-133.png`
| branch | IDAT size | Time |
|-|-|-|
| master | 1076430 | 56.1s |
| alphafixes | 1030377 | 3.6s |

The left and right alphas were also not giving the expected result, only extending pixels once due to incorrect tracking of the `last_pixel`.

It would probably be good to add tests for these, though I'm not sure how you would do that.

Thoughts for further improvements: A typical method for alpha optimisation is to "smear" in all directions at once, which can give a more compressible result. An example of this can be seen if you remove the alpha from the bear picture in the rgba test files:
rgba_8_reduce_alpha_black

### DIFF
--- a/src/reduction/alpha.rs
+++ b/src/reduction/alpha.rs
@@ -108,7 +108,7 @@ fn reduced_alpha_to_up(png: &PngImage, bpc: usize, bpp: usize) -> Vec<u8> {
                 current_line.extend_from_slice(pixel);
             }
         }
-        last_line = current_line.clone();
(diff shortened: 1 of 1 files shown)


shssoichiro (approved):
That is quite an impressive speed difference... Thanks for the fix!

PR #461, opened 2022-11-06
New filter strategies

### LINKED ISSUE #183: Additional heuristic filter strategies
There are a number of different filter heuristics that perform better than the one used in libpng documented [here](http://bjoern.hoehrmann.de/pngwolf/).

I have implemented two additional strategies on the [better-filters](https://github.com/shssoichiro/oxipng/compare/master...fhanau:better-filters) branch. There are a number of caveats with these that need to be discussed, so I am opening this as an issue instead of a pull request.

- The bigrams strategy is more expensive than minsum (the current heuristic), but on average delivers better results than minsum. Trying both strategies yields the smallest size.
- The distinct bytes strategy rarely delivers the best result in my tests and on average performs worse than minsum, so it's questionable whether it is worth including.
- Enabling these strategies by default inevitably increases runtime, so I am not sure what the best way to integrate them is. As a possible solution, I have written some code that will add the option to use heuristics to choose the color type, zlib strategy and filter strategy at a smaller compression level to decrease the time spent deflating. I plan to open a seperate pull request for those changes.
- For more aggressive optimizations, there are some additional filter strategies based on the size of the deflated scanline or the shannon entropy as implemented in [zopflipng](https://github.com/lvandeve/lodepng/blob/f6155a4206046a31a9532d70caf517845af64c83/lodepng.cpp#L5002)
- I also added an experimental bias factor to the bigrams strategy, which slightly improves compression on average, but needs more work.

andrews05:
This PR adds 4 new heuristic filter strategies. These are all generally better than the standard number 5 strategy (MinSum) and can save even more when combined together.
- 6: Shannon entropy algorithm, as seen in [LodePNG](https://github.com/lvandeve/lodepng)/Zopfli.
- 7: Count of distinct bigrams, as seen in [pngwolf](https://bjoern.hoehrmann.de/pngwolf/).
- 8: Shannon entropy of bigrams, inspired by the two above.
- 9: Brute force by compressing each filter on each line. This is currently configured to compress 4 lines at a time (the current filter attempt plus 3 previous lines to provide context). It's very effective and not as slow as one might expect, thanks to libdeflate's high-speed level 1.

Benchmarks on a set of test files with `-o4 -a -i0 -s`:
| Filters | Total size | Time | Filters | Total size | Time |
|-|-|-|-|-|-|
| 0,5 | 27572705 | 1:24 | 0,1,2,3,4,5 | 27411540 | 1:49 |
| 0,6 | 27404150 | 1:24 | 0,1,2,3,4,6 | 27321509 | 1:50 |
| 0,7 | 27376244 | 1:23 | 0,1,2,3,4,7 | 27292046 | 1:51 |
| 0,8 | 27406071 | 1:30 | 0,1,2,3,4,8 | 27305417 | 1:57 |
| 0,9 | 27348063 | 1:38 | 0,1,2,3,4,9 | 27170547 | 2:06 |
| | | | 0-9 | 27147885 | 3:29* |

*My machine has 8 threads, so using all 10 filters pushed me over my thread count.

Additional changes made in this PR:
- Filter types now use an enum, making it much easier to keep track of them all.
- The filter byte is included in the output from `filter_line`, which means this is included in all heuristic evaluations. This is how LodePNG does it with the entropy strategy, and I think it makes sense.
- [bit-vec](https://crates.io/crates/bit-vec) has been replaced with [bitvec](https://crates.io/crates/bitvec), which includes a fast bitset necessary for the bigrams strategy. As a bonus, reductions are now about 10% faster.
- [rustc-hash](https://crates.io/crates/rustc-hash) has been added, so the bigram entropy strategy can use the fast FxHasher (this strategy is still relatively slow though, I'm open to ideas about faster approaches that might be used). Color to palette reductions now also use this hasher for a 30% speed improvement.
- Additional tests added for changing interlacing, which helped me ensure the interlacing was correct when I replaced the bitvec.
- List of all filters shown in help.

Note that these filters are currently all opt-in. I will investigate including these in the default optimisation levels in future.

Closes #183

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -32,12 +32,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
-name = "bit-vec"
-version = "0.6.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
-
-[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -380,7 +374,6 @@ checksum = "9ff7415e9ae3fff1225851df9e0d9e4e5479f947619774677a63572e55e80eff"
 name = "oxipng"
 version = "6.0.1"
 dependencies = [
- "bit-vec",
  "bitvec",
  "clap",
  "crossbeam-channel",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,6 +44,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitvec"
+version = "1.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1bc2832c24239b0141d5674bb9174f9d68a8b5b3f2753311927c172ca46f7e9c"
+dependencies = [
+ "funty",
+ "radium",
+ "tap",
+ "wyz",
+]
+
+[[package]]
 name = "bytemuck"
 version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -193,6 +205,12 @@ dependencies = [
  "crc32fast",
  "miniz_oxide",
 ]
+
+[[package]]
+name = "funty"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
 
 [[package]]
 name = "glob"
@@ -363,6 +381,7 @@ name = "oxipng"
 version = "6.0.1"
 dependencies = [
  "bit-vec",
+ "bitvec",
  "clap",
  "crossbeam-channel",
  "filetime",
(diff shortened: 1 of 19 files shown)


andrews05:
@kornelski I'm not sure if the bitvec `opt-level = 3` setting is still useful with the new crate? I didn't experience any performance concerns in debug either way.

kornelski (approved):

AlexTMjugador:
It's good to see all of these PRs coming from you lately. Thank you for your great work! :heart:

> [rustc-hash](https://crates.io/crates/rustc-hash) has been added, so the bigram entropy strategy can use the fast FxHasher (this strategy is still relatively slow though, I'm open to ideas about faster approaches that might be used). Color to palette reductions now also use this hasher for a 30% speed improvement.

My two cents here: this may be a bit stupid and you probably have thought about it already, but have you considered using [`aHash`](https://github.com/tkaitchuck/aHash) instead? According to its benchmarks, its faster than the vanilla FxHash, and it might be faster than the custom hasher that `rustc-hash` uses.

andrews05:
@AlexTMjugador Thanks for your comment!

I tried a number of hashers, which generally weren't as good as I was expecting them to be.
- The default (SipHash): cryptographic hashers are obviously not the fastest
- nohash: as the keys are just u16s I was expecting this to be a great option but it was only a small improvement
- twox-hash: this was okay but not great - the xxh3 implementation was actually horribly slow
- wyhash: this was decent, though also not as good as I was expecting based on the claims it makes
- ahash: don't recall specific results but wasn't as good as fxhash
- metrohash: as above

Other ideas that weren't as fast:
- Plain vec (65k long), like the base Entropy strategy
- IndexMap (with FxHasher)

Thought I just had:
- Sparse vec?

Note oxipng is my first experience with rust, so any feedback is welcome :)

kornelski:
@shssoichiro I think it's good to merge

shssoichiro:
Thanks, this looks amazing! I apologize for the slow response in getting it merged. That's completely on me.

PR #463, opened 2022-11-19
Refactor use_heuristics for fast filter evaluation

andrews05:
This PR is an intermediary step between the introduction of new filters in #461 and a future rebalancing of optimisation levels.

This is all about improving speed, including new fast filter evaluation (similar to #185).
1. The ability to set multiple compression levels has been removed. In my experience, this was highly ineffective (like saving 5 bytes in 20MB) and removing it allowed some helpful simplifications which are relevant here. E.g. filtering is now done within the trial loop rather than beforehand, which possibly helps with performance a little. The change does mean that levels 5 and 6 are currently equivalent to 4, but the intention is to adjust these later to use the new filter strategies (which *are* effective). This isn't strictly necessary though, so could be reverted if you'd prefer to keep this ability.
2. The baseline evaluator trial now passes right through the evaluator and may be selected as the best result (but will be flagged as non-reduction). This has two immediate, though minor, effects: Firstly, if there are no other compression trials and the compressed baseline happened to be smaller than the original, it can be output directly rather than always reverting to the original. Secondly, it fixes an issue of non-determinism where if a reduction evaluation happened to complete *before* the baseline, but the baseline evaluation was actually smaller, the reduction would still be chosen for the output.
3. The baseline will only be set if any reductions are being attempted. This means that if there are no reductions, no work is actually done during the evaluation step, resulting in a helpful speed boost. Additionally, this fixes an issue where the baseline was never set if the interlacing flag was used, even if the interlacing didn't actually change.
4. The evaluator compression level and filters are now passed into the constructor, rather than being fixed constants.
5. The `use_heuristics` option is now `fast_evaluate` and switches into a different path for compression trials. This will take the output of the reduction evaluation, combine it with a similar evaluation of any additional filters (ones that weren't tested in reduction), and then perform a single main compression trial with the best filter. This is much faster and is more effective when only one filter was originally selected because it still actually has two filter results to compare if reductions were attempted. It is slightly less effective for multiple filters, though this doesn't currently affect any of the default levels. Additionally, if the specified compression level matches the one used for evaluations (as it does with o0), it can just output the evaluation result directly.

Here are some results from a set of sample images (original size 42955740):

| options | master size | master time | PR size | PR time |
|-|-|-|-|-|
| -o0 | 41229590 | 9.9s | 40993901 | 6.5s |
| -o1 | 32062204 | 36.3s | 31885429 | 33.1s |
| -o2 | 30405937 | 54.3s | 30405937 | 51.7s |
| -o1 -f0-5 | 30288484 | 47.0s | 30313128 | 36.8s |

- o0 is slightly more effective because filters 0 and 5 are both always tested in reductions. It is much faster because the baseline isn't being tested when there are no reductions, and because it doesn't actually do any compression trials.
- o1 is much the same, though doesn't get the same speed boost as it still has to do a compression trial.
- o2 is unchanged in size because it doesn't set the `fast_evaluate` option, but still receives some of the speed benefits.
- o1 with additional filters is slightly less effective because the filters are using the fast evaluation, but is of course much faster.

Things that didn't make the cut:
- I wanted to keep the filtered idat data from the evaluator so it didn't have to be refiltered for the compression trial but it would only be a small improvement and I decided it wasn't worth the additional refactoring that would be necessary. Maybe another time.
- I spent a *lot* of time testing faster evaluator settings, similar to #188, but ultimately decided to leave them as is.

Apologies for the long PR, I hope this is all helpful :)

Closes #185.
(Should probably close #188 as well.)

### DIFF
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -1,5 +1,3 @@
-use indexmap::IndexSet;
-
 mod deflater;
 pub use deflater::crc32;
 pub use deflater::deflate;
@@ -17,8 +15,8 @@ pub use zopfli_oxipng::deflate as zopfli_deflate;
 pub enum Deflaters {
     /// Use libdeflater.
     Libdeflater {
-        /// Which compression levels to try on the file (1-12)
-        compression: IndexSet<u8>,
+        /// Which compression level to use on the file (1-12)
+        compression: u8,
     },
     #[cfg(feature = "zopfli")]
     /// Use the better but slower Zopfli implementation
@@ -29,3 +27,5 @@ pub enum Deflaters {
         iterations: NonZeroU8,
     },
 }
+
+impl Copy for Deflaters {}
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -10,7 +10,7 @@ mod zopfli_oxipng;
 #[cfg(feature = "zopfli")]
 pub use zopfli_oxipng::deflate as zopfli_deflate;
 
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 /// DEFLATE algorithms supported by oxipng
 pub enum Deflaters {
     /// Use libdeflater.
@@ -27,5 +27,3 @@ pub enum Deflaters {
         iterations: NonZeroU8,
     },
 }
-
-impl Copy for Deflaters {}
--- a/src/atomicmin.rs
+++ b/src/atomicmin.rs
@@ -27,7 +27,8 @@ impl AtomicMin {
         &self.val
     }
 
-    pub fn set_min(&self, new_val: usize) {
-        self.val.fetch_min(new_val, SeqCst);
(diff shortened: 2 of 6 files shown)


andrews05:
One other thought I had: It might be good to expose the use_heuristics option in the cli, and maybe rename it to something that better reflects what it does here. `--fast-filter-evaluation` seems a bit much though 😅. Thoughts?

src/deflate/mod.rs
```diff
@@ -29,3 +27,5 @@ pub enum Deflaters {
         iterations: NonZeroU8,
     },
 }
+
+impl Copy for Deflaters {}
```
shssoichiro, line 31: This should be cleaner if we add it to the derives list

shssoichiro:
Thanks, other than the one nit I think this looks good! I don't mind the addition of a CLI option for it as well.

src/deflate/mod.rs
andrews05: Thanks! Not sure why I thought I needed an impl...

PR #464, opened 2022-11-20
Fast deinterlace for 8-bit+

andrews05:
This is an easy one: A seperate deinterlace by bytes function, 10x faster than the bits functions.

Bench: deinterlacing_8_bits
Before: 62,269,705 ns/iter
After: 5,763,391 ns/iter


shssoichiro (approved):

shssoichiro:
Impressive, thanks!

Issue #459, opened 2022-10-29
Oxipng produces corrupted paletted images


DangerPie:
Pngcheck says "bKGD index (61) falls outside PLTE (61)"
[wrong_bKGD_index.zip](https://github.com/shssoichiro/oxipng/files/9893334/wrong_bKGD_index.zip)

andrews05:
Just FYI, I've been looking into this as it came up during some experiments I was doing. The bKGD colour can get dropped from the palette if it isn't used in the idat. I hope to get a PR up to fix this sometime, but it may be a little more involved than it seems...

DangerPie:
IrfanView refuses to show that image.

PR #466, opened 2022-11-29
Palette reduction fixes

andrews05:
This is primarily a fix for #459, which I ran into myself during some experiments. The issue is that the bKGD colour for indexed images may not be present in the idat and so the colour may get dropped from the palette, leaving an invalid index. This could happen not only due to palette reductions but also during output of the headers where the palette was being truncated to match the bit depth.

The following changes have been made:
- Reduction tests now verify that the bKGD index exists in the palette. This reveals failures in about 4 tests, which are fixed by later changes.
- Palette truncation on output will now consider the bKGD.
- Palette reduction will also ensure the bKGD colour is retained, even if not used in the idat.
- The palette_map_to_byte_map function will check the entire map when considering whether any change is necessary. This is important if there was an invalid index used in the idat and the palette needs to be extended to fix it.
- Stripping is now performed before reductions, to make sure the bKGD colour *won't* be retained in the palette if it is being stripped.
- The "safe" chunk list has been updated. This is not strictly relevant to the fixes here but as the bKGD is a bit annoying to deal with I wanted to make sure it's considered "safe" to strip. Happy to revert any of these if you disagree though:
  - Removed bKGD: This is explicitly ignored by web browsers and has little to no relevance in other modern software.
  - Removed hIST and sPLT: These are historic chunks that don't actually affect rendering on modern devices/software.
  - Removed gAMA and cHRM: These are generally outdated and also explicitly ignored by web browsers. [This article](https://hsivonen.fi/png-gamma/) explains why the gAMA is usually incorrect anyway, and could cause harm in older browsers.
  - Added cICP: This is very new chunk which overrides iCCP when recognised. [More info](https://w3c.github.io/PNG-spec/#cICP-chunk).


shssoichiro (approved):
Thanks!

PR #467, opened 2022-12-02
Rebalance presets

### LINKED ISSUE #106: An option to use zopfli but only once (for the final pass with probably-best filters/palette/etc.)
As we know Zopfli is very slow, so it's quite wasteful to run it several times.  
What I propose is a "fast zopfli" option which will make it work like this:
1) test various filters/palette/etc. combinations with a faster compression scheme (zlib-something)
2) pick the best filters/palette/etc. combination and use zopfli compression with them for the final result.  

I tested this approach on a collection of images (without modifying any code, just applying zopfli compression with `advdef` utility as a separate step), and it seems to give better compression/time tradeoff.

andrews05:
The final PR related to the new filters, this adds the `--fast` cli flag corresponding to the `fast_evaluation` option, and rebalances all the presets to take advantage of the new filters. The Bigrams filter is now the primary heuristic filter, replacing MinSum in reduction evaluations, as well as levels 0 and 1.
Additionally this changes the PngData struct to retain the filtered, uncompressed data in a separate property. This is important for level 2 with the `--fast` flag so that it doesn't have to refilter the data before the final compression.

The new optimisation levels are as follows:
```
-o 0   =>  --zc 5 --fast                (1 trial, determined heuristically)
-o 1   =>  --zc 10 --fast               (1 trial, determined heuristically)
-o 2   =>  --zc 11 -f 0,1,6,7 --fast    (1 trial, determined by fast evaluation)
-o 3   =>  --zc 11 -f 0,7,8,9           (4 trials)
-o 4   =>  --zc 12 -f 0,7,8,9           (4 trials; same as `-o 3` for zopfli)
-o 5   =>  --zc 12 -f 0,1,2,5,6,7,8,9   (8 trials)
-o 6   =>  --zc 12 -f 0-9               (10 trials)
```
Benchmarks with 241 images, totalling 42955740 bytes.
(master here is at [15a9b4a](https://github.com/shssoichiro/oxipng/tree/15a9b4a3e79f50a758f3f0a4f7e9aaf5ae649a21), prior to #463)
Level | master size | master time | PR size | PR time | PR time (-t1)
-- | -- | -- | -- | -- | --
-o0 | 41229590 | 10.32 | 33203093 | 11.78 | 14.47
-o1 | 32062164 | 36.10 | 30687260 | 34.24 | 38.29
-o2 | 30405937 | 54.50 | 29895213 | 52.85 | 1:05.04
-o3 | 30047538 | 1:11.41 | 29768631 | 1:11.60 | 3:22.97
-o4 | 29921386  | 1:49.01 | 29649297 | 1:42.22 | 5:10.51
-o5 | 29920135 | 3:22.76 | 29627181 | 2:21.63 | 9:45.03
-o6 | 29920135 | 3:41.16 | 29625437 | 3:11.03 | 12:01.32

Note level 0 is slightly slower but far more effective, due to no longer using the `--nz` flag.

One other minor detail to note: The heuristic selection used in levels 0 and 1 will now choose Bigrams for 8-bit indexed images. None is only used for bit depths < 8, where delta filters are often ineffective.

[edit]
This should close #106, as the `--fast` option should work well with zopfli.


shssoichiro (approved):
Thanks! Huge -o0 improvement!

PR #470, opened 2022-12-07
Fix Grayscale depth reduction with tRNS

### LINKED ISSUE #426: PNG files are often corrupted when inputs are monochrome.
```
Processing: Mona_Lisa-PNG8-Scatter-BW.png
    404x600 pixels, PNG format
    1x8 bits/pixel, Grayscale
    IDAT size = 18171 bytes
    File size = 18242 bytes
Reducing image to 1x1 bits/pixel, Grayscale
Trying: 6 combinations
Found better combination:
    zc = 0  zs = 0  f = 0        16966 bytes
    IDAT size = 16966 bytes (1205 bytes decrease)
    file size = 17037 bytes (1205 bytes = 6.61% decrease)
The resulting image is corrupted and will not be outputted.
This is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues
The resulting image is corrupted
```

I don't know what's happening here, but I ran the command line `oxipng -s -o 6 -Z --fix *BW.png` in an attempt to figure out why earlier runs (without `-Z`, as well as with and without `-D`) had some corrupt images, but only for monochrome black and white PNGs (1-bit). All of the attached PNGs had corrupt output when given the above command line. I an on Windows 10 64-bit, and used a just-downloaded 5.0.1 release. Zopfli compression claimed it had significant size reductions, but then also that the output was corrupted and so was not saved. I don't have a Rust dev environment set up, and I suspect compiling oxipng from source would be a challenge for me, so I guess I'm just reporting what it says to report, and providing some of the images that failed.

Mona_Lisa-Gif-BlueNoise-BW
Mona_Lisa-Gif-ChaoticNoise-BW
Mona_Lisa-Gif-Diffusion-BW
Mona_Lisa-Gif-Gradient-BW
Mona_Lisa-Gif-Neue-BW
Mona_Lisa-Gif-None-BW
Mona_Lisa-Gif-Pattern-BW
Mona_Lisa-Gif-Scatter-BW

andrews05:
This is a fix for #426. The tRNS chunk was not being accounted for when reducing the bit depth of a grayscale image.
There are two possibilities we need to account for here:
1. If the tRNS value can be losslessly reduced to the new bit depth, we must do so.
2. If it can't be, this means the value does not appear in the IDAT and the tRNS must be removed.

I've added two tests for both of these cases.

The example images in #426 are in grayscale-8 with a tRNS value of 1 (near-black). This colour does not appear in the image and when the depth is reduced to 1-bit, the tRNS value of 1 now refers to white, making the image black and transparent instead of black and white.

Closes #426.

[edit]
Would be good if you could double-check my logic for validating the tRNS value. Or see if you have any ideas for achieving the same thing without a loop. I note the current grayscale reduction only supports reducing from 8-bit (not from 4 or 2), but the tRNS reduction/validation should continue to work even if that limitation is removed in future (it would be a bit of work though, so not inclined to do it right now).


andrews05:
Just realised tRNS isn’t accounted for in 16 to 8 reductions either. It’s probably highly unlikely to happen but I should fix that too…

andrews05:
So it seems the high bytes of the tRNS chunk are ignored for 8-bit, which means there's no need to reduce it during the 16 to 8 reduction.

shssoichiro (approved):

PR #474, opened 2022-12-10
Fix fast with zopfli

andrews05:
So I made a little mistake when I said the `--fast` flag would work well with zopfli: I failed to actually test this 😬
Here's a quick fix to make it actually work.


shssoichiro (approved):
Looks good, if I can get Github Actions to cooperate. Thanks.

PR #468, opened 2022-12-04
More Reductions

andrews05:
This PR adds a new reduction from Grayscale+Alpha to Indexed, as discussed in #195.
Additionally, if alpha reductions are enabled, the black reduction is now performed prior to colour type reductions, which can allow reductions from alpha to indexed that may not have been possible otherwise. (This resulted in the bigger win in my tests).

Comparison on a sample set totalling 42,955,740 bytes:
Master: 27,751,163
PR: 27,457,738

Fixes #195.

PR #475, opened 2022-12-10
🚀 Revamp alpha optimisation

andrews05:
This PR introduces an entirely different approach to alpha optimisation, which is more effective than before and much faster. I think this is quite cool, so I'll start by showing off some benchmarks.

Set of non-reducible RGBA files totalling 15,737,898 bytes, compressed using `-o3`:
| Test | Size | Time |
|-|-|-|
| Without `-a` | 9,868,361 | 20.463 |
| With `-a`, master | 8,813,177 | 24.155 |
| With `-a`, PR | 8,754,222 | 20.488 |

Following comments on #468, I continued searching for better alpha techniques and eventually came across [CryoPNG](https://encode.su/threads/1260-CryoPNG-short-introduction). The key here is to alter the data such that it becomes zero when filtered. While CryoPNG does this assuming a single delta filter for the whole image, we can take it to the next level by integrating the technique into the actual filtering stage. This allows it to work for heuristic filters and actually help guide the heuristic selection.

The speed advantage is that there's no need to do any evaluator trials. Being tailored to the specific filter being applied, the output is near-optimal - this doesn't mean it's the best possible but does mean it would be difficult to improve upon. As such, the original image data is no longer considered - there are still images where the original data could be better but these would be quite rare and it would be impractical to evaluate in the new setup.

An example of what the technique looks like:
elephant elephant-reveal

I've removed all the redundant tests and associated files (and also some redundant deflate benches leftover from before the switch to libdeflate).


andrews05:
Additional improvement for indexed images: Palettes with multiple fully transparent entries will be appropriately reduced when alpha optimisation is enabled. This would generally be uncommon but can happen if e.g. an RGBA image with dirty alpha is reduced to palette *without* alpha optimisation and then is later re-run *with* alpha optimisation.

ace-dent:
Looks really interesting.
Cautionary comment: you may be effectively 'breaking' the images if then run through other compression engines. Some Users chain compression engines (and combine them! e.g. '[Huffmix](https://encode.su/threads/1313-Huffmix-a-PNGOUT-r-catalyst)')- so block split thresholds and freedom to choose other filters will break greatly affected based on these *hard* changes made to the image.

I'd like to be able to specify a dirty transparency technique applied - like 'Up' in cryopng- to then pass through different compression engines, etc. With Oxipng, can I force the filter (`0`–`5`) and generate 6 different images to evaluate?

andrews05:
@ace-dent Thanks for your comment! Yes, you can force the filter with `-f`. Note that oxipng tries multiple filter strategies according to the preset you use, with all 10 (5 delta + 5 heuristic) being evaluated at the highest preset (`-o6`).

I guess you could try all 10 independently and run each of those through your chain, though I would be surprised if you end up finding a filter strategy that worked better than the one oxipng would have chosen itself (for RGBA, it’s usually the Brute strategy). Filters and good alpha optimisation have to work hand in hand, so any other tool that you use that tries to select better filters without some attempt to optimise alpha for that filter is unlikely to succeed. Does this make sense? Feel free to correct me if I’m misunderstanding here. 

(Also note that any type of alpha optimisation is necessarily a “hard” change to the image, so this PR is no different from oxipng’s existing alpha techniques in that sense.)

andrews05:
As an example, the mouse png demonstrated by CryoPNG came out at 105,035 using a mishmash of tools. Oxipng previously could only achieve around 109k, but with this PR it can single-handedly reach 104,308 (or 103,921 in zopfli mode).

As a *counter* example, the coloured spots image in #460 has very nice gradient background when you remove the alpha - highly structured data that works well with delta filters. In current oxipng this would optimized 20,391 bytes - the evaluator would choose to keep the existing background. But in this PR it instead comes out at 24,348 because it makes no attempt to evaluate the existing background. This type of image is very much the exception, but the difference is significant enough (nearly 20%) that it may be worth attempting to resolve...

andrews05:
@shssoichiro Thanks for merging! I continued investigating the regression with the particular image I mentioned above. The good news is the problem turned out to be greatly reduced at `-o6` when the paeth filter kicked in.
I did toy around with disabling alpha optimisations for the paeth filter specifically, which completely mitigated the issue for that image with negligible impact on any others, but it seemed very hacky and over-specialised. So I'm inclined to leave it as is, but let me know if you have any thoughts.

shssoichiro:
If you think there's a strong case for disabling alpha opts for paeth because there's negligible improvements to be had, I'd be fine with adding that logic, as long as we're confident it can be generalized across a majority of images with paeth filtering. What I don't want to do is over-specialize to one or two edge case images at the cost of the majority of images.

ace-dent:
Thanks @andrews05 and @shssoichiro for your replies. 
Over the years (working on pngslim) I've found many examples that compressed better with the original 'dirty alpha' left intact. I'd love to do some testing with Oxipng... might you add w32 builds to future releases?

andrews05:
@shssoichiro Well I was hoping it would give improvements on other images where the original alpha was slightly better too, but it turned out to only work for that one image in my set. And the downside of disabling alpha on paeth would be that if you had a version of that same image where the alpha had been cleaned/blackened, the paeth filter *with* alpha optimisation would actually do quite a good job with it. So the result is it specialises one or two edge cases at the cost of one or two other edge cases, which is probably not the right thing to do.

@ace-dent Oh, you're the pngslim author? Not one I've tried myself (not a Windows user) but great to see your work there.
You're absolutely right that some images are better with the original alpha left intact. This is far less so with the changes here, but still true all the same. There can indeed be losses with oxipng now that it no longer evaluates this, but these are both small and rare and I think the improvements in the average case far outweigh this. If you want to get the best result and speed is no object, you could always try running it both with and without alpha optimisations (`-a`) and see which one is better.

shssoichiro:
@ace-dent There used to be w32 builds as part of the Github releases, but it ended up being removed a while back due to being difficult to maintain for something that sees relatively little use. For the foreseeable future, if you're on 32-bit Windows you'll need to compile oxipng yourself.

PR #481, opened 2022-12-19
Change validation to a debug assert

### LINKED ISSUE #473: Resulting image is corrupted with oxipng 6.0.1 and 7.0.0 but not 5.0.1
Curiously, .7z compresses this very good.
[Oxiorig.zip](https://github.com/shssoichiro/oxipng/files/10198201/Oxiorig.zip)

### LINKED ISSUE #478: Thinking about output validation
I know I said I may not do any more work here, but I was thinking about the raw API in #376 which could be quite helpful for my use case. That led me to think about the output validation and how that would work for a raw API and I decided it might be worth having a separate discussion about it.

The issue with a raw API is that it may not be practical/easy to validate that the output pixels match the input pixels, since the input is raw data rather than a png image. This would also be an issue with hypothetical lossy options that may get added one day (e.g. forcing 16-bit to 8-bit). The validation also takes time and adds a bit to the binary size, so I think at the least it may be a good to idea to make it an optional feature.

That said, there may also be a case for simply removing it. I'm sure it has been a great help over oxipng's development but oxipng is quite stable and mature now, with an extensive test suite. Since 2019, there have been four different issues of "corrupt output" reported. One of these (#426) was a bug in oxipng resulting in mismatched pixels. Another one (#449) was a bug in zopfli resulting in an invalid file. The other two (#167 and #473) were bugs in the library used to perform the validation and not actually corrupt output, the latter of which is still waiting to be fixed. So it seems it may be causing just as many problems as it is solving.

If it were to be removed, we could instead use oxipng's own parsing to check that the output is a valid png. This should catch critical issues like the zopfli bug but not mismatched pixels. We would also want to continue using the existing validation for the tests, but keep it separate from oxipng itself.

What do you (or others) think?

andrews05:
As discussed in #478, assuming you're happy to proceed with this :)

As far as I can tell, there's no way to control dependencies or features based on the build profile, which means we have to use debug asserts (either that or restrict it to tests only). It still keeps the release binary smaller though.

P.S. Can we remove issue-167 now? 🤣 (this is the really big one that was only failing because of a bug in the image crate)

Closes #478
Fixes #473

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -657,28 +657,9 @@ fn optimize_png(
         );
     }
 
-    let (old_png, new_png) = rayon::join(
-        || load_png_image_from_memory(original_data),
-        || load_png_image_from_memory(&output),
-    );
+    debug_assert!(validate_output(&output, original_data));
 
-    if let Ok(new_png) = new_png {
-        if let Ok(old_png) = old_png {
-            if images_equal(&old_png, &new_png) {
-                return Ok(output);
-            }
-        } else {
-            // The original image might be invalid if, for example, there is a CRC error,
-            // and we set fix_errors to true. In that case, all we can do is check that the
-            // new image is decodable.
-            return Ok(output);
-        }
-    }
-
-    error!(
-        "The resulting image is corrupted and will not be outputted.\nThis is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues"
-    );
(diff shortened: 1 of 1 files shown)


shssoichiro:
Oh, removing the test for issue-167 because it's slow? Yes, I'd be fine with that. 😅

shssoichiro (approved):

Issue #432, opened 2022-01-17
Attempt to recompress iCCP, zTXt, and iTXt


gsnedders:
These are all also deflated chunks, and as such it's probably worthwhile trying to recompress these with whatever compressor is in use.

TPS:
Perhaps also judicious conversion of `tEXt`→`iTXt`→`zTXt` _before_ recompression to minimize input text representation?

Of course, that's assuming any of these chunks need to stay @ all.

gsnedders:
The motivating thing here was iCCP, given that is preserved by `--strip safe`, with its implications on rendering. But by the point I filed the issue, it seemed like I may as well include all other compressed chunks given once you've done one, you may as well do them all.

andrews05:
iCCP recompression has been added now. The best way to minimise text chunks is with `--strip` 😅

PR #482, opened 2022-12-23
Raw API

### LINKED ISSUE #376: An API to compress raw image data?
Right now, in Squoosh, OxiPNG is the only codec where we need to encode the raw image data we already have (RGBA) into PNG with browser API, only to pass to OxiPNG, which will decompress it back.

I wonder if there could be a way to expose an API that would accept raw RGBA byte array and try to compress that directly. This could make it both faster and smaller by excluding the decoder from the process.

andrews05:
This is a proposal for a raw API, as discussed in #376.

```rust
pub struct RawImage {
    /// Construct a new raw image definition
    ///
    /// * `width` - The width of the image in pixels
    /// * `height` - The height of the image in pixels
    /// * `color_type` - The color type of the image
    /// * `bit_depth` - The bit depth of the image
    /// * `data` - The raw pixel data of the image
    pub fn new(
        width: u32,
        height: u32,
        color_type: ColorType,
        bit_depth: BitDepth,
        data: Vec<u8>,
    ) -> Result<Self, PngError>

    /// Add a png chunk, such as "iTXt", to be included in the output
    pub fn add_png_chunk(&mut self, chunk_type: [u8; 4], data: Vec<u8>)

    /// Add an ICC profile for the image
    pub fn add_icc_profile(&mut self, data: &[u8])

    /// Create an optimized png from the raw image data using the options provided
    pub fn create_optimized_png(&self, opts: &Options) -> PngResult<Vec<u8>>
}
```
(Closes #376)


andrews05:
@Zerowalker I was hoping to get feedback/input from people wanting to use it to see whether the implementation is actually suitable. Do you have any thoughts?

Zerowalker:
> @Zerowalker I was hoping to get feedback/input from people wanting to use it to see whether the implementation is actually suitable. Do you have any thoughts?

Ah.

hmm, is the raw data expected to be an empty `PngImage` with the `data` part set?
I would generally expect a raw image function to ask for the raw data, width, height and bit depth.
So i find that part a bit odd, but perhaps i'm missing something?

andrews05:
@Zerowalker
I was thinking for maximum flexibility it would be best for the input to be a `PngImage`, with at least the `data` and `ihdr` set. This way the input can be any colour type, including indexed (by setting the `palette`), and can include any additional headers desired such as iCCP etc.

AlexTMjugador:
I've just noticed this PR exists, and a project I maintain could benefit from this change, as it also encodes intermediate images generated by `imagequant` and other optimization techniques to PNG, only for them to be decompressed back by `oxipng`.

I'll give these changes a shot and let you know my thoughts about its API as an user :heart:

src/lib.rs
```diff
@@ -448,6 +448,30 @@ pub fn optimize_from_memory(data: &[u8], opts: &Options) -> PngResult<Vec<u8>> {
...
+        return Err(PngError::new(&format!(
+            "Data length {} does not match the expected length {}",
+            raw.data.len(),
+            expected_len
+        )));
+    }
+
+    let deadline = Arc::new(Deadline::new(opts.timeout));
+
+    if let Some(png) = optimize_raw(Arc::new(raw), opts, deadline, None) {
```
AlexTMjugador, line 468: Regarding your question about making `optimize_from_raw` take `PngImage` by reference, I don't think we should do that. Like `Box`es, `Arc`s own the data they point to, and you can't own a reference, so you would have to clone the image. Therefore, even if such a refactor can be done without throwing ergonomics out the window, I think it's out of scope for this PR.

Instead, I propose refactoring `optimize_from_raw` to generically take anything that can be converted to an `Arc<PngImage>`:

```suggestion
pub fn optimize_from_raw<T: Into<Arc<PngImage>>>(raw: T, opts: &Options) -> PngResult<Vec<u8>> {
    info!("Processing from raw image data");

    let raw = raw.into();

    let bpp = (raw.ihdr.bit_depth.as_u8() * raw.channels_per_pixel()) as usize;
    let row_bytes = (bpp * raw.ihdr.width as usize + 7) / 8;
    let expected_len = row_bytes * raw.ihdr.height as usize;
    if raw.data.len() != expected_len {
        return Err(PngError::new(&format!(
            "Data length {} does not match the expected length {}",
            raw.data.len(),
            expected_len
        )));
    }

    let deadline = Arc::new(Deadline::new(opts.timeout));

    if let Some(png) = optimize_raw(raw, opts, deadline, None) {
```

This refactor is relatively simple in comparison, and allows client code to avoid cloning the `PngImage` if it already has it inside an `Arc`, which is likely to be a common case.

src/lib.rs
```diff
@@ -476,22 +500,69 @@ fn optimize_png(
 
     // Do this first so that reductions can ignore certain chunks such as bKGD
     perform_strip(png, opts);
-    let stripped_png = png.clone();
```
AlexTMjugador, line 479: Why remove this variable? Its value was used in the `if let Some((opts, idat_data)) = best { ... }` statement at what's now the `optimize_raw` function, so maybe this introduced an unintended behavior change?

src/lib.rs
```diff
@@ -611,55 +681,37 @@ fn optimize_png(
...
+                "    zc = {}  f = {}  {} bytes",
+                eval_compression,
+                result.filter,
+                idat_data.len()
+            );
+            return Some(result.image);
         }
-    } else if png.idat_data.len() >= idat_original_size {
-        *png = stripped_png;
-    }
```
AlexTMjugador, line 627: `stripped_png` was also used here too.

AlexTMjugador (changes requested):
I've finally gotten around to reviewing this PR, refactoring my application to make use of its changes.

In general, I liked the new API. I found it well-designed, and I had no problems using it to implement my use cases more efficiently :smile: 

However, I found that, in its current state, the PR introduced regressions when trying to optimize [an example file](https://github.com/ComunidadAylas/PackSquash/raw/baab04d023250030e5c4e979e164f46e585323dd/packages/packsquash/src/pack_file/png_file/example.png):

Incorrect output panic

I found that the cause of the problem was that the `png` variable was no longer modified as expected (see the review comments below for more discussion and the fixes I came up with). I think this issue should be fixed before merging.

Other than addressing this review and performing some extra testing, I think that these changes are really neat and useful, and are good to merge! :tada:

Edit: well, in case it was not obvious: before merging, it's also necessary to resolve merge conflicts, and rename the now public `internal_tests` package that contains types such as `PngImage` to something less internal :wink:

andrews05:
Thanks for the review @AlexTMjugador! It'll be great to move this forward again. Of course it's been quite a while since I worked on this, so I'll have to find some time and give myself a refresher before looking at all your comments 😅

> Edit 2: I ran Criterion benchmarks on my application and performance has improved around 50% thanks to this change. I really like this ❤️

That's awesome, great to hear!

src/lib.rs
andrews05: `stripped_png` was used to revert back to the original state if reductions failed. Because of the way the functions have been refactored, it's no longer necessary to revert anything - the value of `png` here in `optimize_png` will only be modified if reductions were successful.

andrews05:
> I found that the cause of the problem was that the `png` variable was no longer modified as expected (see the review comments below for more discussion and the fixes I came up with). I think this issue should be fixed before merging.

Looks like this was just a silly mistake - I've added the `png` assignment back in here.

src/lib.rs
andrews05: Great idea, thanks!

andrews05:
> Edit: well, in case it was not obvious: before merging, it's also necessary to resolve merge conflicts, and rename the now public `internal_tests` package that contains types such as `PngImage` to something less internal 😉

I think I should leave `internal_tests` alone, instead I've just made all the necessary types public in the main package.

I've been wondering how much error checking `optimize_from_raw` should do. Currently it just checks that the data length is correct but there are other things that could be checked.
- Interlaced input is currently unsupported (despite the field being required in the ihdr), due to the data length check. Should guard against this being set? Or should interlaced input be supported? (I was assuming there would be no use case for this)
- Should we guard against invalid color_type/bit_depth combinations?
- Should we require a non-empty palette if color_type is Indexed? Reductions currently assume black for undefined palette entries.

AlexTMjugador:
Great work! My two cents about your questions and plans:

> I think I should leave `internal_tests` alone, instead I've just made all the necessary types public in the main package.

That sounds good. There is no need to couple clients to the internal layout of the project.

> Interlaced input is currently unsupported (despite the field being required in the ihdr), due to the data length check. Should we guard against this being set? Or should interlaced input be supported? (I was assuming there would be no use case for this)

I can't find any realistic use case that requires support for interlaced data in the raw API either. One of the first things that the PNG decoders I have used do is deinterlacing, so by the time you get to know the palette, header, and pixel color information, deinterlacing is already out of the question. I would vote to just not support interlaced input here (but of course make sure that doesn't affect OxiPNG's support for interlaced PNG files via the traditional API).

> Should we guard against invalid color_type/bit_depth combinations? This is not currently done for file inputs either.

If possible, and not too unwieldy, I would try to exploit the type system to statically guarantee that: let the color type or bit depth determine the set of allowed values for the other, using `enum`s or something similar. But if that's too impractical, I'd just leave it alone.

> Should we require a non-empty palette if color_type is Indexed? Reductions currently assume black for undefined palette entries.

An empty palette is just a special case of a palette with more referenced than declared entries, which is an occurrence in several PNG files found in the wild, and OxiPNG already handles them this way. So I think that keeping the current behavior is fine.

> Should `PngImage.alt_headers` allow `None`, so you don't have to construct an empty indexmap?

Constructing an empty `IndexMap` is a cheap operation because it does not allocate heap memory or invoke the hash function. I think that the ergonomic advantages of not using an `Option` here outweigh the potential minimal efficiency gains.

andrews05:
@AlexTMjugador You've made some great points there. Initially I just used the same structs that were used internally as a matter of convenience for getting this API working, but it's true it isn't necessarily the best structure for input in a user-facing API.

To start with, I'm looking into a refactor of the ColorType enum. I'll probably post it as a separate PR if it works out, but here's what I'm thinking:
```Rust
pub enum ColorType {
    /// Grayscale, with one color channel
    Grayscale { transparent: Option<u16> },
    /// RGB, with three color channels
    RGB { transparent: Option<RGB16> },
    /// Indexed, with one byte per pixel representing one of up to 256 colors in the image
    Indexed { palette: Vec<RGBA8> },
    /// Grayscale + Alpha, with two color channels
    GrayscaleAlpha,
    /// RGBA, with four color channels
    RGBA,
}
```

> If possible, and not too unwieldy, I would try to exploit the type system to statically guarantee that: let the color type or bit depth determine the set of allowed values for the other, using `enum`s or something similar. But if that's too impractical, I'd just leave it alone.

My Rust may be a little rusty for this. I'm not sure how to do that without separate enums for each colour type?

AlexTMjugador:
> To start with, I'm looking into a refactor of the ColorType enum. I'll probably post it as a separate PR if it works out, but here's what I'm thinking. The palette and transparency pixel will be moved into here instead of being properties on `PngImage`.

That'd be a great start! :heart: Also, since the image struct used by the raw API is no longer specifically tied to the PNG format (it's feasible to decode e.g. a JPEG image, pass its data to OxiPNG via the raw API, and get a PNG as a result), it could be renamed to something more generic like `Image`.

> My Rust may be a little rusty for this. I'm not sure how to do that without separate depth enums for each colour type?

My vague idea also revolved around using different depth enums, but now that I think about it more, I realize that using different enums to represent mostly the same concept might feel a bit awkward. [I also pondered the idea of using unit structs and traits](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ecd8a4a03fa91d30bfcd560e32d2ce85), but I find that even more uncomfortable because it requires adding generic parameters to `ColorType`. So, unless someone else comes up with a better idea, from my point of view it's either using different enums or abandoning the idea.

andrews05:
@AlexTMjugador I've introduced a new `RawImage` struct for input to the API. This is primarily necessary to omit the interlaced field in Ihdr.
It still needs to be converted into a PngImage though, so in order to avoid copying the data I removed the Into<Arc>. Do you think this is okay, or would the Arc be better (with copying)?

andrews05:
The `aux_headers` field will likely only be useful as part of a png processing where the headers already exist in the appropriate form. In other scenarios, it may be cumbersome to work with.
Should we add an easier way to include, e.g., an ICC profile?

andrews05:
@AlexTMjugador In lieu of a non-awkward way of enforcing valid depth using the type system, I've simply added manual validation of this similar to the length validation. Which raises another question: Is an `Err` the appropriate response here, or should I use assertions?

AlexTMjugador:
> I've introduced a new `RawImage` struct for input to the API. This is primarily necessary to omit the interlaced field in Ihdr. It still needs to be converted into a PngImage though, so in order to avoid copying the data I removed the Into. Do you think this is okay, or would the Arc be better (with copying)?

The only use of `RawImage` in OxiPNG right now is to tear it down to construct a `PngImage` struct. If the objective is to minimize copies, which I find reasonable, `optimize_from_raw` should take a `PngImage` like it did, but the `interlaced` field should be changed to be crate-private. This way external crates would need to call a constructor that forces `interlaced` to `None` to get an instance of `PngImage`, which guarantees the invariant we want without adding moves or clones.

> The `aux_headers` field will likely only be useful as part of a png processing pipeline where the headers already exist in the appropriate form. In other scenarios, it may be cumbersome to work with.
Should we add an easier way to include, e.g., an ICC profile?

I think it'd be reasonable for the API to accept arbitrary [ICC profile data](https://www.color.org/specification/ICC.2-2019.pdf), which can be easily embedded into a [PNG iCCP chunk](https://www.w3.org/TR/2003/REC-PNG-20031110/#11iCCP). Ideally, such an API would avoid imposing extra copies if the external code happens to have a iCCP chunk at hand. However, color management is a complex and not-so-frequent topic, so perhaps such a change would be better off in a separate PR. (By the way, I realized while writing this that OxiPNG does not even try to recompress ICC profile data, so that might be an avenue for further optimization :wink:)

> In lieu of a non-awkward way of enforcing valid depth using the type system, I've simply added manual validation of this similar to the length validation. Which raises another question: Is an `Err` the appropriate response here, or should I use assertions?

I think that `Err` is the most appropriate way to handle that situation. In my opinion, panics are better used to signal unrecoverable situations that should cause the program or thread to stop. Passing an invalid depth is an usage mistake, but it can be reasonably recovered from by e.g. not optimizing some image.

andrews05:
> The only use of `RawImage` in OxiPNG right now is to tear it down to construct a `PngImage` struct. If the objective is to minimize copies, which I find reasonable, `optimize_from_raw` should take a `PngImage` as it does, but the `interlaced` field should be changed to be crate-private. This way external crates would have to call a constructor that forces `interlaced` to `None` to get an instance of `PngImage`, which guarantees the invariant we want without adding moves or clones.

Hm, I think that could work. So you would call `PngImage::new()`, passing in all your parameters, and then pass that into `optimize_from_raw()`. But then that would beg the question: why not just pass all the parameters straight into `optimize_from_raw()` and not have `PngImage` public at all?
Trying to come up with an ergonomic interface is probably the bigger objective here. There are many possibilities and I'm not sure what's best 🤯

> I think it'd be reasonable for the API to accept arbitrary [ICC profile data](https://www.color.org/specification/ICC.2-2019.pdf), which can be easily embedded in a [PNG iCCP chunk](https://www.w3.org/TR/2003/REC-PNG-20031110/#11iCCP). Ideally, such an API would avoid imposing extra copies if the external code happens to have a iCCP chunk at hand. However, color management is a complex and not-so-common topic, so perhaps such a change would be better off in a separate PR. (By the way, I noticed while writing this that OxiPNG does not even try to recompress ICC profile data, so that might be an avenue for further optimization 😉)

This should be very straightforward. We don't need to have any understanding of an ICC profile or colour management, we can just take the profile as input and compress it to create the iCCP chunk, probably not bothering with a profile name. (And you're right about the lack of recompression currently - there's an open issue for this: #432.)

> I think that `Err` is the most appropriate way to handle this situation. In my opinion, panics are better used to signal unrecoverable situations that should cause the program or thread to stop. Passing an invalid depth is an usage mistake, but it can be reasonably recovered from by e.g. not optimizing some image, or ignoring an animation frame (if the input is an animated image format).
> 
> Edit: I've just found [this interesting write-up about panicking in Rust](https://www.lurklurk.org/effective-rust/panic.html). I think it's worth the read when it comes to designing APIs!

Good, that confirms what I was thinking - just wanted to check 🙂

AlexTMjugador:
> Hm, I think that could work. So you would call `PngImage::new()`, passing in all your parameters, and then pass that into `optimize_from_raw()`. But then that would beg the question: why not just pass all the parameters straight into `optimize_from_raw()` and not have `PngImage` public at all? Trying to come up with an ergonomic interface is probably the bigger objective here. There are many possibilities and I'm not sure what's best exploding_head [edit] Thinking about this further, there could be some advantages to your suggested approach. I’ll try some things out and see…

Yes, that's the idea. Some advantages of using a struct instead of passing the necessary parameters to `optimize_from_raw` are that the caller can reuse or clone the `PngImage` struct as it deems fit, and that the complexity of the `optimize_from_raw` signature is minimized.

> This should be very straightforward. We don't need to have any understanding of an ICC profile or colour management, we can just take the profile as input and compress it to create the iCCP chunk, probably not bothering with a profile name. (And you're right about the lack of recompression currently - there's an open issue for this: #432.)

Good! I didn't know that issue existed until now :joy:

andrews05:
@AlexTMjugador Alright, I've implemented the PngImage constructor. The struct renamed to `RawImage` when exposed publicly.
The constructor performs the depth and data length validations. I haven't changed anything with internal `PngImage` or `IhdrData` structs, so all fields are still public. But because `IhdrData` isn't part of the public api, you're forced to use the constructor. (Sure, it's still possible to change values after construction, but we probably don't need to worry about that)
Aux headers are added via `add_png_header`, rather than being passed into the constructor. There's an additional function `add_icc_profile` for convenience.

AlexTMjugador:
I just took another look at the code, and I think the latest changes sound good!

At first, I doubted the by-reference receiver of `RawImage::create_optimized_png` and whether `RawImage` should wrap an `Arc<PngImage>` at all. However, after pondering the code for a bit, I realized that this design is good because it allows applications to cheaply test different OxiPNG option combinations, which is an interesting possibility.

Right now, my only suggestions would be:

- Specify the expected ICC color data format of the `RawImage::add_icc_profile` method in its documentation as precisely as possible. I think it the additional verbosity is warranted because it is possible to accept color profile data from other image formats with possibly slightly different definitions of valid color profiles.
- Rename `RawImage::create_optimized_png` to `RawImage::optimize`. This name is shorter and more consistent with other public methods.

andrews05:
> I just took another look at the code, and I think the latest changes sound good!

Nice, thanks! 🙂

> * Specify the expected ICC color data format of the `RawImage::add_icc_profile` method in its documentation as precisely as possible. I think it the additional verbosity is warranted because it is possible to accept color profile data from other image formats with possibly slightly different definitions of valid color profiles.

To the best of my knowledge, ICC is an independent specification that doesn't change between image formats. You should be able to, e.g., extract a profile from a jpeg and pass it directly into here. We only need to worry about constructing the chunk correctly, which the function takes care of.

> * Rename `RawImage::create_optimized_png` to `RawImage::optimize`. This name is shorter and more consistent with other public methods.

My thought was that the scenario we're dealing with here is a little different. In the other methods, we have an existing png (either file or in memory) and are going to optimise it. Here though, we don't have a png but would like to create one. We aren't just optimising the `RawImage` itself. I'm open to changing it if you still think it would be appropriate though.

If you're happy with everything, would you like to try it out in your application again? There's a couple of minor things I want to tidy, and I should probably write some tests, but other than that I think this is pretty much ready to go. Or at least to come out of draft 😄

AlexTMjugador:
> To the best of my knowledge, ICC is an independent specification that doesn't change between image formats. You should be able to, e.g., extract a profile from a jpeg and pass it directly into here. We only need to worry about constructing the chunk correctly, which the function takes care of.

While that's correct, [there are several major versions of the ICC color profile format specification](https://www.color.org/icc_specs2.xalter), so ICC profiles that target different versions of the standard are not be interchangeable. Therefore, I think it is a good idea to specify the version required by the PNG specification.

The PNG specification is not crystal-clear about the specifics of the supported ICC color profile format version, but its "Normative references"section refers to "Specification ICC.1: 1998-09, File Format for Color Profiles" and its addendum, which suggests that PNG decoders capable of color management should accept profiles that conform to ICC v2.

> > * Rename `RawImage::create_optimized_png` to `RawImage::optimize`. This name is shorter and more consistent with other public methods.
> 
> My thought was that the scenario we're dealing with here is a little different. In the other methods, we have an existing png (either file or in memory) and are going to optimise it. Here though, we don't have a png but would like to create one. We aren't just optimising the `RawImage` itself. I'm open to changing it if you still think it would be appropriate though.

Fair enough! I like the current method name now 😄 

> If you're happy with everything, would you like to try it out in your application again? There's a couple of minor things I want to tidy, and I should probably write some tests, but other than that I think this is pretty much ready to go. Or at least to come out of draft 😄

Sure, I can try the latest changes out. I'll let you know the results when I get to it.

AlexTMjugador:
> This has since been actioned in the draft of the third edition of the spec.

Hmmm... The normative references of that draft no longer mention ICC v2, so technically PNG files with ICC v2 profile data are no longer conformant? I find that messy enough to warrant opening an issue on the PNG spec repo, but OxiPNG can't document unclear requirements either.

andrews05:
The way I understand it is that there is no intentional restriction on what types of ICC are allowed, so we shouldn’t try to impose any here, especially as common encoders/decoders apparently don’t restrict anything either (even libpng doesn’t mention anything in the docs for `[png_set_iCCP](https://github.com/glennrp/libpng/blob/libpng16/libpng-manual.txt)`.

AlexTMjugador:
I agree with your interpretation of the intent of the specification, and given the circumstances I think that OxiPNG's docs on the new method are fine :smile:

However, I think that the PNG specification should explicitly mention it. In my view, a graphics format that emphasizes portability should not be ambiguous about a feature that's all about color reproducibility. (Or maybe it's just me that I can't read the spec properly, though.)

andrews05:
@shssoichiro I think this is finally ready for review 🙂

@AlexTMjugador I moved the sanity checks mod you made into a separate file so that it can be used in tests.

shssoichiro:
Unfortunately it looks like this needs a fairly significant rebase after one of the other changes got merged.

andrews05:
@shssoichiro Done.

@kornelski Do you have any thoughts on this?

src/lib.rs
```diff
@@ -307,6 +314,88 @@ impl Default for Options {
...
+                    interlaced: Interlacing::None,
+                },
+                data,
+                aux_headers: IndexMap::new(),
+            }),
+        })
+    }
+
+    /// Add a png header chunk, such as "iTXt", to be included in the output
+    pub fn add_png_header(&mut self, chunk_type: [u8; 4], data: Vec<u8>) {
```
kornelski, line 372: Header is not a good name, since in PNG iHDR is the header. Everything is a chunk.

Is correct order of chunks enforced somewhere? Some must be before pLTE or before iDAT.

src/lib.rs
```diff
@@ -307,6 +314,88 @@ impl Default for Options {
...
+        // We can guarantee this will succeed - failure indicates a bug
+        let png = Arc::get_mut(&mut self.png).unwrap();
+        png.aux_headers.insert(chunk_type, data);
+    }
+
+    /// Add an ICC profile for the image
+    pub fn add_icc_profile(&mut self, data: Vec<u8>) {
+        // Compress with default compression level
+        if let Ok(mut compressed) = deflate::deflate(&data, 11, &AtomicMin::new(None)) {
```
kornelski, line 381: Maybe take it gzipped, so that compression can be optimized too?

src/lib.rs
andrews05: Good point, changed.
And yes, the output function enforces certain chunks before/after the PLTE.

src/lib.rs
andrews05: Not quite sure what you're meaning, is it common for icc profiles to be provided in gzip form?
There's an existing issue #432 for recompression of iCCP chunks, if that's relevant.

kornelski: Yes, the motivation was to be able to optimize compression of iCCP. I suppose this method is not necessary for it, and an optimizer can use `add_png_chunk(*b"iCCP` instead.


BTW: if this method is not keeping `data` value permanently, then it shouldn't be taking ownership via `Vec`, and take a `&[u8]` slice instead.

andrews05: Yeah if you have an existing iCCP chunk you should use `add_png_chunk`. Future work can implement recompression, independently of this api.

Ah, good call on the slice. I've changed this now.

kornelski (approved):

AlexTMjugador (approved):
I've finally tested the latest commits in my application and I think this PR is ready to go! :tada:

I found the new `RawImage` struct pleasant to use, as it reduced the amount of lines of code I needed to accomplish the same task. I don't have any other suggestions for now. Great work!

andrews05:
Thanks heaps @AlexTMjugador and @kornelski! This is much better now than my original draft 😅

Over to you @shssoichiro 🙂

shssoichiro (approved):

PR #484, opened 2023-01-05
Fix reduction reporting

andrews05:
I made a little mistake in #476: A reduction would always be reported even if the baseline was chosen by the evaluator.
This PR fixes that, plus makes a couple of tidy ups:
- Removed an incorrect comment I added in the grayscale depth reduction (I thought it was true at the time 😅)
- Removed test-167, following up from #481


shssoichiro (approved):
Thanks!

Issue #419, opened 2021-08-23
Implement --terse CLI flag


folknor:
Hi,

Currently, oxipng supports two relevant flags; `-v` for verbose, which as far as I can tell is only useful for debugging of oxipng itself, and `-q` for quiet mode, which completely suppresses any output. 

Here's an example with the default output:
```
$ oxipng oxipng --dir dist/www -p -s -a -D dist/www/*.png
Processing: dist/www/flagg.0f2b78c7.png
    1200x872 pixels, PNG format
    8 bits/pixel, 62 colors in palette
    IDAT size = 3358 bytes
    File size = 3840 bytes
Reducing image to 8 bits/pixel, 36 colors in palette
Trying: 2 combinations
Found better combination:
    zc = 0  zs = 0  f = 5        3067 bytes
    IDAT size = 3067 bytes (291 bytes decrease)
    file size = 3317 bytes (523 bytes = 13.62% decrease)
Output: dist/www/flagg.0f2b78c7.png
```

I think this is exceptionally verbose for something that is part of a build chain.

An option like `--terse` (or something else equivalent) to produce output like for example

```
$ oxipng ...
dist/www/flagg.0f2b78c7.png 13.62%
```

Which would be enough for me.

Or perhaps

```
$ oxipng ...
dist/www/flagg.0f2b78c7.png dist/www/flagg.0f2b78c7.png 3840|3317|13.62%
```
(src dest origsize compressed %)
Or some other way that makes it easy to parse programmatically (perhaps \t separators).

Thank you for an awesome tool, and thank you for reading my ticket!

TPS:
How about #489 as a solution for this?

folknor:
Perhaps. The pull request doesn't show an example of how the output looks with or without.

TPS:
#488 gives more info, but reading #489's diff, it looks like it'd make the `info` level very similar to above.

chrispy-snps:
I would also find this useful, for the same reasons. We process hundreds of images per document, and a summary of one line per image would be perfect.

PR #489, opened 2023-03-29
Refactoring to use appropriate log level for debug logs

### LINKED ISSUE #488: Noisy logs in INFO category during processing
This is a bit of opinioned view, but I'd say that following logs do not belong to INFO category.

```
2023-03-28 09:40:39.853  INFO oxipng: Processing from memory
2023-03-28 09:40:39.868  INFO oxipng:     900x900 pixels, PNG format
2023-03-28 09:40:39.868  INFO oxipng:     4x8 bits/pixel, RGB + Alpha (non-interlaced)
2023-03-28 09:40:39.868  INFO oxipng:     IDAT size = 1287449 bytes
2023-03-28 09:40:39.868  INFO oxipng:     File size = 1312653 bytes
2023-03-28 09:40:40.043  INFO oxipng: Reducing image to 3x8 bits/pixel, RGB (non-interlaced)
2023-03-28 09:40:40.115  INFO oxipng: Trying: Bigrams
2023-03-28 09:40:40.405  INFO oxipng: Found better combination:
2023-03-28 09:40:40.405  INFO oxipng:     zc = 11  f = Bigrams   1096372 bytes
2023-03-28 09:40:40.406  INFO oxipng:     IDAT size = 1096372 bytes (191077 bytes decrease)
2023-03-28 09:40:40.406  INFO oxipng:     file size = 1096450 bytes (216203 bytes = 16.47% decrease)
```

While it is helpful information when you want to inspect process, it would be better to move details into `DEBUG`, or at least I believe so since `INFO` is what I usually use as default for all logs, except when I want to debug problems

DoumanAsh:
Fixes #488

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -466,13 +466,13 @@ fn optimize_png(
     // Print png info
     let file_original_size = original_data.len();
     let idat_original_size = png.idat_data.len();
-    info!(
+    debug!(
         "    {}x{} pixels, PNG format",
         png.raw.ihdr.width, png.raw.ihdr.height
     );
     report_format("    ", &png.raw);
-    info!("    IDAT size = {} bytes", idat_original_size);
-    info!("    File size = {} bytes", file_original_size);
+    debug!("    IDAT size = {} bytes", idat_original_size);
+    debug!("    File size = {} bytes", file_original_size);
 
     // Do this first so that reductions can ignore certain chunks such as bKGD
     perform_strip(png, opts);
@@ -560,7 +560,7 @@ fn optimize_png(
                     None
                 }
             } else {
-                info!("Trying: {}", trial.filter);
+                debug!("Trying: {}", trial.filter);
                 let original_len = idat_original_size;
                 let best_size = AtomicMin::new(if opts.force { None } else { Some(original_len) });
                 perform_trial(&png.filtered, opts, trial, &best_size)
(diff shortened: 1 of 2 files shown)


TPS:
Might also solve #419

src/lib.rs
```diff
@@ -631,27 +631,27 @@ fn optimize_png(
...
-        info!(
+        debug!(
             "    IDAT size = {} bytes ({} bytes increase)",
             png.idat_data.len(),
             png.idat_data.len() - idat_original_size
         );
     }
     if file_original_size >= output.len() {
-        info!(
+        debug!(
```
shssoichiro, line 647: It might make sense to leave this and the other one below it in the else block (i.e. both of the ones for "file size") as "info". I think that still accomplishes the goal of this ticket, will reduce the number of log lines from 11 to 1. Moving this to debug would silence everything, which I'm not sure if that is what we want.

shssoichiro:
Thanks! It looks like the `verbose_mode` test might need to be updated as well.

src/lib.rs
DoumanAsh: Yes, it logs only following:
```
info!("Processing from memory")
info!("Image already optimized")
```

Similarly for both `optimize` and `optimize_from_memory`

But yes, since it is size indication, might be ok to leave it as `info`

DoumanAsh:
@shssoichiro  I applied comments and re-worked test for verbose mode

PR #491, opened 2023-04-07
Update CI

XhmikosR:
* add `workflow_dispatch` for manual running
* add `fail-fast` false in matrix
* specify `persist-credentials: false` for actions/checkout
* update actions to the latest versions
* move cache before installing the toolchain
* cache more stuff
* limit deploy to shssoichiro/oxipng repository
* reindent

Non-whitespace diff: https://github.com/shssoichiro/oxipng/pull/491/files?w=1

---

There's one warning which needs to be fixed later regarding `set-output`:

>The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

The other warning about Node.js 12 actions that were not updated in this PR because there's no newer version, needs to be fixed upstream, e.g. in https://github.com/actions-rs/toolchain:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions-rs/toolchain@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -7,10 +7,13 @@ on:
   pull_request:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
   test:
+    runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         target:
           - x86_64-unknown-linux-gnu
@@ -45,12 +48,25 @@ jobs:
             toolchain: beta
           - target: x86_64-unknown-linux-musl
             toolchain: nightly
-    runs-on: ${{ matrix.os }}
+
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
+        with:
+          persist-credentials: false
       - name: Install musl tools
         run: sudo apt-get install musl-tools
         if: "contains(matrix.target, 'musl')"
+      - name: Cache cargo registry
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+          key: ${{ runner.os }}-${{ matrix.toolchain }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-${{ matrix.toolchain }}-cargo-registry-
       - name: Install ${{ matrix.toolchain }}
         uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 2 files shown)


XhmikosR:
There's one warning which needs to be fixed later regarding `set-output`:

>The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/oxipng.yml
```diff
@@ -58,13 +63,16 @@ jobs:
...
-      - name: Cache cargo registry
-        uses: actions/cache@v1
+      - name: Cache cargo
+        uses: actions/cache@v3
         with:
-          path: ~/.cargo/registry/cache
-          key: ${{ matrix.target }}-${{ matrix.toolchain }}-cargo-registry-${{ hashFiles('**/Cargo.lock') }}
-          restore-keys: |
-            ${{ matrix.target }}-${{ matrix.toolchain }}-cargo-registry-
+          path: |
```
XhmikosR, line 69: This is taken from https://github.com/actions/cache/blob/main/examples.md#rust---cargo

Let me know if you want me to make any changes :)

AlexTMjugador:
> The other warning about Node.js 12 actions that were not updated in this PR because there's no newer version, needs to be fixed upstream, e.g. in https://github.com/actions-rs/toolchain

The repositories at the `actions-rs` organization are abandoned, so it's unlikely that they will be updated soon. I'd suggest migrating to alternative actions, such as [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

XhmikosR:
Yeah, someone else can take care of that later.

XhmikosR:
All right, I made a couple more tweaks. Notes:

1. clippy seems to fail from forks, but the errors aren't fatal. Ideally it should be skipped when not running from the main repo, see https://github.com/actions-rs/clippy-check/issues/2. Another solution would be to move clippy to its own workflow and just run it on pushes only
2. the https://github.com/actions-rs/toolchain action needs to be replaced by someone more familiar with rust later with https://github.com/dtolnay/rust-toolchain
3. cache could be used in the deploy workflow too later
4. bonus: we could upload the binaries for each run so that people can easily try "nightlies"

shssoichiro:
Regarding clippy it seems to still be running the lints against forked repos at least, just failing to post the lints as comments on the PR, which I'm okay with although it would be nice if we could silence the error message.

I'm good with these changes as a starting point. :+1: thanks!

PR #492, opened 2023-04-07
Create dependabot.yml for action updates

XhmikosR:
Should be merged after #491

EDIT: let me know if you want me to add cargo too.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: monthly


shssoichiro (approved):
Thanks, this looks good to me. As for Cargo dependencies, I usually update them in bulk before publishing a release. In the past Dependabot has frustrated me with dependency updates because it creates a separate PR for each individual update, which gets extremely noisy/tedious at times. If there was a way to weekly batch all patch updates as one PR, that would make it more palatable. Dependabot, last I checked, lacks this functionality. We attempted to do this in rav1e using an alternate tool (renovate) and it seems to fail to pick up a majority of the dependencies, so that was also disappointing.

PR #494, opened 2023-04-07
Update .editorconfig

XhmikosR:


AlexTMjugador:
A `.editorconfig` file is not necessary for Rust projects that rely on `rustfmt` for formatting, such as OxiPNG, as `rustfmt` has its own configuration file. Also, `rustfmt` is supported by IDEs and editors with proper Rust support. Why did you open this PR?

PR #496, opened 2023-04-15
Gate dependency on `crossbeam-channel` behind the `parallel` feature

AlexTMjugador:
`crossbeam-channel` is only used if the `parallel` flag is enabled, so it can be gated behind that feature flag to potentially reduce the size of the build dependency tree for dependent projects.

While at it, I've fixed several warnings emitted in `tests/flags.rs`.


shssoichiro (approved):
Good catch, thanks!

PR #497, opened 2023-04-15
Remove now unused `itertools` dependency

AlexTMjugador:
The `itertools` dependency became unused at a3b104a2edd7e7a7d2944d8ee06eacc77dc6da8c, so there is no point in declaring it as a dependency any longer.

This was detected with the help of the [`cargo-udeps`](https://github.com/est31/cargo-udeps) tool. I verified that it was not a false positive.


shssoichiro (approved):
Thanks!

PR #498, opened 2023-04-15
Make dependency on `image` optional

AlexTMjugador:
After PR https://github.com/shssoichiro/oxipng/pull/481 was merged, the `image` dependency became unused when building with debug assertions disabled, as it is only used to implement output sanity checks when such assertions are enabled.

The `image` crate transitively pulls a significant amount of dependencies, so it's useful for OxiPNG users to get rid of them when not needed:

image crate dependency tree

[Cargo does not allow specifying dependencies that are only pulled when debug assertions are
enabled](https://github.com/rust-lang/cargo/issues/7634), so the next best way to give users some flexibility is to gate those debug assertions behind a feature flag.

These changes add a `sanity-checks` feature flag that controls whether the `image` crate and the related sanity checks are compiled in. This feature is enabled by default to keep debug builds useful to catch problems during development.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,10 @@ name = "oxipng"
 path = "src/main.rs"
 required-features = ["binary"]
 
+[[bench]]
+name = "zopfli"
+required-features = ["zopfli"]
+
 [dependencies]
 zopfli = { version = "0.7.1", optional = true }
 rgb = "0.8.33"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ optional = true
 version = "2.1.0"
 
 [dependencies.image]
+optional = true
 default-features = false
 features = ["png"]
 version = "0.24.3"
@@ -65,6 +66,7 @@ binary = ["clap", "wild", "stderrlog"]
 default = ["binary", "filetime", "parallel", "zopfli"]
 parallel = ["rayon", "indexmap/rayon", "crossbeam-channel"]
 freestanding = ["libdeflater/freestanding"]
+sanity-checks = ["image"]
 
 [lib]
 name = "oxipng"
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -78,7 +78,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           args: -- -D warnings
       - name: Run tests
-        run: cargo test
+        run: cargo test --features sanity-checks
       - name: Build benchmarks
         if: matrix.toolchain == 'nightly'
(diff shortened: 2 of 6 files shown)


andrews05:
Nice! I'll note that unused dependencies are omitted on build anyway (at least that appeared to be the case for me), which is why I didn't try resolve the issue any further, but this could be helpful regardless.

Cargo.toml
```diff
@@ -60,9 +65,10 @@ rustc_version = "0.4.0"
 
 [features]
 binary = ["clap", "wild", "stderrlog"]
-default = ["binary", "filetime", "parallel", "zopfli"]
+default = ["binary", "filetime", "parallel", "zopfli", "sanity-checks"]
```
shssoichiro, line 68: Unfortunately this will enable the feature for both release and debug builds. It would be great if Cargo had a feature to enable dependencies only for debug mode, but sadly no such feature exists... what probably makes the most sense is to:
- Leave this feature out of the default set
- Update the CI to enable the `sanity-checks` feature when it runs tests

AlexTMjugador: I left it enabled to make sure that developers using debug builds generated with a simple `cargo run` or `cargo build` didn't accidentally opt-out of potentially useful assertions by default, but I agree that putting it out of the default feature set could also be a good idea. I don't have any hard preference towards either choice. Shall I change it? 😄

andrews05: The output validation has certainly been a great help for me when developing changes to the project, so it would be a shame to not have it on by default in debug builds. But I'm sure I can remember to manually enable it when I need.

As noted above though, there's no real harm in having it enabled in release since it shouldn't make any difference to the binary size. @AlexTMjugador what was your motivation for making this change? Does it build faster with the feature disabled?

AlexTMjugador: Yes, the first build is faster and takes less disk space because Cargo doesn't need to download and build the extra dependencies, even though it is smart enough to remove their traces from the generated executables.

andrews05: Right, makes sense.
So assuming people using it as a dependency won't be using the default options anyway, I would be in favour of keeping it enabled in default. But I won't mind if @shssoichiro prefers otherwise.

shssoichiro: I wouldn't say I have a strong preference, although I think I would lean toward keeping build times down when possible. As long as it's running in CI as a final "safety check", I'm content with that.

AlexTMjugador: Alright, I'm removing it from the default feature set and adding it to CI 👍

shssoichiro (approved):

shssoichiro:
Thanks!

PR #500, opened 2023-04-20
ColorType Refactor

andrews05:
This PR refactors the `ColorType` enum to incorporate both the palette and transparency data, moving them out of `PngImage`. This doesn't really simplify anything but the intention is to improve the interfaces that will end up being public as part of the raw API (see discussion in #482).

`BitDepth` has also been changed slightly to work with the underlying u8 representation along with a `TryFrom` implementation, which allows some simplifications.

Lastly, I've added some convenience functions `ColorType::is_rgb()`, `ColorType::has_alpha()` and `PngImage::bytes_per_channel()` to allow more tidy and consistent code in some places.

The main annoyance is it is not as easy to make assertions about colour types in the tests. My solution here was to define some constants in the test files to compare the header code rather the enum itself.

Ping @AlexTMjugador.

### DIFF
--- a/src/colors.rs
+++ b/src/colors.rs
@@ -1,14 +1,15 @@
+use rgb::{RGB16, RGBA8};
 use std::fmt;
 
-#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+#[derive(Debug, PartialEq, Eq, Clone)]
 /// The color type used to represent this image
 pub enum ColorType {
     /// Grayscale, with one color channel
-    Grayscale,
+    Grayscale { transparent: Option<u16> },
     /// RGB, with three color channels
-    RGB,
+    RGB { transparent: Option<RGB16> },
     /// Indexed, with one byte per pixel representing one of up to 256 colors in the image
-    Indexed,
+    Indexed { palette: Vec<RGBA8> },
     /// Grayscale + Alpha, with two color channels
     GrayscaleAlpha,
     /// RGBA, with four color channels
@@ -22,9 +23,9 @@ impl fmt::Display for ColorType {
             f,
             "{}",
             match *self {
-                ColorType::Grayscale => "Grayscale",
-                ColorType::RGB => "RGB",
-                ColorType::Indexed => "Indexed",
+                ColorType::Grayscale { .. } => "Grayscale",
+                ColorType::RGB { .. } => "RGB",
+                ColorType::Indexed { .. } => "Indexed",
                 ColorType::GrayscaleAlpha => "Grayscale + Alpha",
                 ColorType::RGBA => "RGB + Alpha",
             }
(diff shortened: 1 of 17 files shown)


src/reduction/mod.rs
```diff
@@ -199,10 +201,15 @@ pub fn reduce_color_type(
...
+    if grayscale_reduction
+        && matches!(
+            reduced.ihdr.color_type,
+            ColorType::RGBA | ColorType::RGB { .. }
+        )
+    {
         if let Some(r) = reduce_rgb_to_grayscale(&reduced) {
             reduced = Cow::Owned(r);
-            should_reduce_bit_depth = reduced.ihdr.color_type == ColorType::Grayscale;
+            should_reduce_bit_depth = reduced.ihdr.color_type != ColorType::GrayscaleAlpha;
```
AlexTMjugador, line 212: I'm not sure why this comparison was changed. What's the rationale behind this change?

src/reduction/mod.rs
```diff
@@ -18,10 +18,10 @@ pub(crate) use crate::bit_depth::reduce_bit_depth;
 /// Attempt to reduce the number of colors in the palette
 /// Returns `None` if palette hasn't changed
 pub fn reduced_palette(png: &PngImage, optimize_alpha: bool) -> Option<PngImage> {
-    if png.ihdr.color_type != ColorType::Indexed {
+    let ColorType::Indexed { palette } = &png.ihdr.color_type else {
         // Can't reduce if there is no palette
         return None;
-    }
+    };
```
AlexTMjugador, line 24: This is the first time I see refutable patterns in `let` statements being used in practice, neat! :joy:

AlexTMjugador (commented):
For what it's worth, I haven't tested these changes, but I like the overall approach. Good work! :+1:

src/reduction/mod.rs
andrews05: It was easier than changing it to a `matches!`, which is required to compare with Grayscale 😂
At this point we know it's either Gray or GrayAlpha, so the result is still the same. It's perhaps a bit of an anti-pattern though. Maybe I should add helper functions to `ColorType` to do common matches for rgb/gray/alpha...

andrews05: I've done some additional refactoring here to tidy up and simplify things. The logic is still ultimately the same.

AlexTMjugador: In my opinion, It looks much better now. Nice! :smile:

src/colors.rs
```diff
@@ -1,14 +1,17 @@
...
 
-#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+use crate::PngError;
+
+#[derive(Debug, PartialEq, Eq, Clone)]
 /// The color type used to represent this image
 pub enum ColorType {
     /// Grayscale, with one color channel
-    Grayscale,
+    Grayscale { transparent: Option<u16> },
```
shssoichiro, line 10: Need to make sure we are maintaining the documentation for these fields that were moved:

```
/// The palette containing colors used in an Indexed image
/// Contains 3 bytes per color (R+G+B), up to 768
pub palette: Option<Vec<RGBA8>>,
/// The pixel value that should be rendered as transparent
pub transparency_pixel: Option<Vec<u8>>,
```

I think it's also better to keep the name `transparency_pixel`, even though it's longer, the name `transparent` strongly implies it should be a bool when it's actually a pixel value.

shssoichiro:
Thanks, looks good, just one small request re. clarity/documentation.

src/colors.rs
andrews05: I kept meaning to add that documentation in there but then kept forgetting. Added now :)

I've changed the name to `transparent_shade` for grey and `transparent_color` for rgb, do you think that will work? I thought it might be good to name them differently, since they have different value types.

shssoichiro (approved):

Issue #418, opened 2021-08-17
Leanify tool provides better lossless compression than Oxipng.


RDKRACZ:
https://github.com/JayXon/Leanify build from the source.
`leanify -p` vs `oxipng -arvZ -o max --strip all --fix`. Not all images were better compressed after using leanify on them, but overall results were better than Oxipng ones.

Assets in this repository were used for this test https://github.com/Waoweens/WaoTweaks.

toscompliantname:
>Assets in this repository were used for this test https://github.com/Waoweens/WaoTweaks.



> remove unnecessary item
> 5bdad11
> 22 days ago

nigguh

andrews05:
This should be resolved by #499, which will avoid changing RGBA to indexed in very tiny images like these. Note some of the images here are currently indexed and oxipng will not attempt to convert indexed to RGBA (which might be smaller), so it still isn't optimal but it does give a better overall result than leanify.

I did notice some of the images have some very faint partial transparency in them, which may be unintended. If you were to clean these up it should help improve compression, e.g. by allowing conversion to RGB+trns which might be smaller than RGBA.

TPS:
> This should be resolved by #499

& further improved by #502.  Thanks, @andrews05! 🙇🏾‍♂️

andrews05:
@TPS & further improved again 😄

Issue #72, opened 2017-07-21
Implement possible size optimizations from Color Type from TruePNG


shssoichiro:
https://css-ig.net/articles/truepng.php

> Colortype reductions
> TruePNG does some other trials that are not done automatically (or not done at all) by other tools:
>
- [x] TruePNG can test grayscale if image is graylevel
- [ ] it tests tRNS chunk instead of alpha channel if applicable
- [x] it reduces truecolor+alpha to grayscale+alpha
- [ ] it can choose truecolor, even if image contains less than 256 colors (or less)
- [ ] TruePNG can encode to palette mode even if image is graylevel

fhanau:
I'm not sure if TruePNG does this, but oxipng should also try to do alpha reductions for palette images (for example by making all transparent pixels use the color of the first one, thus reducing palette size, trns chunk size and making the IDAT chunk more compressible).

TPS:
@shssoichiro : #502 mentions that it fixes №4 in the top post.  Could that be checked off? 🙇🏾‍♂️

PR #502, opened 2023-05-04
Minor reduction improvements

### LINKED ISSUE #73: Implement possible size optimizations from Bit Depth from TruePNG
https://css-ig.net/articles/truepng.php

> The bitdepth should match to the colortype, with some exceptions in paletted mode. encoders have the choice to use 8, 4, 2 or 1 bit by pixel. most of the time, the best choice is to select the lower bitdepth, that should decrease the filesize. this feature is included in most of optimizers.
>
> most of encoders, including TruePNG, choose the minimal bitdepth possible considering the number of colors in the image. sometimes, it is better to choose the 8 bits / pixels, even for an image that have 16 colors or less. TruePNG can check that, and it reduces filesize that way.

- [ ] Test 8 bits/pixel even if image could support less

andrews05:
This started off as minor a bug fix but then expanded into a larger refactor.

The bug occurred when a GrayscaleAlpha image with a `bKGD` chunk tried to reduce to Indexed. The bKGD conversion was assuming RGB and would fail the reduction entirely if the chunk wasn't 6 bytes. It now correctly reduces a Grayscale bKGD and thereby allows the reduction to Indexed. (My fault, since I added the GrayAlpha to Indexed support)

However, due to the new found ability to reduce such images to indexed, "issue-60.png" ended up regressing slightly as the indexed version was actually larger than the gray alpha. Ultimately I ended up completely refactoring the reduction and evaluation sequence, combining the `perform_reductions` function with `reduce_color_type` (I've put the code in the reduction mod, mainly to reduce the amount of code in `lib`).
The result is a bit tidier and has a few notable outcomes:
- Grayscale can now be reduced from 16 bit to depths *lower* than 8, where before they would stop at 8. I've added an extra test for this.
- Conversions to Indexed can now be evaluated at 8-bit *as well as* a lower depth where applicable. This can help with small data where 8-bit is sometimes better (e.g. some images in #418). Fixes #73.
- Certain reductions, such as 16-to-8 and rgb-to-gray, are performed unconditionally with no evaluation, as these are just byte removals. These reductions then become the baseline that subsequent reductions are evaluated against, which again can help with small data (previously in a chain of RGBA->RGB->Indexed, the RGB in the middle would not be considered). Checks off item 4 in #72.

The end result of all this is... "issue-52-04.png" now compresses 3 bytes smaller 😂
Performance is essentially unchanged (I did measure a fractional improvement at level 0).


andrews05:
Well, this turned into a larger refactor than I had planned 😆

@kornelski I ended up removing the extra 4-bit evaluation. I managed to find some images which did prefer 4-bit to 2-bit, but (perhaps due to improvements since that code was added) they were actually best at 8-bit (or even rgb). So I think it's not providing any benefit anymore, but do let me know if you disagree.

src/reduction/mod.rs
```diff
@@ -1,247 +1,115 @@
...
-    // Attempt RGBA alpha reduction after palette, so it can be skipped if palette was successful
-    if reduced.ihdr.color_type == ColorType::RGBA {
-        if let Some(r) = reduced_alpha_channel(&reduced, optimize_alpha) {
-            reduced = Cow::Owned(r);
+    // Attempt to reduce to palette
+    if opts.color_type_reduction && !deadline.passed() {
+        if let Some(reduced) = reduce_to_palette(&png) {
+            png = Arc::new(reduced);
+            // Make sure the palette gets sorted (ideally, this should be done within reduce_to_palette)
+            if let Some(reduced) = reduced_palette(&png, opts.optimize_alpha) {
```
kornelski, line 94: can `reduce_to_palette` be moved before line 80 to try `reduced_palette` once?

kornelski (approved):

src/reduction/mod.rs
andrews05: The two calls to `reduced_palette` are mutually exclusive: The earlier one will run if the image is already indexed (the function will simply return if it isn't), while this one will run if we've just converted to indexed.

shssoichiro:
Thanks, looks good, just needs a rebase :+1:

andrews05:
@shssoichiro Thanks, done 🙂

PR #499, opened 2023-04-17
Include palette and transparency size in evaluations

andrews05:
This is super minor but I started it back in January (in relation to #485) and thought I might as well open a PR for it.

For very small data, including the palette size in evaluations can sometimes achieve a better result.

issue-52-02 master: 1379 bytes
issue-52-02 PR: 1171 bytes

issue-52-03 master: 1555 bytes
issue-52-03 PR: 1315 bytes

issue-52-06 master: 249 bytes
issue-52-06 PR: 231 bytes (apparently 32bpp can be smaller than 2bpp!!)

I also tested this on the images mentioned in #418 and found oxipng now gives a better overall result than leanify.

PR #504, opened 2023-05-09
Further small reduction improvements

### LINKED ISSUE #72: Implement possible size optimizations from Color Type from TruePNG

### LINKED ISSUE #418: Leanify tool provides better lossless compression than Oxipng.

andrews05:
The primary purpose of this development is to add a new "reduction" trial for indexed to RGB/grayscale. Implementing this revealed a couple of flaws though, so it ended up a bit bigger:
- The size estimation added in #499 was helpful but I neglected to make use of it outside of the evaluator. I've made changes to take this into account for primary compression trials as well.
- The palette wasn't consistently reduced, which could cause the estimated size to be inflated. I've separated palette reduction and sorting into two different functions. The reduction is unconditional and won't be evaluated. This means truncation of the palette is no longer necessary on output, which simplifies things. "palette_8_should_be_palette_8.png" compresses better now as it prefers the unsorted palette which wasn't previously being reduced.

With both those sorted, the new indexed conversion is able to achieve some small gains. The previously unused test file "palette_8_should_be_grayscale_8.png" now successfully converts to grayscale, saving 141 bytes 🙂 (yeah okay, only 1%).
Note that conversions to a higher bpp (excludes grayscale) will only be evaluated for small files, so performance impact will be minimal.

Closes #418, with further improvements to the results there.

[edit] Updated with support for grayscale to indexed evaluation, so it can convert in either direction. This closes #72, checking off the final item 5.


kornelski (approved):

andrews05:
Just adding some better tests, will update shortly

andrews05:
Gah, I got the background conversion wrong in indexed_to_channels. Fixed now.

shssoichiro:
Not surprisingly, the Raw API merge caused a conflict with this PR 😅

andrews05:
@shssoichiro Sorted 🙂

shssoichiro (approved):
Thanks!

PR #505, opened 2023-05-17
Refactor aux chunk handling

### LINKED ISSUE #417: Thumb properties stripped even without stripping enabled
When oxipng is let lose on a file that contains Thumb::* metadata, things get lost (the example is utilizing that imagemagick's converter is producing Thumb::URI, Thumb::MTime and others automatically according to the [thumbnail specification](https://specifications.freedesktop.org/thumbnail-spec/thumbnail-spec-latest.html)):

```
$ convert -thumbnail 128x128 rose: out.png
$ exiftool out.png |grep Thumb
Thumb Document Pages            : 1
Thumb Image Height              : 46
Thumb Image Width               : 70
Thumb Mimetype                  : image/pnm
Thumb Size                      : 9673BB
Thumb URI                       : file://rose:
$ oxipng out.png
Processing: out.png
    128x84 pixels, PNG format
    4x8 bits/pixel, RGBA
    IDAT size = 19588 bytes
    File size = 20091 bytes
Reducing image to 3x8 bits/pixel, RGB
Trying: 8 combinations
Found better combination:
    zc = 9  zs = 3  f = 5        16773 bytes
    IDAT size = 16773 bytes (2815 bytes decrease)
    file size = 16962 bytes (3129 bytes = 15.57% decrease)
Output: out.png
$ exiftool out.png |grep Thumb
Thumb URI                       : file://rose: 
$
```

Setting `--strip=none` or similar doesn't change anything here.

What's curious is that I find no mention of these particular names (or really anything "thumb" related) in oxipng, but still, that's what I get.

(Use case for reference: I've tried to use https://github.com/jesjimher/genthumbs/ with oxipng instead of optipng, and it took me quite some while to find why the newly generated thumbs weren't taken any more).

[edit: Note about `file://rose:` not being really correct removed, it just distracted from the difference]

### LINKED ISSUE #432: Attempt to recompress iCCP, zTXt, and iTXt

andrews05:
The primary purpose of this PR is to fix #417 by allowing multiple chunks of the same type to be preserved. My approach here required some significant refactoring as well as BC breaks. Here's an overview of the changes:
- Structs/vars etc to now refer to 'chunks' rather than 'headers', since @kornelski pointed out that this is more correct/less confusing.
- `Strip` and `Keep` variants of `StripChunks` (previously `Headers`) now both take an `IndexSet<[u8; 4]>` rather than `Vec<String>` and `IndexSet<String>` respectively.
- Aux chunks are now stored in a vec in `PngData` rather than a map in `PngImage`, with the exception of "bKGD" which is now a dedicated `background` property on `PngImage`. This change removes the need to copy the chunks with every reduction.
- Stripping is now performed on read so we never copy data for chunks that we aren't going to keep. (This means that tests and benches now have to pass in a full `Options` set when working with `PngData`.)
- "iCCP" will now be recompressed (fixes #432).
- "sBIT" is no longer kept with the `Safe` option. It is no longer transformed by reductions and will instead be forcibly removed, along with "hIST", if reductions occur.

Now regarding that last point with forcibly stripping chunks, I should explain a bit more. There are three aux chunk types whose values depend on the colour type/palette of the image: bKGD, sBIT and hIST. These do not affect rendering and can be safely stripped, but if not stripped they may be invalidated by reductions and require complex handling to rewrite them. Oxipng previously did this for bKGD and sBIT but I've only retained bKGD in this PR and there may even be a case for removing this too [update: bKGD is now stripped as well].
- bKGD: I've found no evidence of this actually being used by any software other than IE 6, where people would set a colour for it to show instead of transparency which it didn't support. Gimp is happy to write this chunk using the background colour from your tool palette, but it won't read it. Software built on libpng such as IrfanView will complain if the bKGD is invalid, but still won't actually use it for anything. One of the main annoyances about bKGD is it can actually prevent a reduction to indexed if there isn't enough room in the palette.
- sBIT: This chunk allows a program to know the original depth of the image that the png was created from, so it can restore the data in its original form if it so desires to. This is not a normal thing to do (and I can't really imagine the use case for it). If the colour type has been changed though, there's no record of what the original type was and therefore such restoration is no longer possible. This is especially true if the source was rgb565 or rgba5551 and we've removed some channels. Inkscape appears to write this chunk to tell you that an 8-bit PNG has 8 significant bits, which is of course redundant.
- hIST: This is a historical chunk to allow hardware with limited colour support to prioritise what colours from the palette it should display. It has no relevance today.

I'm certainly interested to hear other opinions on this though. Is it okay to be forcibly stripping sBIT and hIST if the colour type changes? Should I do the same for bKGD to remove a lot of complexity? I'm not even sure the current bKGD handling is totally correct since there's no way to verify it currently.

Other than that, is my overall approach here reasonable?


andrews05:
Huh, so I've just learned that "the number of palette entries must not exceed the range that can be represented in the image bit depth". The bKGD colour is not exempt from this (I previously assumed it was), which gives us a problem: issue-159.png.

@kornelski Added this in #158 where he made sure the palette was truncated to fit the bit depth. However this ended up trading one error with the image for another: the palette length is now valid but the bKGD refers to a non-existent entry. I unwittingly traded the error back again in #466.

It leaves the image in a no-win situation: It's a 1-bit image so the palette can only have two entries, but there's a third colour required for the bKGD. The only way to fix it is to either increase the depth (not a function oxipng currently has) or strip the bKGD chunk - see question in my post above 😂

Any input on this would be appreciated.

kornelski:
I think it's fine to completely drop `bKGD`.

PR #507, opened 2023-05-21
Deinterlace by default

andrews05:
Changes the default options to deinterlace, with a new `-i keep` cli argument in case anyone wants interlacing to remain unchanged.

This was previously mentioned [here](https://github.com/shssoichiro/oxipng/pull/182#issuecomment-511820902) and I think it's a good thing to do personally, but won't mind if anyone disagrees.


shssoichiro (approved):
Makes sense to me. It's the year 2023.

PR #508, opened 2023-05-21
Add scale16 option to force 16-bit reduction

andrews05:
This adds a new option `--scale16` (or `opts.scale_16 = true`) to force reduction of 16-bit to 8-bit. For improved accuracy it does this by scaling the values rather than just dropping bytes, matching libpng's `PNG_TRANSFORM_SCALE_16` option. This probably isn't a concern for anyone but it has the benefit that it still passes the sanity checks (apparently the image crate is doing the same scaling).


shssoichiro (approved):
Seems good. Having the option to dither might be nice but can come in a future patch.

PR #510, opened 2023-05-23
Fix grayscale_reduction option

andrews05:
Seems I accidentally removed the grayscale_reduction checks at some point. This adds them back in, so the `--ng` flag works again.


shssoichiro (approved):
I see, so the option to disable it was broken. Makes sense, thanks for the fix.

PR #495, opened 2023-04-08
libdeflater 0.14.0 and minor changes

parasew:
Updated libdeflater from 0.11.0 to 0.13.0. Zopfli now at 0.7.2. Clap from 3.2.20 -> 3.2.23. Tested extensively and here are [reproducible performance improvements and demo](https://gist.github.com/parasew/8a4fc4bd9253b7270d908e0a23a75d7e)

PR #511, opened 2023-05-23
Allow APNG with reductions disabled

andrews05:
This is a bit of a hack, not sure if it's something that should actually be added, but it provides rudimentary support for APNG files by disabling all reductions and preserving original chunk order. See #79.
The savings to the test file `apng_file.png` are indeed minimal, achieving only 0.19% with default options and 0.41% with `-o6 -a`.

Note the required chunks are retained with strip safe, though you could still opt to strip all to turn it into a standard png.

### DIFF
--- a/src/headers.rs
+++ b/src/headers.rs
@@ -86,7 +86,9 @@ pub enum StripChunks {
 
 impl StripChunks {
     /// List of chunks that will be kept when using the `Safe` option
-    pub const KEEP_SAFE: [[u8; 4]; 4] = [*b"cICP", *b"iCCP", *b"sRGB", *b"pHYs"];
+    pub const KEEP_SAFE: [[u8; 4]; 7] = [
+        *b"cICP", *b"iCCP", *b"sRGB", *b"pHYs", *b"acTL", *b"fcTL", *b"fdAT",
+    ];
 
     pub(crate) fn keep(&self, name: &[u8; 4]) -> bool {
         match &self {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,7 +389,7 @@ impl RawImage {
     /// Create an optimized png from the raw image data using the options provided
     pub fn create_optimized_png(&self, opts: &Options) -> PngResult<Vec<u8>> {
         let deadline = Arc::new(Deadline::new(opts.timeout));
-        let mut png = optimize_raw(self.png.clone(), opts, deadline, None)
+        let mut png = optimize_raw(self.png.clone(), opts, deadline.clone(), None)
             .ok_or_else(|| PngError::new("Failed to optimize input data"))?;
 
         // Process aux chunks
@@ -399,7 +399,7 @@ impl RawImage {
             .filter(|c| opts.strip.keep(&c.name))
             .cloned()
             .collect();
-        postprocess_chunks(&mut png, opts, &self.png.ihdr);
+        postprocess_chunks(&mut png, opts, deadline, &self.png.ihdr);
 
         Ok(png.output())
     }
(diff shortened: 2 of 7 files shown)


TPS:
The greatest advantage to this is, when wholesale running over large #s of PNGs, is to avoid absolutely destroying APNGs in the process.  That, in & of itself, should increase OxiPNG's uptake & acceptance immediately.

@javiergutierrezchamorro 🙇🏾‍♂️ Comments, please?

andrews05:
@TPS Oxipng currently skips APNGs entirely, so not sure what destruction you're referring to?

I've just added recompression of fdATs which is pretty straightforward and gets us up 0.74% savings on the apng test file. This is still very limited though as reductions are still disabled and it won't re-filter the fdATs. This is all possible in theory and could come some day, but again this is more of a quick hack to get something working.

TPS:
@andrews05 I clarified my meaning above; sorry for the confusion.

I mean that, much of the implicit premise of #79 ("Supported" ≠ "Optimized") is that `OxiPNG` lives in a world of static PNGs.  I'm not aware of anything in the docs that mentions it's APNG-aware, much less that it's already safe for such.  When did that happen?

This PR takes it to the next step: the very tiniest beginning of APNG optimization.  That, combined w/ a note in the ReadMe & probably a brief mention in the built-in help text, is really welcome. 🙇🏾‍♂️

andrews05:
@TPS Ah, looks like it happened after that issue was created (#88), so understandable that people weren't aware.

Yeah, it could definitely do with a note in the Read Me. There's a bunch of other updates that should be made to the Read Me as well but it's tied in with a benchmark script that I've been reluctant to delve into. I might check it out after #495 lands, which will affect benchmarks.

TPS:
@andrews05 I love the iterative nature of how you put together your PRs.  Did https://github.com/shssoichiro/oxipng/commit/070c736d6532eaffa03eb1ab62730e0b81fd02ca crack 1% savings?

andrews05:
@TPS Ha, most of the time it's because I keep making mistakes that I need to fix 🤣. In this case the CI failed with the parallel feature disabled because I used a function that hadn't been shimmed for non-parallel yet. The new commit doesn't affect output.

shssoichiro:
This looks like a nice improvement, even though the compression savings are relatively small. How safe is this to run in bulk? The primary reason why oxipng has continued to skip APNG files is because there has not been a way to verify the resulting image is identical to the input--above all, I want to ensure oxipng never causes damage to a file (unless the user is explicitly using non-default settings that are noted as possibly altering the image).

andrews05:
The sanity checks will cover the main IDAT. I'm not sure how you would verify the rest of the animation though. I tested it on about a dozen images and am reasonably sure it's doing everything right, but hard to be certain 🤷‍♂️

andrews05:
Right, I've added validation of all frames 🙂

javiergutierrezchamorro:
Cool. Looking forward to updated binaries and see if ready for prime time!

andrews05:
I've now tested on 32 files - still not a huge amount but I am pretty confident that it's safe.
I've also added some warnings about stripping APNGs with `--strip all`.

TPS:
> I've now tested on 32 files - still not a huge amount but I am pretty confident that it's safe.

@andrews05 Here's a bunch you could try: https://commons.m.wikimedia.org/wiki/Category:Animated_PNG_files

andrews05:
@TPS Oh thanks, that's really helpful. I've now tested on around 350MB of APNGs, all successful 🙂

TPS:
In your tests on these, did it still hold to <1% reduction?  Anything noteworthy?

andrews05:
Overall still <1% but I think some individual files got up to 8%.

ace-dent:
@andrews05 - you probably have this already, but a APNG test corpus is here: https://philip.html5.org/tests/apng/tests.html

andrews05:
@ace-dent Thanks, I did come across that. That seems to be more to do with correct decoding and rendering though. I think it will be helpful when full support eventually gets implemented but for now we aren't actually decoding the acTL/fcTL chunks.

shssoichiro (approved):

PR #512, opened 2023-05-27
Update Zopfli and several other depenencies

AlexTMjugador:
I've just published a new Zopfli release, v0.7.3, which includes several new features and internal refactors. Performance should be a tad bit better, but I didn't test it throughly, so YMMV. Perhaps more importantly for OxiPNG, its dependency tree is smaller, and Gzip-exclusive compression code can be excluded at compile time thanks to new feature switches. As I mentioned on https://github.com/shssoichiro/oxipng/pull/495#issuecomment-1552669505, I plan on delivering more significant Zopfli performance improvements at some point, but for now I think it's good to give the new release more real-world usage and testing 😄

While at it, I've upgraded other dependencies that are not performance-critical to their latest semver-compatible versions. This excludes `libdeflater` on purpose, as its performance characteristics are said to be somewhat different.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9,12 +9,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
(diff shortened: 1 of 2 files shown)


shssoichiro (approved):
Thanks!

Issue #515, opened 2023-05-31
Chooses 2-bit grayscale over 1-bit indexed


Pr0methean:
When I optimize a 4096x4096, 4x8-bit RGBA image where each pixel is either transparent or opaque black, it is output as 8-bit grayscale, even though it could actually be 1-bit indexed (as it is when the opaque pixels are the same non-gray color). Likewise when the opaque pixels are 0xaaaaaa, it is output as 2-bit grayscale, and 0x515151, 8-bit grayscale. Only white seems to avoid this issue.

Pr0methean:
NB: I experienced this with 8.0.0, and it looks to me like this may already be fixed in master; if so, when can we expect a release?

Pr0methean:
Update: as of 86fccf0 this is fixed for all but the case of 0x515151, which is still 2-bit grayscale.

andrews05:
Regarding the 2-bit grayscale: Currently to keep things simple and avoid additional reduction attempts, it won't try low-depth indexed if it can do low-depth grayscale - it makes an assumption here that the low-depth grayscale will be better (which most of the time it is). Oxipng has extensive reductions, but not exhaustive.
The black preferring 1-bit indexed sounds odd though, could you post the image?

ace-dent:
@andrews05 I've seen this before and it relates to the palette optimization / `0` filter interaction I mentioned. 
The savings with setting edge color to palette entry `0` can outweigh the overhead of storing a palette. IMO it is worth checking indexed image in addition to grayscale (perhaps just for higher optimization levels?). I have a large collection of 1 bit images [here](https://github.com/ace-dent/8x8.me/tree/main/previews).

Pr0methean:
Here's the image:
![black_stained_glass_pane_top](https://github.com/shssoichiro/oxipng/assets/4961925/6ea31d4f-932f-45aa-a363-b63fb8a4c7d0)

Pr0methean:
As a workaround, I'm testing writing the original PNG in grayscale or indexed mode in cases where I can know all the colors based on how the image was generated.

andrews05:
Oh I see, it's caused by the tRNS chunk which is set to a grey level of 1. If it was 255 (white) then it would successfully reduce to 1-bit, but oxipng doesn't currently perform this kind of optimisation of tRNS values. Were there other tools you tried that successfully reduced this?

Pr0methean:
[...]

PR #520, opened 2023-06-18
Try to pick tRNS value that's valid at low depth

andrews05:
In #515, an example image is black and transparent. Currently oxipng will pick a tRNS value of 0x01 and then later find it is unable to reduce the bit depth in grayscale, ultimately choosing 1-bit indexed instead.
This changes the tRNS selection so it will prioritise black, white and two shade of grey, which will be valid at low depth grayscale.


ace-dent:
Nice! Works for 1bpp and 2bpp.
Is there any merit in adding support for 4bpp (needs 12more grays added) ?

andrews05:
Possibly, but I would think it's massively unlikely. Images that fit in 2 or 4-bit grayscale are pretty rare in practice, and one that fits the exact criteria here even more so.

shssoichiro (approved):

PR #522, opened 2023-06-20
Provide BufferedZopfliDeflater and allow user to pass in a custom Deflater

### LINKED ISSUE #521: Use Zopfli DeflateEncoder with customizable buffer size
I have a program that creates large PNGs (up to 48 MiB of raw pixel data, but usually 12 or 16 MiB) and uses Zopfli as the deflater for PNG encoding. At these sizes, compression ratios are usually between 99.4% and 99.9%. Compression is currently slow because of the splits at every ZOPFLI_MASTER_BLOCK_SIZE (currently 1 million) bytes. When I raised this issue at https://github.com/zopfli-rs/zopfli/issues/18, they suggested to instead use a DeflateEncoder wrapping a large BufWriter. Could we please have the option to do that? The best way would be to have the user supply a mutably borrowed DeflateEncoder, so that we could reuse the buffer for multiple images.

Pr0methean:
Fixes #521.


Pr0methean:
This currently fails `tests::raw::from_file`; may I please have some help fixing it?

shssoichiro:
Sorry, I ended up breaking the PR when trying to rebase since you created your PR off a branch called master. Git kind of hates it when there are two branches named the same thing. I recreated the PR off a new branch #530

PR #524, opened 2023-06-22
Fix deadlock in parallel mode

### LINKED ISSUE #517: Parallel mode hangs when invoked from Rayon global thread pool
My app needs to generate and optimize multiple images in parallel, so it uses the Rayon global thread pool. However, the time these images take varies unpredictably, so utilization at the end would be higher if I could also use Oxipng's parallel mode. Unfortunately, Oxipng deadlocks when I invoke oxipng from within a Rayon task in the global thread pool. Here's the app where I'm experiencing this: https://app.circleci.com/pipelines/github/Pr0methean/OcHd-RustBuild?branch=oxipng-parallel

andrews05:
When oxipng is run within an existing rayon context, there may not be any spare threads to execute the evaluations, causing a deadlock. This change adds a yield loop so that the main thread can execute the evaluations when necessary.
Fixes #517

### DIFF
--- a/src/evaluate.rs
+++ b/src/evaluate.rs
@@ -85,7 +85,10 @@ impl Evaluator {
     #[cfg(feature = "parallel")]
     pub fn get_best_candidate(self) -> Option<Candidate> {
         let (eval_send, eval_recv) = self.eval_channel;
-        drop(eval_send); // disconnect the sender, breaking the loop in the thread
+        // Disconnect the sender, breaking the loop in the thread
+        drop(eval_send);
+        // Yield to ensure evaluations are finished - this can prevent deadlocks when run within an existing thread pool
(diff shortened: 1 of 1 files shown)


Pr0methean:
When I tested this in https://github.com/Pr0methean/OcHd-RustBuild/commit/7f2f5f3be7b34f50300752b6d9a8c3102c1f4a42 (which was also using #524 and https://github.com/zopfli-rs/zopfli/pull/21), I got a stack overflow on 64x64 images.

andrews05:
Oh, I thought you said it was working? Or is just specific sizes that do/don't work?

Pr0methean:
It seems to depend in a weird way on what else I'm running.

andrews05:
I suspect it may not be an issue with oxipng. You may have to debug that one yourself.

shssoichiro (approved):

PR #525, opened 2023-06-25
Update rust and dependencies

andrews05:
- Updated minimum rust version from 1.61 to 1.65. I don't know if there's a particular policy here, but I figure we don't want to require the absolute latest (1.70) if not necessary. We do need at least 1.64 for the new versions of Clap/IndexMap and I thought 1.65 would be nice so we can use refutable lets.
- Updated Clap to v4. This was a major update which required significant changes in main.rs.
- Updated IndexMap to v2. No code changes required for this.
- Updated all other dependencies, including libdeflater 0.14. Supersedes #495. (See https://github.com/shssoichiro/oxipng/pull/495#issuecomment-1605080490 for benchmarks)

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -22,7 +22,7 @@ jobs:
           - x86_64-apple-darwin
         toolchain:
           # Minimum stable
-          - "1.61.0"
+          - "1.65.0"
           - stable
           - beta
           - nightly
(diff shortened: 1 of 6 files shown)


shssoichiro (approved):
Thanks, this is really helpful!

At some point I'd love to migrate to the derive parser for clap, but it's a low priority. Getting onto v4 is a huge and welcome improvement.

PR #530, opened 2023-07-08
Provide BufferedZopfliDeflater and allow user to pass in a custom Deflater

shssoichiro:


andrews05:
Heh, so I guess it's a little late now, but thought I should comment here anyway as I have some concerns with this PR.
@Pr0methean
- As noted in the comments, part of the code is "forked" from zopfli-rs. I'm not sure what's different, but surely it would be better to get the required changes submitted to the zopfli library? It doesn't seem like a great idea to keep partially forked code here.
- There's a notable speed regression at level 0 as it no longer skips the final (redundant) compression trial.
- The API seems a little confusing/more difficult now. You set the deflater in the options yet you must also pass in an additional deflater to the optimize function. Although this is only done for the raw API - what about the memory/file APIs?
- Unsure of the actual benefits of this? You said the intention was to improve performance yet it didn't actually achieve that, right?

shssoichiro:
If it's particularly problematic I can always revert it back out, although I think it's better if we can fix the issues with it here. Sorry, it's rough trying to keep up with all the changes coming in. Everyone's working too hard 😂

andrews05:
Hm, I do think it would be best to back it out for now. It may take some time to work through the concerns, especially if some of the changes need to happen in the zopfli library.

PR #534, opened 2023-07-10
Revamp CI workflow to upload artifacts, cross-compile ARM64 binaries, and more

AlexTMjugador:
As commented in issues https://github.com/shssoichiro/oxipng/issues/444 and https://github.com/shssoichiro/oxipng/issues/518, there is some user interest for distributing binaries for each unstable commit, and target ARM64 platforms. Personally, I think both suggestions are useful for the project, as uploading binary artifacts for each commit might help interested users to catch regressions and give feedback earlier, and powerful ARM64 platforms are becoming increasingly popular due to some cloud services (e.g., Amazon EC2, Azure VMs, Oracle Cloud) offering cheaper plans for this hardware, in addition to the well-known push for ARM by Apple with their custom M1 chips.

These changes make the CI target ARM64 as a first-class citizen. Because the public GitHub actions runners can only be hosted on x64 for now, I resorted to cross-compilation, [Debian's multiarch](https://elinux.org/images/d/d8/Multiarch_and_Why_You_Should_Care-_Running%2C_Installing_and_Crossbuilding_With_Multiple_Architectures.pdf), and QEMU to build, get ARM64 C library dependencies, and run tests, respectively.

When the CI workflow finishes, a release CLI binary artifact is now uploaded, which can be downloaded from the workflow run page on the GitHub web interface.

In addition, these changes also introduce some cleanup and miscellaneous improvements and changes to the CI workflow:

- Tests are run using [`nextest`](https://nexte.st/) instead of `cargo test`, which substantially speeds up their execution. (On my development workstation, `cargo test --release` takes around 10.67 s, while `cargo nextest run --release` takes around 6.02 s.)
- The dependencies on unmaintained `actions-rs` actions were dropped in favor of running Cargo commands directly, or using `giraffate/clippy-action` for pretty inline annotations for Clippy. This gets rid of the deprecation warnings for each workflow run.
- Most CI steps are run with a nightly Rust toolchain now, which allows to take advantage of the latest Clippy lints and codegen improvements. In my experience, when not relying on specific nightly features or compiler internals, Rust does a pretty good job at making it possible to rely on a rolling-release compiler for CI, as breakage is extremely rare and thus offset by the improved features.
- The MSRV check was moved to a separate job with less steps, so that it takes less of a toll on total workflow run minutes.

## Pending tasks

- [x] Generate universal macOS binaries with `lipo` (i.e., containing both `aarch64` and `x64` code)
- [x] Tirelessly fix the stupid errors that tend to happen when deploying a new CI workflow for the first time
- [x] Think what to do with the `deploy.yml` workflow. Should it fetch artifacts from the CI job instead of building them again?
- [x] Maybe bring back 32-bit Windows binaries. Are they actually useful for somebody, or just a way to remember the good old days?

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,9 @@
+# Remove this if targeting AArch64 from an AArch64 Linux box
+[target.'cfg(all(target_os = "linux", target_arch = "aarch64"))']
+runner = 'qemu-aarch64'
+
+[target.aarch64-unknown-linux-gnu]
+linker = 'aarch64-linux-gnu-gcc'
+
+[target.aarch64-unknown-linux-musl]
+linker = 'aarch64-linux-musl-gcc'
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,157 +5,76 @@ on:
     tags:
       - 'v*.*.*'
 
+permissions:
+  actions: read
+  contents: write
+
 jobs:
-  create-windows-binaries:
-    runs-on: windows-latest
-    if: github.repository == 'shssoichiro/oxipng'
-
-    strategy:
-      matrix:
-        conf: [x86_64]
-# Temporarily disable i686 binaries, they are failing on linking libdeflate
-# and I don't have a Windows machine set up to experiment with fixing it.
-#        conf: [x86_64, i686]
-
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          persist-credentials: false
-
-      - name: Install stable
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          target: ${{ matrix.conf }}-pc-windows-msvc
-          override: true
-
-      - name: Build oxipng
-        run: |
-          cargo build --release --target ${{ matrix.conf }}-pc-windows-msvc
-
-      - name: Get the version
-        shell: bash
-        id: tagName
-        run: |
-          VERSION=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
-          echo "::set-output name=tag::$VERSION"
-
-      - name: Build package
-        id: package
-        shell: bash
-        run: |
-          ARCHIVE_TARGET="${{ matrix.conf }}-pc-windows-msvc"
-          ARCHIVE_NAME="oxipng-${{ steps.tagName.outputs.tag }}-$ARCHIVE_TARGET"
-          ARCHIVE_FILE="${ARCHIVE_NAME}.zip"
-          mv LICENSE LICENSE.txt
-          7z a ${ARCHIVE_FILE} \
-               ./target/${{ matrix.conf }}-pc-windows-msvc/release/oxipng.exe \
-               ./CHANGELOG.md ./LICENSE.txt ./README.md
-          echo "::set-output name=file::${ARCHIVE_FILE}"
-          echo "::set-output name=name::${ARCHIVE_NAME}.zip"
-
-      - name: Upload artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: ${{ steps.package.outputs.name }}
-          path: ${{ steps.package.outputs.file }}
-
-  create-unix-binaries:
-    runs-on: ${{ matrix.os }}
-    if: github.repository == 'shssoichiro/oxipng'
-
-    strategy:
-      matrix:
-        os: [ubuntu-latest, macos-latest]
-        include:
-          - os: ubuntu-latest
-            target: x86_64-unknown-linux-musl
-          - os: macos-latest
-            target: x86_64-apple-darwin
-
-    steps:
-      - uses: actions/checkout@v3
-        with:
-          persist-credentials: false
-
-      - name: Install Rust
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-          target: ${{ matrix.target }}
-          override: true
-
-      - name: Install musl
-        if: contains(matrix.target, 'linux-musl')
-        run: |
-          sudo apt-get install musl-tools
-
-      - name: Build oxipng
-        run: |
-          cargo build --release --target ${{ matrix.target }}
-
-      - name: Strip binary
-        run: |
-          strip target/${{ matrix.target }}/release/oxipng
-
-      - name: Get the version
-        id: tagName
-        run: |
-          VERSION=$(cargo pkgid | cut -d# -f2 | cut -d: -f2)
-          echo "::set-output name=tag::$VERSION"
-
-      - name: Build package
-        id: package
-        run: |
-          ARCHIVE_TARGET=${{ matrix.target }}
-          ARCHIVE_NAME="oxipng-${{ steps.tagName.outputs.tag }}-$ARCHIVE_TARGET"
-          ARCHIVE_FILE="${ARCHIVE_NAME}.tar.gz"
-          mkdir "/tmp/${ARCHIVE_NAME}"
-          cp README.md CHANGELOG.md LICENSE \
-             target/${{ matrix.target }}/release/oxipng \
-             /tmp/${ARCHIVE_NAME}
-          tar -czf ${PWD}/${ARCHIVE_FILE} -C /tmp/ ${ARCHIVE_NAME}
-          echo ::set-output "name=file::${ARCHIVE_FILE}"
-          echo ::set-output "name=name::${ARCHIVE_NAME}.tar.gz"
-
-      - name: Upload artifacts
-        uses: actions/upload-artifact@v3
-        with:
-          name: ${{ steps.package.outputs.name }}
-          path: ${{ steps.package.outputs.file }}
-
-
   deploy:
+    name: Deploy release
+
     runs-on: ubuntu-latest
-    if: github.repository == 'shssoichiro/oxipng'
-    needs: [create-windows-binaries, create-unix-binaries]
+    timeout-minutes: 30
+
+    # Prevent job from running on forks
+    if: ${{ !github.event.repository.fork }}
+
+    strategy:
+      matrix:
+        target:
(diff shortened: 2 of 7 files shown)


andrews05:
Nice! This is really not my area so can't comment too much, but I do have some thoughts.
- I don't think there's any need to create a universal Mac binary, having them separate is fine (and allows people to download a smaller binary).
- Regarding fetching artifacts for deploy.yml, how would you know which is the correct job to fetch them from?
- Uploading the artifacts may impose higher costs, is @shssoichiro okay with doing so?

AlexTMjugador:
It's nice to read your thoughts anyway!

> [...] how would you know which is the correct job to fetch them from?

I can see two options:

- The deploy job always picks artifacts from the latest successful CI job, or the CI job that is linked to the same commit as the pushed tag.
- The deploy job is manually triggered instead, requiring as an input parameter the ID of the workflow run to fetch artifacts from.

> Uploading the artifacts may impose higher costs, [...]

What costs do you mean by this? As far as I know, GitHub Actions do not have any costs for uploading artifacts at public repositories.

andrews05:
> * The deploy job always picks artifacts from the latest successful CI job, or the CI job that is linked to the same commit as the pushed tag.

Definitely shouldn't use the latest successful job, it could be a PR that hasn't been merged yet. If you can match the commit then that sounds like the correct thing to do. Although I'm sure there's no harm in just rebuilding them as it does now.

> What costs do you mean by this? As far as I know, GitHub Actions do not have any costs for uploading artifacts at public repositories.

Ah, all good then!

ace-dent:
@AlexTMjugador - this will be very useful; I'd love to test new features before a general release!
Would be *eternally* grateful if you could also provide w32 builds (discussed in other issue).
Thanks so much for these contributions!

andrews05:
@AlexTMjugador Looks like the workflow failed to upload the windows artifact?

AlexTMjugador:
> Looks like the workflow failed to upload the windows artifact?

Good catch! The upload artifact step doesn't consider the `.exe` extension, so I'll have to modify its input parameters. I'll fix that while I address other pending tasks in this PR.

Apart from that, what are your thoughts on changes such as using nightly toolchains?

andrews05:
Hm, good question. I’ve had some PR workflows catch me out previously for my inadvertent use of experimental features, and I think we should ensure that these checks are retained (along with adherence to the MSRV). As long as this can still be done with the nightly then it would be fine with me.
On the other hand, we could switch to nightly for the deploy workflow (where codegen will be more important) and stick with standard toolchain here.

AlexTMjugador:
> As long as this can still be done with the nightly then it would be fine with me.

That is done with this new workflow too. MSRV checks are run in parallel to the actual build with nightly Rust, so we should get the best from both worlds.

andrews05:
Oh right, you did mention that

AlexTMjugador:
I polished the new workflow a bit and brought back 32-bit Windows binaries, which appears to be an easier task than I anticipated. I suspect that the build problems that OxiPNG saw in the past with 32-bit Windows are no longer happening, which is good. These binaries pass tests and seem to work fine under Wine for me.

I've also decided not to pursue building universal MacOS binaries, as it's reasonable to assume that the choice of architecture is not inconvenient for our target users, given the feedback from @andrews05.

At this point, the only remaining task to mark this PR as ready for merge is to revisit the deployment workflow.

AlexTMjugador:
I have just rebased this PR on top of master and finished refactoring the release deploy workflow, which now picks artifacts from the CI job linked to the same commit as the pushed tag. I tested the changes on a private repository to make sure they didn't surprise anyone at the worst possible time :wink:

Release created by deploy workflow

I'm marking this as ready to review. @shssoichiro, please feel free to give the final approval and merge at your convenience! :rocket:

AlexTMjugador:
Also pinging @ace-dent here in case they want to give a try to the 32-bit (i686) Windows executables!

ace-dent:
@AlexTMjugador - neat! So when this gets merged, will there be nightly builds ...or a build per PR? Is it obvious where to download the files?

AlexTMjugador:
> So when this gets merged, will there be nightly builds ...or a build per PR? Is it obvious where to download the files?

Each commit and PR will get builds done automatically. They can be downloaded from the [workflow run page for the related commit](https://github.com/shssoichiro/oxipng/actions/runs/5661654849), which can be accessed by clicking the green tick icon for a commit and then "Details".

andrews05:
This looks great, @AlexTMjugador 🙂

Just a note to @ace-dent, @TPS and anyone else interested in these builds going forward: As Alex mentioned, the workflow will run for *pull requests* - this means that not only are they unofficial, they don't even reflect the master branch. The PR may be behind master, it may change before it gets merged, it may even get rejected without merging.
There should always be *some* workflow run that matches master but it won't necessarily be clear which one that is (you would have to match it by commit). The point is, treat these lightly and don't use them in production 🙂

TPS:
Could this be changed to or also provide binaries whenever there's a _commit_, not only PRs?

andrews05:
If you're wanting to find builds that correspond to the latest commit on master, you will have to identify them by matching the latest master commit hash to a workflow run. E.g. master is currently at "b4e98a4" so you would look for that value in the list of workflow runs (you can find this run right now but it won't contain any binaries because this new update hasn't gone live yet).

TPS:
> If you're wanting to find builds that correspond to the latest commit on master, you will have to identify them by matching the latest master commit hash to a workflow run.

I did see that, but, I mean, update this to further clarify in the generated binaries' names.
- For just PR builds, have it marked "-PR" & the # here (e.g., for this issue, "-PR534")
- For accepted commits, have it marked "-committed"+hash #

Or, do some variation of semantic versioning, where version could be `major.minor.patch.commited.PR`-type string.

This would make it easier to tell apart the binaries' continuum from production release to bleeding edge.

TPS:
> Of course, as @andrews05 said, these builds may contain breaking and experimental changes.

This is exactly why many would like clear naming of these.  An accepted commit is likely to be of higher stability & quality than 1 that's just a  PR.  Failing regular binary releases (even βs), that's best we power end-users can hope for.  Honestly, whatever scheme you could do would better than no differentiation @ all! 🙇🏾‍♂️

andrews05:
@TPS The distinction is not actually possible. The workflow is run on commits pushed to the PR - when the PR is merged, no new commits are created and thus no workflow is run. So the workflow run that corresponds to the current master actually originated in a PR.

Note that the binaries uploaded here are not intended to serve as convenient “nightly” builds, nor to make up for current lack of regular releases. They’re simply artifacts generated by the automated testing process that those technically minded can find and download for testing.

I realise the lack of releases has been frustrating and hopefully that situation can be improved in future. One of the reasons for the holdup is because there have been breaking changes necessitating a major version bump. This is not something we want to happen regularly - we want to know that we’re unlikely to make further breaking changes in the near future and it can take time to be sure we’re at that point. 
(Disclaimer: I can’t actually speak for @shssoichiro here and some of this is just assumption on my part)

ace-dent:
@AlexTMjugador 
Did artefact upload fail for this recent PR? https://github.com/shssoichiro/oxipng/actions/runs/5673356981

andrews05:
@ace-dent Until *this* PR gets merged, you won't find builds for any other PR.

XhmikosR:
Great job!

One suggestion: shouldn't the "nightly" archive files include LICENSE etc files too like the deploy action does in main?

TPS:
Do we have many security scans added to this (e.g., [ClusterFuzzLite](https://google.github.io/clusterfuzzlite/running-clusterfuzzlite/github-actions/))? _(Obliquely related to #542)_

AlexTMjugador:
> Shouldn't the "nightly" archive files include LICENSE etc files too like the deploy action does in main?

I find adding license texts to non-release artifacts to be a bit excessive, but I would not mind doing so if there is consensus to add them.

> Do we have many security scans added to this [...]?

No, this PR is unrelated to any automated fuzzing. However, I think it might help paving the way for such things in the future.

andrews05:
> > Shouldn't the "nightly" archive files include LICENSE etc files too like the deploy action does in main?
> 
> I find adding license texts to non-release artifacts to be a bit excessive, but I would not mind doing so if there is consensus to add them.

Agreed. To reiterate, these are *not* "nightlies" and shouldn't be treated as such.

ace-dent:
\>bump<
Good to merge? 🙏

AlexTMjugador:
> Good to merge?

For my part, yes, this is good to merge! :rocket:

AlexTMjugador:
I have noticed that the GitHub UI doesn't allow merging this PR because it expects the older CI status checks to pass, but they no longer exist, so they can't pass:

Missing old CI checks

This is normal, but it'll be necessary to merge this PR manually, or temporarily change the repository settings to not require those checks. I guess @shssoichiro is already aware of this, but I wanted to be extra clear about it :smile:

shssoichiro:
Thanks, I'll temporarily disable the old checks then reenable them after they run in master (this is how github makes you do it)

shssoichiro:
The ARM debian builds seem to be failing, unable to download index files for apt-get update it seems. I'm not a debian expert, is Jammy definitely supported for ARM?

AlexTMjugador:
Thanks for merging! :heart:

> The ARM debian builds seem to be failing, unable to download index files for apt-get update it seems. I'm not a debian expert, is Jammy definitely supported for ARM?

Yes, it should be. I've used Jammy runners for cross-compiling Rust projects before. The workflow ran fine weeks ago, and I'm not aware of any relevant change in the Debian/Ubuntu repositories.

However, this isn't the first time I've seen this kind of weird download errors when cross-compiling. I guess non-x86 Debian repositories are just a bit more brittle: for me, they ended up fixing themselves after a few days. For now, I'd just ignore these errors, but I will check back in a few days and take further action if they continue to happen.

AlexTMjugador:
I have just directly pushed a fix for the build issue, everything should be green now :tada:

PR #535, opened 2023-07-11
Switch from stderrlog to env_logger to resolve security issue with atty

andrews05:
The `atty` dependency of `stderrlog` is unmaintained with an outstanding security issue that we need to address. @AlexTMjugador has apparently filed a PR with `stderrlog` to switch to `is-terminal` but there doesn't appear to be any movement on it. Instead I thought it might be good to switch to `env_logger` which is more flexible, better maintained, and much more widely used.

The only difference to the user is some changes to colours:
Info is now uncoloured (was yellow) - I've done this explicitly as oxipng uses it for normal user output
Warn is now yellow (was magenta)
Debug is now blue (was cyan)
Trace is now cyan (was blue)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,17 +58,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


AlexTMjugador (approved):
Looks good to me, it's good to see someone noticed my PR 😂

PR #536, opened 2023-07-11
Added support for glob patterns in quotes

### LINKED ISSUE #373: Windows: Cannot specify infile in double quotes
When trying to specify files for optimization in double quotes (e.g. when path to file(s) contain spaces),
`oxipng -o max "folder with spaces\*.png"`
 then oxipng throws error "Failed to open file for reading".
Double quotes work though when used in option
`--dir "folder name with spaces\and subfolder"`
Is there a chance for this to be fixed?
Thanks in advance

SalOne22:
Added support for glob patterns in windows paths with spaces that surrounded with quotes

This pull request can be closed if you decided to not change the input files behavior.  Because this can be less robust than `wild` crate. In `wild` special characters in quotes ("*") are intentionally not expanded.

closes #373


src/main.rs
```diff
@@ -383,6 +391,16 @@ fn collect_files(
     in_out_pairs
 }
 
+#[cfg(windows)]
+fn apply_glob_pattern(path: PathBuf) -> Vec<PathBuf> {
+    let input_path = path.to_str().unwrap();
+
+    glob::glob(input_path)
+        .expect("Failed to read glob pattern")
+        .map(|path| path.unwrap())
```
andrews05, line 400: If you change this to `flatten` it can skip errors without panicking.
```suggestion
        .flatten()
```

andrews05 (changes requested):
It's unfortunate that windows neither handles this automatically nor even has a "standard" way of dealing with it, but I think this is a good change as it seems glob is more widely used than wild.

Question: Is it still possible to provide a path containing a literal `*`?

Cargo.toml
```diff
@@ -52,9 +52,9 @@ version = "1.7.0"
 optional = true
 version = "4.3.8"
 
-[dependencies.wild]
+[dependencies.glob]
```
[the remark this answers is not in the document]
SalOne22: Updated in c411ab3

src/main.rs
```diff
@@ -383,6 +391,16 @@ fn collect_files(
     in_out_pairs
 }
 
+#[cfg(windows)]
+fn apply_glob_pattern(path: PathBuf) -> Vec<PathBuf> {
+    let input_path = path.to_str().unwrap();
+
+    glob::glob(input_path)
+        .expect("Failed to read glob pattern")
```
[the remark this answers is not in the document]
SalOne22: Added error handling in e87920b

src/main.rs
SalOne22: Updated in e87920b

SalOne22:
Handling `*` on linux is easy, just need to escape that character with \\, but on windows is impossible to escape `*`, and impossible to name a file with `*`.

AlexTMjugador:
I think we need to think more about how to properly fix this.

The [`wild` crate documentation](https://gitlab.com/kornelski/wild/-/blob/master/README.md) has this remark about how trivially expanding each argument one by one might introduce issues with quoting:

> It is more robust than using [glob()](https://lib.rs/crates/glob) on values from `std::env::args()`, because this crate is aware of argument quoting, and special characteres in quotes ("*") are intentionally not expanded.

`clap` may take care of such quotation shenanigans already, but it may not.

> [...] on windows is impossible to escape `*`, and impossible to name a file with `*`.

Actually, it is possible to name a file `*` on Windows if the [Win32 file namespaces are used](https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces), which is something that most applications and users aren't aware of. For example, Java programs can create files with names that make the Windows Explorer choke. And NTFS itself supports those file names without issues: it is possible to mount a NTFS block device on Linux and do `touch '*'` on it to create such a file.

> [...] it seems glob is more widely used than wild.

[`wild` depends on `glob`](https://gitlab.com/kornelski/wild/-/blob/master/Cargo.toml#L22), so I'd argue that its fitness for purpose is not caused by putting `glob` into the equation: it was already there.

How do built-in Windows commands handle these quotations and glob expansion? Given that Windows is being awkward here as it usually tends to be, I think that's the closest thing to a "source of truth" we have for OxiPNG.

SalOne22:
Here is comparison of different programs (cp, oxipng with wild, oxipng with glob)

There is how `cp` command works in PowerShell:
```
PS C:\Users\vladd\Pictures> ls '.\Camera Roll\'


    Catalog: C:\Users\vladd\Pictures\Camera Roll


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        29.04.2023     12:00         640922 edwin-splisser-GC7FLmgU5Z4-unsplash@s.jpg
-a----        11.07.2023     17:31       11979363 edwin-splisser-GC7FLmgU5Z4-unsplash@s.png
-a----        29.04.2023     11:59         640922 edwin-splisser-GC7FLmgU5Z4-unsplash@u.jpg
-a----        11.07.2023     17:31       11979363 edwin-splisser-GC7FLmgU5Z4-unsplash@u.png
-a----        02.04.2023     20:42       15254494 felix-rottmann-n-Ky-79zeXM-unsplash.jpg
-a----        11.07.2023     17:31      132507102 felix-rottmann-n-Ky-79zeXM-unsplash.png
-a----        29.04.2023     12:00        7795303 felix-rottmann-n-Ky-79zeXM-unsplash@s.jpg
-a----        11.07.2023     17:31      127038626 felix-rottmann-n-Ky-79zeXM-unsplash@s.png
-a----        29.04.2023     11:59        7795303 felix-rottmann-n-Ky-79zeXM-unsplash@u.jpg
-a----        11.07.2023     17:31      127038626 felix-rottmann-n-Ky-79zeXM-unsplash@u.png
-a----        02.04.2023     20:42        4714469 gustavo-zambelli-SIC3k8IMhhA-unsplash.jpg
-a----        11.07.2023     17:31       52923463 gustavo-zambelli-SIC3k8IMhhA-unsplash.png


PS C:\Users\vladd\Pictures> cp '.\Camera Roll\*.png' ..\PicTest\
PS C:\Users\vladd\Pictures> ls ..\PicTest\


    Catalog: C:\Users\vladd\PicTest


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----        11.07.2023     17:31       11979363 edwin-splisser-GC7FLmgU5Z4-unsplash@s.png
-a----        11.07.2023     17:31       11979363 edwin-splisser-GC7FLmgU5Z4-unsplash@u.png
-a----        11.07.2023     17:31      132507102 felix-rottmann-n-Ky-79zeXM-unsplash.png
-a----        11.07.2023     17:31      127038626 felix-rottmann-n-Ky-79zeXM-unsplash@s.png
-a----        11.07.2023     17:31      127038626 felix-rottmann-n-Ky-79zeXM-unsplash@u.png
-a----        11.07.2023     17:31       52923463 gustavo-zambelli-SIC3k8IMhhA-unsplash.png
```

this is how `oxipng` with `wild` works:
```
PS C:\Users\vladd\Pictures> oxipng.exe --dir ..\PicTest\ '.\Camera Roll\*.png'
error: The following required arguments were not provided:
    <files>...

USAGE:
    oxipng.exe --dir <directory> <files>...

For more information try --help
PS C:\Users\vladd\Pictures> oxipng.exe --dir ..\PicTest '.\Camera Roll\*.png'
Processing: .\Camera Roll\*.png
Failed to open file for reading
```

version with `glob`:
```
PS C:\Users\vladd\Documents\GitHub\oxipng> cargo r -- --dir C:\Users\vladd\PicTest\ 'C:\Users\vladd\Pictures\Camera Roll\*.png'
    Finished dev [optimized + debuginfo] target(s) in 0.08s
     Running `target\debug\oxipng.exe --dir C:\Users\vladd\PicTest\ "C:\Users\vladd\Pictures\Camera Roll\*.png"`
Processing: C:\Users\vladd\Pictures\Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@u.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash@u.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash.png
Processing: C:\Users\vladd\Pictures\Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@s.png
Processing: C:\Users\vladd\Pictures\Camera Roll\gustavo-zambelli-SIC3k8IMhhA-unsplash.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash@s.png
...
PS C:\Users\vladd\Documents\GitHub\oxipng> cargo r -- --dir C:\Users\vladd\PicTest 'C:\Users\vladd\Pictures\Camera Roll\*.png'
    Finished dev [optimized + debuginfo] target(s) in 0.09s
     Running `target\debug\oxipng.exe --dir C:\Users\vladd\PicTest "C:\Users\vladd\Pictures\Camera Roll\*.png"`
Processing: C:\Users\vladd\Pictures\Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@s.png
Processing: C:\Users\vladd\Pictures\Camera Roll\gustavo-zambelli-SIC3k8IMhhA-unsplash.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash@u.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash.png
Processing: C:\Users\vladd\Pictures\Camera Roll\felix-rottmann-n-Ky-79zeXM-unsplash@s.png
Processing: C:\Users\vladd\Pictures\Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@u.png
...
```

AlexTMjugador:
This is interesting. The results suggest that if native Windows commands don't allow escaping `*` either, then they may not work that well with files containing asterisks in their names. @kornelski, escuse the mention, but given that you authored the `wild` crate and seem knowledgeable about this stuff, can you give some insight of what's the most correct thing to do here?

andrews05:
> I believe we need to think more about how to properly fix this.

You're right, we do need to sure about this and I'll continue trying to research it...

> The [`wild` crate documentation](https://gitlab.com/kornelski/wild/-/blob/master/README.md) has this remark about how trivially expanding each argument one by one might introduce issues with quoting:
> 
> > It is more robust than using [glob()](https://lib.rs/crates/glob) on values from `std::env::args()`, because this crate is aware of argument quoting, and special characteres in quotes ("*") are intentionally not expanded.

I think this issue concerns using glob on other systems where globbing is already performed. Given @SalOne22 has restricted usage here to Windows only, I think the behaviour is correct (as in, as normal/expected as anything else on Windows).

src/main.rs
```diff
@@ -295,7 +295,7 @@ Heuristic filter selection strategies:
     8  =>  BigEnt    Highest Shannon entropy of bigrams
     9  =>  Brute     Smallest compressed size (slow)",
         )
-        .get_matches_from(wild::args());
+        .get_matches();
```
andrews05, line 298: I'll just note that there's a subtle difference to be aware of here: `wild::args()` corresponds to `env::args()`, while `get_matches()` uses `env::args_os()` (corresponding to `wild::args_os()`). It affects what happens with paths that contain invalid UTF-8.
Not suggesting to change anything, just pointing this out.

andrews05:
Everything I can find suggests that while it is technically possible for paths containing `*` or `?` to _exist_, they simply aren't _valid_ in Windows and you shouldn't expect anything to support them. So they should always be treated as wildcards in the shell, regardless of quotes.

That said, I've just realised the glob function can return nothing if there were no matches, _even if no wildcard was used_. So if you do `oxipng /non-existant/path` you confusingly get no output (see https://github.com/shssoichiro/oxipng/issues/170#issuecomment-1074999156).
@SalOne22 I think the function should be restructured something like this, where it falls back to the input path on any error or empty result:
```rust
fn apply_glob_pattern(path: PathBuf) -> Vec<PathBuf> {
    let matches = path.to_str()
        .and_then(|pattern| glob::glob(pattern).ok())
        .map(|paths| paths.flatten().collect::<Vec<_>>());
    match matches {
        Some(paths) if !paths.is_empty() => paths,
        _ => vec![path]
    }
}
```

AlexTMjugador:
> So it makes sense that they should always be treated as wildcards in the shell, regardless of quotes.

That is a reasonable choice, in my opinion. However, in Unix-like shells, it is possible to pass `"paths like"/this*.png` to commands to escape spaces while allowing glob expansion in the latter part. If Windows also allows this syntax, using it would allow OxiPNG to both work with odd filenames and expand globs. Of course, this syntax is arguably strange to most people, it may not work well on Windows, and I reckon I'm biased to handle these tricky filenames properly because I mount my Windows volumes in Linux environments, but maybe that's why `wild` decided not to expand globs in this case.

SalOne22:
Here is some tests from windows PowerShell:
```
PS C:\Users\vladd\Pictures> oxipng.exe -P '.\Camera Roll\'*.png
.\Camera Roll\ is a directory, skipping
Processing: 2022-10-25_21.10.34.png
    1920x1080 pixels, PNG format
    4x8 bits/pixel, RGB + Alpha (non-interlaced)
    IDAT size = 1316423 bytes
    File size = 1316513 bytes
Trying: Entropy
Found better combination:
    zc = 11  f = Entropy   1182298 bytes
    IDAT size = 1182298 bytes (134125 bytes decrease)
    file size = 1182376 bytes (134137 bytes = 10.19% decrease)
Running in pretend mode, no output
Processing: 5120x2880.png
    5120x2880 pixels, PNG format
    3x8 bits/pixel, RGB (non-interlaced)
    IDAT size = 9497362 bytes
    File size = 9500519 bytes
Trying: None
Found better combination:
    zc = 11  f = None      5294159 bytes
    IDAT size = 5294159 bytes (4203203 bytes decrease)
    file size = 5297208 bytes (4203311 bytes = 44.24% decrease)
Running in pretend mode, no output
```
```
PS C:\Users\vladd\Pictures> oxipng.exe -P '.\Camera Roll'\*.png
.\Camera Roll is a directory, skipping
Processing: \*.png
Failed to open file for reading
```
This is tests from cmd:
```
C:\Users\vladd\Pictures>oxipng.exe -P '.\Camera Roll\'*.png
Processing: '.\Camera
Failed to open file for reading
Processing: Roll\'*.png
Failed to open file for reading

C:\Users\vladd\Pictures>oxipng.exe -P ".\Camera Roll\"*.png
Processing: .\Camera Roll"*.png
Failed to open file for reading

C:\Users\vladd\Pictures>oxipng.exe -P ".\Camera Roll"\*.png
Processing: Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@s.png
    4288x3216 pixels, PNG format
    4x8 bits/pixel, RGB + Alpha (non-interlaced)
    IDAT size = 11979306 bytes
    File size = 11979363 bytes
Reducing image to 3x8 bits/pixel, RGB (non-interlaced)
Trying: Entropy
Found better combination:
    zc = 11  f = Entropy   5701194 bytes
    IDAT size = 5701194 bytes (6278112 bytes decrease)
    file size = 5701251 bytes (6278112 bytes = 52.41% decrease)
Running in pretend mode, no output
Processing: Camera Roll\edwin-splisser-GC7FLmgU5Z4-unsplash@u.png
    4288x3216 pixels, PNG format
    4x8 bits/pixel, RGB + Alpha (non-interlaced)
    IDAT size = 11979306 bytes
    File size = 11979363 bytes
Reducing image to 3x8 bits/pixel, RGB (non-interlaced)
```

andrews05:
Can you try `oxipng -r -P C:\` using your branch here?

SalOne22:
```
PS C:\Users\vladd\Documents\GitHub\oxipng> oxipng.exe -r -P C:\
```
(no output)
```
PS C:\Users\vladd\Documents\GitHub\oxipng> cargo r -q -- -r -P C:\
```
(no output)

SalOne22:
Sorry for the question, but what is the `--recursion` flag for? We can use `**/*` pattern to recursively process directories. This works with `wild` as well.

AlexTMjugador:
> Sorry for the question, but what is the `--recursion` flag for? We can use `**/*` pattern to recursively process directories. This works with `wild` as well.

That's a good question! I think that flag can be useful for very dense directory hierarchies, where shell glob expansion would cause a too long command line parameter string.

andrews05:
Hm, must be an unrelated issue. 

Yeah, good question about why recursive exists. @shssoichiro?

AlexTMjugador:
So if the way I see things is correct, we should choose from two approaches here:

- Meet most people expectations and imitate what built-in Windows commands do, which always expand glob characters, at the cost of introducing problems with technically legal but practically odd filenames that contain those characters.
- Keep OxiPNG's glob expansion behavior consistent between platforms, which implies not expanding glob characters between quotes and allows handling odd filenames, but may be a bit more awkward at first.

What do you think it's the right tradeoff here? I'd lean towards the correctness and logical predictability of keeping glob expansion behavior consistent between platforms, but UX is also an important concern.

andrews05:
On a Mac, consistent UX _within_ the platform is more important to me than consistent UX _across_ platforms. I'm bothered by applications that use non-native UIs or follow Windows conventions. A good cross-platform application is one that adapts to each platform to "fit in" appropriately.
Expectations are a little different on the command line as opposed to a GUI app, but it would still be weird as heck if you installed some command that was ported from Windows and it used `\x` for switches instead of `-x`.

Given we are currently breaking people's expectations and making it harder for them to use oxipng, I'm in favour of switching to `glob`. Note also that `".\Camera Roll\"*.png` works in both unix shells and typical Windows commands but doesn't work with `wild`, so the idea that we're currently consistent between platforms doesn't always hold true.

AlexTMjugador:
Alright, at the moment I'm convinced that expanding globs between quotations is the better way to go on Windows. People with advanced needs will bend over backwards with unusual filenames anyway. Let's see if Kornel or Josh have something to add to the discussion.

andrews05 (approved):

andrews05:
@SalOne22 Could you update from master and fix merge conflicts?

Issue #542, opened 2023-07-28
RFC: Removing the `backup` option


andrews05:
Hi all, I'm hoping to get some feedback from users regarding the `backup` option. @AlexTMjugador and I very briefly discussed the possibility of removing this, so I want to outline why the idea came up and some reasons why we might want to do this.

**The issue:**
For user's of the API, the `backup` option is currently in the wrong place. It's provided as part of the main `Options` struct for all uses of the API, even though it's only relevant when outputting to a file. It would be appropriate to move the backup option to be part of the `OutFile` instead but the downside is that this makes the API more complicated to use.

**Why we might want to remove it**
- Removing it would reduce complexity.
- There's a strong possibility that nobody actually uses it.
- There's no particular use case for it.
- It's not typical practice for a program like this to provide such a feature.
- As best I understand, the feature exists in the first place because it was inherited from Optipng, which oxipng was originally based on. This is fine, but doesn't mean we need to keep it today.

**Why we might want to keep it**
- It's not really hurting - the API can happily remain as is.
- Somebody might be using it?

**Alternatives for people who might be using it**
- Output to a separate file/directory so as not to overwrite the original.
- Copy the file beforehand.
- Use proper backup software.
- Don't bother? Oxipng is super reliable these days and with an extensive test suite and no known corruption issues.

Any comments on this would be welcome 🙂

@shssoichiro Definitely keen to get your input also.

TPS:
I fall heavily into the "Keep It" camp **_&_** do whatever you need to make it maintainable.

My biggest reason is something I didn't see above: it's an absolutely breaking change, & typically wouldn't be noticed by someone who updates to this version _(eventually, ahem)_ until something in their workflow breaks, & they must have those presumably backed-up files.

andrews05:
Thanks for the input!
Breaking changes needn't be a concern. That's the whole point of semantic versioning and bumping the major version number - it signals that there are breaking changes and users should check the release notes to see how it might affect them.
There are already breaking changes for the next release, such as deinterlacing by default. Removing the backup option would actually be quite safe in the sense that if you tried to use it then it would error out due to an unrecognised option.

ace-dent:
+1 for removing backup. Seems odd and unnecessary.
Please make sure that oxipng self-destructs if called with unknown parameters! 🔥 
(i.e. User will become aware that `-backup` was removed if oxipng doesn't run and reports an error to the CLI).

andrews05:
@ace-dent good points!
[...]

PR #546, opened 2023-08-05
Remove `backup` and `check` options

andrews05:
Tidy up the API by removing a couple of options we don't really need.
Backup was discussed in #542
Check was discussed in  #439

@shssoichiro Just say if you prefer to keep either of these 🙂

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -155,10 +155,6 @@ pub struct Options {
     ///
     /// Default: `false`
     pub fix_errors: bool,
-    /// Don't actually run any optimizations, just parse the PNG file.
-    ///
-    /// Default: `false`
-    pub check: bool,
     /// Write to output even if there was no improvement in compression.
     ///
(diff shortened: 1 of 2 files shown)


AlexTMjugador (changes requested):
As per https://github.com/shssoichiro/oxipng/pull/529#issuecomment-1727578998, I'd revert the removal of `backup` before merging this. Otherwise, I think it's good to go, great work! :tada:

The removal of the `check` option has been discussed on the related linked issue and I think there is consensus for removing it, given that its functionality can be achieved by cleverly combining other options.

andrews05:
@AlexTMjugador Updated with backup removal notice:
`The --backup flag is no longer supported. Please use --out or --dir to preserve your existing files.`

AlexTMjugador (approved):
Sweet, I think we can merge this then. Great work!

Issue #553, opened 2023-09-19
Increased file sizes with latest commits versus latest stable release


Rainboom-Dash:
I'm getting increased file sizes with the latest commits that I built on Sept 4th versus the latest stable release
I've tested the default libdeflate compression and zopfli and both are increasing file size considerably
![sourceimagetocompress](https://github.com/shssoichiro/oxipng/assets/38089421/adb0e20c-47e9-4112-9883-6497989f5d2c)
This is not the only image that is getting increased file sizes but it is one of the more significant ones that has a few percent increase (maybe because the file size is smaller to begin with)
On average it seems the latest commits decrease file size slightly... but there are some regressions
I thought maybe it was just the libdeflate since the version was changed but it seems zopfli is affected as well...
![withlateststablerelease](https://github.com/shssoichiro/oxipng/assets/38089421/4107613d-fd18-4aae-8586-0f28ba9b6be1)
![withlatestcommits](https://github.com/shssoichiro/oxipng/assets/38089421/f0e3f0ef-e2cb-4129-910c-fbfe1a26f904)
The image I tested with has had optipng run on it (the site I downloaded from automatically does that)

andrews05:
Hi @Rainboom-Dash, thanks for posting this.
This is quite interesting. The image has many duplicate entries in the palette which somehow are helping it to compress better. The current code assumes that duplicate entries are bad and will always remove them to reduce the palette size. Some investigation will be needed to understand this better and see how we can improve it, but in the meantime you can avoid palette reduction by using the `--np` flag.

@ace-dent Do you know anything about deliberate duplicates in a palette like this? I'm pretty sure optipng wouldn't do that.

ace-dent:
@Rainboom-Dash - very interesting file. Where did it come from and do you know what software produced it? 
@andrews05 - never seen this before... will investigate! Metadata suggests software called [GLDPNG](https://github.com/tk3fftk/PocketSaver/blob/main/source/lib/GLDPNG/GLDPNG3.txt) ...? There is always the potential that the input palette is more optimal; so perhaps a `--np` pass should always be run for the most extreme settings?... not sure if some smarter detection is possible...?

Rainboom-Dash:
I downloaded from [https://derpibooru.org/images/3198446](url) which was taken from [https://www.deviantart.com/boxwari/art/Flurry-Musicante-981750954](url) which I just now realized I probably should of downloaded from Deviantart just to get closer to the source

[...]

PR #554, opened 2023-09-22
Combine palette reduction and sorting

andrews05:
This PR resolves the regression issue in #553.
When a palette reduction occurs, we no longer force it to be the baseline. Instead we combine it with the sorting step and only make it the baseline if sorting fails. This way we can retain the original image as the baseline in these situations without changing the number of evaluations (therefore performance remains unchanged).

The mystery of how #553's palette is weirdly optimised remains unsolved.


AlexTMjugador:
This PR looks reasonable, but I guess it's a draft because it's a specific hotfix for the mentioned issue that may cause regressions or unwanted behavior changes in some edge cases?

andrews05:
Yeah, I marked it as draft just because I want to consider my logic a little more. And indeed, it seems almost hacky for one specific image.

Winterhuman:
I have a good number of images that show weird output size variations when OxiPNG is combined with `pngquant` if you need them for analysis (I think the differences are due to the palette optimisations like what this pull tackles), however, I suspect the new release might fix this discrepancy so I might hold off until then.

Let me know if you'd like the images anyways

andrews05:
@Winterhuman You can try out development builds from the workflow runs if you like. Find them on the Actions page, e.g. https://github.com/shssoichiro/oxipng/actions/runs/6305282807

Winterhuman:
Just tried out that build with some test images. Unfortunately, the output size (just `oxipng` on its own) varies widely in comparison to the current version of `oxipng`; it's hard for me to draw any conclusions on if it's better or worst than the current version.

I'll make sure to open a separate issue with what I've found

andrews05:
@AlexTMjugador I've updated the logic here. It all comes down what we do when a palette reduction alters the image data (as opposed to just truncating the palette, for instance). In these cases, the logic goes like this:
1. We want to keep the original image as the baseline, not assume the "reduced" version is necessarily better.
2. The palette is effectively in a random order right now, neither matching the original order nor having been explicitly sorted yet. Therefore, it's unlikely to be "good" and we shouldn't waste time evaluating it until after we sort it.

It's similar to what I first posted but more correct in edge cases.

AlexTMjugador:
Neat! I think your suggested changes are better, but I would hate to rush them into the v9 release and not think them through. For now, let's focus on releasing v9, which already has a lot of changes. A minor release can be pushed relatively quickly with this patch if we need to.

andrews05:
Sounds good! Might help if we can get more reports of similar issues after release.

Winterhuman:
Ran the newest force-push out of curiosity, and it's looking good!

```
Filename:		'section-tags.png'
Original Filesize:	240 bytes
PNGQuant Filesize:	186 bytes
OxiPNG Filesize:	185 bytes	<-	188 bytes
OxiPNG-ZF Filesize:	185 bytes	<-	185 bytes
Oxi-Quant Filesize:	185 bytes	<-	186 bytes
Oxi-ZF-Quant Filesize:	185 bytes	<-	185 bytes
```

Most of the results are a couple bytes of improvement, them staying the exact same, or a few bytes regressions for the non-Zopfli versions. The only notable result being this one which was one of the images included in https://github.com/shssoichiro/oxipng/issues/558 (though it's still an improvement over v8.0.0):

```
Filename:		'section-series.png'
Original Filesize:	197 bytes
PNGQuant Filesize:	160 bytes
OxiPNG Filesize:	160 bytes	<-	163 bytes
OxiPNG-ZF Filesize:	160 bytes	<-	157 bytes
Oxi-Quant Filesize:	160 bytes	<-	160 bytes
Oxi-ZF-Quant Filesize:	160 bytes	<-	157 bytes
```

andrews05:
@Winterhuman The changes you're seeing will likely be due to the libdeflate update rather than anything in this PR. (Libdeflate is always used for evaluations, so can affect the outcome even when using Zopfli).
Good to see more consistency though.

ace-dent:
@andrews05 @AlexTMjugador - sorry to disagree... I think this fix should be in v9 release, as it corrects a 'regression' introduced with the new palette optim code. For User's not backing up their images, they may lose an optimal palette, that may be difficult / impossible to regenerate with future fixes...

AlexTMjugador (approved):
Right, fair point @ace-dent. Some weeks have passed by since this PR was made, and looking at it again with fresh eyes I think it's logic is sound now. I vote we merge it!

PR #555, opened 2023-09-22
Update libdeflate to v1.19

andrews05:
Small improvements, plus the new version no longer requires the 9 bytes of slack space.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -361,18 +361,18 @@ checksum = "f92be4933c13fd498862a9e02a3055f8a8d9c039ce33db97306fd5a6caa7f29b"
 
 [[package]]
 name = "libdeflate-sys"
-version = "0.14.0"
+version = "1.19.0"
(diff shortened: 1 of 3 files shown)


AlexTMjugador (approved):
Thanks, this looks good. In general, how are compression ratios doing with this update?

andrews05:
On average it's fractionally worse (see https://github.com/ebiggers/libdeflate/issues/323) but I don't think it's anything to worry about. E.g. I'm seeing 100 bytes gained in a 35MB test set.

There's no imperative to update of course.

ace-dent:
@andrews05 this mirrors my early test results. From a few hundred images no improvement (v1.18=v1.19). Several images worse. More details to follow...

AlexTMjugador:
Thank you both for your input! I think then that this PR is good to merge. Even if there are no significant improvements or regressions, keeping dependencies updated enables closer collaboration with upstream if the need arises, minimizing our technical debt.

PR #557, opened 2023-09-23
Add `--timeout` notice

Winterhuman:
Closes: https://github.com/shssoichiro/oxipng/issues/556

Adds a statement saying that `--timeout` isn't as useful for compression algorithms which use fewer and slower rounds, which OxiPNG tends to use nowadays.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -256,7 +256,7 @@ fn main() {
         )
         .arg(
             Arg::new("timeout")
-                .help("Maximum amount of time, in seconds, to spend on optimizations")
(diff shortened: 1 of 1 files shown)


Winterhuman:
Let me know if you want me to change the wording, such as "rounds" vs "trials", or "not recommended for compression algoirthms with fewer rounds" vs "recommended for compression algorithms with many rounds".

andrews05:
Thanks @Winterhuman! We don't actually have zlib with many rounds any more, so the distinction is kinda moot. I'd probably just say that it's currently of limited effectiveness, or similar.

andrews05 (approved):

AlexTMjugador (approved):

PR #560, opened 2023-09-28
Update `zopfli` to v0.8.0

AlexTMjugador:
This update brings several parameterization and usage flexibility improvements on Zopfli, allowing users to choose to limit execution time by a number of iterations without improvement, and exposing a more advanced `ZlibEncoder` struct to tune compression block sizes and DEFLATE block types. Some minor microoptimizations were also made.

For now, I don't expect this PR to substantially affect how OxiPNG compresses images using its Zopfli mode, but the additional parameter customization may come in handy for future work improving how Zopfli is used in OxiPNG.

### DIFF
--- a/src/deflate/zopfli_oxipng.rs
+++ b/src/deflate/zopfli_oxipng.rs
@@ -2,9 +2,7 @@ use crate::{PngError, PngResult};
 use std::num::NonZeroU8;
 
 pub fn deflate(data: &[u8], iterations: NonZeroU8) -> PngResult<Vec<u8>> {
-    use std::cmp::max;
-
-    let mut output = Vec::with_capacity(max(1024, data.len() / 20));
+    let mut output = Vec::with_capacity(data.len());
     let options = zopfli::Options {
         iteration_count: iterations.into(),
         ..Default::default()
--- a/src/deflate/zopfli_oxipng.rs
+++ b/src/deflate/zopfli_oxipng.rs
@@ -6,10 +6,10 @@ pub fn deflate(data: &[u8], iterations: NonZeroU8) -> PngResult<Vec<u8>> {
 
     let mut output = Vec::with_capacity(max(1024, data.len() / 20));
     let options = zopfli::Options {
-        iteration_count: iterations,
+        iteration_count: iterations.into(),
         ..Default::default()
     };
-    match zopfli::compress(&options, &zopfli::Format::Zlib, data, &mut output) {
+    match zopfli::compress(options, zopfli::Format::Zlib, data, &mut output) {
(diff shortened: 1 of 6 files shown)


src/deflate/zopfli_oxipng.rs
```diff
@@ -6,10 +6,10 @@ pub fn deflate(data: &[u8], iterations: NonZeroU8) -> PngResult<Vec<u8>> {
 
     let mut output = Vec::with_capacity(max(1024, data.len() / 20));
```
andrews05, line 7: While we're here, I noticed this vec with initially small capacity can negatively impact performance in some cases as it needs to reallocate during compression. I think we could increase to maybe 1/4, or even just 1 as is used (by necessity) for libdeflate.

src/deflate/zopfli_oxipng.rs
```diff
@@ -6,10 +6,10 @@ pub fn deflate(data: &[u8], iterations: NonZeroU8) -> PngResult<Vec<u8>> {
...
     let options = zopfli::Options {
-        iteration_count: iterations,
+        iteration_count: iterations.into(),
         ..Default::default()
     };
```
andrews05, line 11: Do you think this would be a good time to expose the entire options struct via the API? I mean, as long as we're making breaking changes 😅

andrews05 (changes requested):
Great to see these improvements, nice work!

src/deflate/zopfli_oxipng.rs
AlexTMjugador: That sounds good, I'll change it!

src/deflate/zopfli_oxipng.rs
AlexTMjugador: I think we should definitely explore using or exposing more Zopfli compression options at some point, but for now I think it's fine as it is. Gotta leave something for the v10 changelog too :joy:

andrews05: Ha, all good, it's just I think @Pr0methean added that new option in the hopes of using it with oxipng eventually.

andrews05 (approved):

src/deflate/zopfli_oxipng.rs
AlexTMjugador: We'll get to it at some point! :tada:

AlexTMjugador:
OxiPNG tests ran fine and Zopfli v0.8.0 has been released for 2 weeks without known regressions. I've also tested this branch with Zopfli mode more casually and it works fine. I think this is good to merge.

Issue #526, opened 2023-06-30
Add man page for the command-line utility


sorairolake:
I think this is useful for understanding the detailed usage of the command-line utility.

TPS:
Probably should be done concurrently w/ #472.

g0blin-NN:
oxipng --help > oxipng.txt

TPS:
Given so many recent updates (most of which haven't merged, much less released), I'm not aware whether even the command-line help is current.

g0blin-NN:
Even like this? Sorry.

TPS:
I vaguely remember a discussion someplace in these Issues re: how out-of-date the `--help` is, & another PR recently (merged? 🤷🏾‍♂️) that is supposed to (partially? 🤔) solve that.

andrews05:
`--help` has been kept up-to-date with all changes, it's just the Read Me that may be out of date.

@sorairolake Not saying there shouldn't be a man page, but was there something in particular you were wanting to understand that wasn't explained sufficiently in `--help`?

sorairolake:
@andrews05 I also think the usage in `--help` is explained sufficiently. However, I think it would be more useful to be able to see how to use `oxipng` on the web (such as <https://man.archlinux.org/>) or a console.

andrews05:
One easy thing we could do is append the help output to the Read Me (in some sort of automated fashion). This would at least make it easily accessible online.

musicinmybrain:
You might (or might not) find `clap_mangen` ([GitHub](https://github.com/clap-rs/clap/tree/master/clap_mangen), [crates.io](https://crates.io/crates/clap_mangen)) useful for this.

andrews05:
Thanks @musicinmybrain, see #596 where I’m using that 🙂 
I hope to pick that up again soon and see if I can make a deb package…

Issue #472, opened 2022-12-08
Update README


shssoichiro:
The instructions in the README for what options are available are very out of date and need to be updated

ace-dent:
+1

andrews05:
Done

PR #563, opened 2023-10-02
Overhaul help

andrews05:
This PR brings a big overhaul to oxipng's help, with new long form descriptions of many options.

The full output (--help) is added as a text file MANUAL.txt. Critiques welcome.

The short output (-h) is simplified and appears as follows:
```
Losslessly improve compression of PNG files

Usage: oxipng [OPTIONS] <files>...

Arguments:
  <files>...  File(s) to compress (use '-' for stdin)

Options:
  -o, --opt <level>       Optimization level (0-6, or max) [default: 2]
  -r, --recursive         Recurse input directories, optimizing all PNG files
      --dir <directory>   Write output file(s) to <directory>
      --out <file>        Write output file to <file>
      --stdout            Write output to stdout
  -p, --preserve          Preserve file permissions and timestamps if possible
  -P, --pretend           Do not write any files, only show compression results
  -s                      Strip safely-removable chunks, same as '--strip safe'
      --strip <mode>      Strip metadata (safe, all, or comma-separated list)
                          CAUTION: 'all' will convert APNGs to standard PNGs
      --keep <list>       Strip all metadata except in the comma-separated list
  -a, --alpha             Perform additional alpha channel optimization
  -i, --interlace <type>  Set PNG interlacing type (0, 1, keep) [default: 0]
      --scale16           Forcibly reduce 16-bit images to 8-bit (lossy)
  -v, --verbose...        Run in verbose mode (use twice to increase verbosity)
  -q, --quiet             Run in quiet mode
  -f, --filters <list>    Filters to try (0-9; see '--help' for details)
      --fast              Use fast filter evaluation
      --zc <level>        Deflate compression level (1-12)
      --nb                Do not change bit depth
      --nc                Do not change color type
      --np                Do not change color palette
      --ng                Do not change to or from grayscale
      --nx                Do not perform any transformations
      --nz                Do not recompress unless transformations occur
      --fix               Disable checksum validation
      --force             Write the output even if it is larger than the input
  -Z, --zopfli            Use the much slower but stronger Zopfli compressor
      --timeout <secs>    Maximum amount of time to spend on optimizations
  -t, --threads <num>     Set number of threads to use [default: num CPU cores]
  -h, --help              Print help (see more with '--help')
  -V, --version           Print version

Run `oxipng --help` to see full details of all options
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -35,9 +35,11 @@ use std::process::exit;
 use std::time::Duration;
 
 fn main() {
-    // Note: Long help descriptions should wrap at 70 characters (column 71)
-    // Indentation is 10 spaces, so this allows it to fit on an 80 character terminal
-    // Short help descriptions should ideally be no more than 54 characters
+    // Note: clap 'wrap_help' is enabled to automatically wrap lines according to terminal width.
+    // To keep things tidy though, short help descriptions should be no more than 54 characters,
+    // so that they can fit on a single line in an 80 character terminal.
+    // Long help descriptions are soft wrapped here at 90 characters (column 91) but this does not
+    // affect output, it simply matches what is rendered when help is output to a file.
     let matches = Command::new("oxipng")
         .version(env!("CARGO_PKG_VERSION"))
         .author("Joshua Holmer <jholmer.in@gmail.com>")
@@ -55,10 +57,9 @@ fn main() {
             Arg::new("optimization")
                 .help("Optimization level (0-6, or max)")
                 .long_help("\
-Set the optimization level preset. The default level 2 is quite fast
-and provides good compression. Lower levels are faster, higher levels
-provide better compression, though with increasingly diminishing
-returns.
+Set the optimization level preset. The default level 2 is quite fast and provides good \
+compression. Lower levels are faster, higher levels provide better compression, though \
+with increasingly diminishing returns.
 
 0   => --zc 5 --fast               (1 trial, determined heuristically)
 1   => --zc 10 --fast              (1 trial, determined heuristically)
@@ -69,9 +70,8 @@ returns.
 6   => --zc 12 -f 0-9              (10 trials)
 max =>                             (stable alias for the max level)
 
-Manually specifying a compression option (zc, f, etc.) will override
-the optimization preset, regardless of the order you write the
-arguments.")
+Manually specifying a compression option (zc, f, etc.) will override the optimization \
+preset, regardless of the order you write the arguments.")
                 .short('o')
                 .long("opt")
                 .value_name("level")
@@ -91,8 +91,8 @@ arguments.")
             Arg::new("recursive")
                 .help("Recurse input directories, optimizing all PNG files")
                 .long_help("\
-When directories are given as input, traverse the directory trees and
-optimize all PNG files found (files with “.png” or “.apng” extension).")
+When directories are given as input, traverse the directory trees and optimize all PNG \
+files found (files with “.png” or “.apng” extension).")
                 .short('r')
                 .long("recursive")
                 .action(ArgAction::SetTrue),
@@ -101,9 +101,9 @@ optimize all PNG files found (files with “.png” or “.apng” extension).")
             Arg::new("output_dir")
                 .help("Write output file(s) to <directory>")
                 .long_help("\
-Write output file(s) to <directory>. If the directory does not exist,
-it will be created. Note that this will not preserve the directory
-structure of the input files when used with '--recursive'.")
+Write output file(s) to <directory>. If the directory does not exist, it will be created. \
+Note that this will not preserve the directory structure of the input files when used with \
+'--recursive'.")
                 .long("dir")
                 .value_name("directory")
                 .value_parser(value_parser!(PathBuf))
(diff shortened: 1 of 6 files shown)


ace-dent:
Thanks so much for attending to this! :-)
I'd like to review. Can you wait a couple of days for feedback?

andrews05:
Sure thing, happy to wait :)

andrews05:
@AlexTMjugador It might be good for the deploy workflow to include the manual, and maybe even generate and commit the newest version.

AlexTMjugador:
Okay, I'll try to get the deployment workflow to include the manual file today. But I'm not sure what you mean by "generate and commit the newest version", can you clarify that?

andrews05:
@AlexTMjugador Yeah, exactly:
```
oxipng -V > MANUAL.txt
oxipng --help >> MANUAL.txt
```
I've added a shell script to do this which we can run manually, but might be good to automate it during deploy if it's easy enough to do.

src/main.rs
```diff
@@ -172,67 +238,104 @@ fn main() {
         )
         .arg(
             Arg::new("filters")
-                .help(format!("PNG delta filters (0-{})", RowFilter::LAST))
+                .help(format!("Filters to try (0-{}; see '--help' for details)", RowFilter::LAST))
+                .long_help("\
+Peform compression trials with each of the given filter types. You can specify a \
```
ace-dent, line 243: Typo: 'Peform' --> 'Perform'.

src/main.rs
```diff
@@ -99,64 +129,100 @@ fn main() {
...
+but take care as this is technically a lossy transformation and may be unsuitable for some \
+applications.")
                 .short('a')
                 .long("alpha")
                 .action(ArgAction::SetTrue),
         )
         .arg(
             Arg::new("interlace")
-                .help("PNG interlace type - Default: 0")
+                .help("Set PNG interlacing type (0, 1, keep)")
```
ace-dent, line 197: Specify default setting, e.g.:
`"Set PNG interlacing type (0, 1, keep). Default: 0."

src/main.rs
```diff
@@ -99,64 +129,100 @@ fn main() {
...
+                .long_help("\
+Set the PNG interlacing type, where <type> is one of:
+
+0     =>  Remove interlacing from all images that are processed
```
ace-dent, line 201: Specify default setting, e.g.:
`0     =>  Remove interlacing from all images that are processed (default)`

ace-dent:
Fixes #472.
Fixes #526.

Please add:
- Comment on parameters being case sensitive.
- Example of short vs long parameter usage, e.g. `-Psavvo6` == `--pretend --strip=safe --alpha --verbose --verbose --opt=6`.
This fixes #539.

src/main.rs
```diff
@@ -99,64 +129,100 @@ fn main() {
...
+                .help("Strip all metadata except in the comma-separated list")
                 .long("keep")
                 .value_name("list")
                 .conflicts_with("strip")
                 .conflicts_with("strip-safe"),
         )
         .arg(
             Arg::new("alpha")
-                .help("Perform additional alpha optimizations")
+                .help("Perform additional alpha channel optimization")
```
ace-dent, line 185: Down to opinion,but I'd emphasise risk here: 
`"Perform additional, lossy alpha channel optimization"`

src/main.rs
```diff
@@ -99,64 +129,100 @@ fn main() {
...
                 .long("interlace")
                 .value_name("type")
-                .value_parser(["0", "1", "keep"]),
+                .default_value("0")
+                .value_parser(["0", "1", "keep"])
+                .hide_possible_values(true),
         )
         .arg(
             Arg::new("scale16")
                 .help("Forcibly reduce 16-bit images to 8-bit")
```
ace-dent, line 216: Tweak: `"Forcibly reduce 16-bit images to 8-bit"` --> `"Forcibly reduce 16-bit per channel images to 8-bit"`

ace-dent:
@andrews05 - looking good! 🏆

src/main.rs
andrews05: Good catch, thanks

src/main.rs
andrews05: Nice, I've expanded the help here a bit more.

src/main.rs
andrews05: Hm, I'd rather not label it as "lossy" outright, since it's really "visually lossless". I've clarified this in the long help but can't really make the distinction in the short form.

andrews05:
> Fixes #472. Fixes #526.
> 
> Please add:
> 
> * Comment on parameters being case sensitive.
> * Example of short vs long parameter usage, e.g. `-Psavvo6` == `--pretend --strip=safe --alpha --verbose --verbose --opt=6`.
>   This fixes [Tolerate lowercase `-z` parameter #539](https://github.com/shssoichiro/oxipng/issues/539).

I might leave these for the Read Me. The Clap library's help system isn't as well-suited for adding extra information like this (at least if you care about exactly where that info gets inserted).

AlexTMjugador:
Coming back to life...

> I've added a shell script to do this which we can run manually, but might be good to automate it during deploy if it's easy enough to do.

Yeah, this is easy enough to do, I've been pretty busy these days though. Will get to it as soon as I can :smile:

src/main.rs
ace-dent: That's great! Thanks.

AlexTMjugador:
I've just changed the release workflow to generate the manual and include it in the release archive. I think that automatically generating and committing new manuals to CI can get a bit noisy and cumbersome due to the need to pull or rebase new commits, while release workflows are kind of infrequent to run anyway, so I think it's fine if we don't commit anything automatically for now.

andrews05:
Thanks @AlexTMjugador! This should be good to merge if you approve.

AlexTMjugador (approved):
Yeah, I think it's good to go! :tada:

Issue #549, opened 2023-08-06
Allow continue compressing after abortion


LuckyTurtleDev:
If the user has a slow pc or using a high compression, he might can not compress all images at once.
But after aborting it is not possible to continue at the last compressed image anymore.
So I would suggests adding an option to define an archive file (similar to youtube-dl) which store the path of all images, which where already compressed. This would make it possible to continue compressing where you have stopped.

I can implement this, if this feature is wanted.

andrews05:
Hm, I understand your issue but this seems like an incredibly niche feature which I would say falls out of scope of what oxipng should be providing.

Can you explain your scenario a little more? Like, could you just run oxipng on a subset of your files at a time so you know where you're up to?

LuckyTurtleDev:
I have run oxipng at my home folder yesterday, but it has take forever and I have abort it. Today I  started it again, but it starts from the beginning. Because I have absolute no idea where it has stop I can not continue without starting from scratch. (It would also be a pain to define this per argument).

Some thing like `oxipng --archive foo.txt .` would allow to save a list of compressed files and to skip them if I execute the same command again.

> could you just run oxipng on a subset of your files at a time so you know where you're up to?

I think this should also be possible. But it much more complicated and annoying . You do not know how long this would take and if you abort the process you still don not know where you have stop. You must also note/remember which files you have compress before.

ace-dent:
> could you just run oxipng on a subset of your files

I would say this is how *all* png optimizers work. Handling of bulk jobs is normally handled by a script / coded GUI, etc. ... Not sure it should be a core function here, due to portability and maintaining that extra code...

andrews05:
Yeah, I hear the frustration but you should be able to script something yourself outside of oxipng. E.g. you could start by constructing a list of all your png files and then making a script to work through them. Or even simpler, you could try only finding files modified before a given date, so as to exclude ones that were recently optimised (assuming Windows has some sort of `find` equivalent that can do this?).

I'm afraid I just don't agree that this should be built-in to oxipng. That is just my opinion though - @AlexTMjugador or @shssoichiro may have different ideas.

AlexTMjugador:
I agree with @andrews05 and @ace-dent that this feature is out of scope for OxiPNG. In my opinion, tracking the progress of a batch optimization job is better suited to a script or higher level application with more context about the job, and I find the alternative approaches suggested by @andrews05 to be spot on.

[...]

PR #564, opened 2023-10-08
Allow caching a history of optimized files to avoid reoptimizing them

aaronfranke:
This PR adds a command-line argument `--cache` that will cache a history of the hashes of optimized files, and will check this cache before optimizing a particular image to check if it's already optimized.

Problem: Let's say I have a folder of 100 PNG files. I may run `oxipng` on this folder, which takes a long time. Then, I add 10 more PNG files to this folder so that there are 110 files. If I want to re-optimize the folder, running `oxipng` on the whole folder again will re-optimize all 110 files. I would like to optimize the folder without repeating the work already done to optimize the first 100 images, so that it only needs to optimize the 10 new images.

Solution: After optimizing, keep track of optimized files by storing their hashes in a file in the user's cache folder (for example: `~/.cache/oxipng/optimized_file_hashes` on Linux). Before optimizing the next time, check this list to see if the file hash is in the list. If it is, skip it. If not, perform the optimization.

Technical details: This PR adds a dependency on the [`dirs`](https://crates.io/crates/dirs/5.0.1) crate, so that we can access the user's cache folder across platforms. A new option has been added to Options `pub cache: bool`, and some code to parse the `--cache` argument. Aside from that, the cache code is entirely self-contained in a new file named `cache.rs`, called from the `if opts.cache {` blocks, so it does not run at all when `opts.cache` is false.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -198,6 +198,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "5.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "44c45a9d03d6676652bcb5e724c7e988de1acad23a711b5217ab9cbecbec2225"
+dependencies = [
+ "dirs-sys",
+]
+
+[[package]]
+name = "dirs-sys"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
+dependencies = [
+ "libc",
+ "option-ext",
+ "redox_users",
+ "windows-sys",
+]
+
+[[package]]
 name = "either"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -277,6 +298,17 @@ name = "funty"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "e6d5a32815ae3f33302d95fdcb2ce17862f8c65363dcfd29360480ba1001fc9c"
+
+[[package]]
+name = "getrandom"
+version = "0.2.10"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "be4136b2a15dd319360be1c07d9933517ccf0be8f16bf62a3bee4f0d618df427"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "wasi",
+]
 
 [[package]]
 name = "glob"
@@ -449,12 +481,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "option-ext"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
+
+[[package]]
 name = "oxipng"
 version = "8.0.0"
 dependencies = [
  "bitvec",
  "clap",
  "crossbeam-channel",
+ "dirs",
  "env_logger",
  "filetime",
  "glob",
@@ -480,6 +519,24 @@ dependencies = [
  "fdeflate",
  "flate2",
  "miniz_oxide",
+]
+
+[[package]]
+name = "proc-macro2"
+version = "1.0.68"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5b1106fec09662ec6dd98ccac0f81cef56984d0b49f75c92d8cbad76e20c005c"
+dependencies = [
+ "unicode-ident",
+]
+
+[[package]]
+name = "quote"
+version = "1.0.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5267fca4496028628a95160fc423a33e8b2e6af8a5302579e322e4b520293cae"
+dependencies = [
+ "proc-macro2",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


andrews05:
Hi @aaronfranke

Thanks for your work on this. Unfortunately we had a recent discussion in #549 and decided that this kind of feature is out of scope for what we want oxipng to be. Keeping track of state between invocations is more of an "application" level feature, and we prefer for oxipng to remain a basic CLI tool.

This is a great feature for a wrapper application/script that invokes oxipng under the hood and we encourage users who would like such a feature to explore doing it this way.

Apologies for any inconvenience this may cause.

aaronfranke:
@andrews05 I considered making a wrapper, but I figured this would be better to have in the app itself.

It's unfortunate that you are not interested in this - I'll just uninstall oxipng and keep using a copy built from this branch.

andrews05:
Sure, that's also a good option!

AlexTMjugador:
Scope considerations aside, I'm not positive that this implementation of this feature will work as reliably as intended. Hash collisions may happen, and as per the [birthday problem](https://en.wikipedia.org/wiki/Birthday_problem), they are more likely to happen than you might think at first glance. Also, it does not take into account whether images in the destination path have been modified, deleted or moved, which would require the user to delete a somewhat hidden cache file to get OxiPNG to output images again.

For your use case, it sounds like a better approach might be to have one directory per batch of input images, tell OxiPNG to optimize them separately to a common output directory, and then symlink/merge all those input files to a common location.

aaronfranke:
@AlexTMjugador The Rust language only offers 64-bit hashes for the built-in hasher, so it seems like it's good enough for Rust. As for the birthday problem, there are 365 birthdays but 2^64 possible 64-bit hashes, so this is not really a problem. Git tools typically only display 6 to 8 hex characters for the commit hash, but the hashes in this PR have 16 characters, so I think the chance of a conflict is so miniscule it's not even worth considering.

> Also, it does not take into account whether images in the destination path have been modified, deleted or moved, which would require the user to delete a somewhat hidden cache file to get OxiPNG to output images again.

No, it handles this case. If you have image A, then optimize it into B, it will cache the hash of B. Then if you copy B 10 times anywhere you'd like and run oxipng recursively, it will skip all instances of B. The only data that is saved are hashes of already optimized images, not hashes of unoptimized images or file paths.

AlexTMjugador:
> The Rust language only offers 64-bit hashes for the built-in hasher, so it seems like it's good enough for Rust. As for the birthday problem, there are 366 birthdays but 2^64 possible 64-bit hashes, so this is not really a problem. Git tools typically only display 6 to 8 hex characters for the commit hash, but the hashes in this PR have 16 characters, so I think the chance of a conflict is so miniscule it's not even worth considering.

About the choice and usage of hashing function:

- The [`DefaultHasher` docs](https://doc.rust-lang.org/std/collections/hash_map/struct.DefaultHasher.html) state that "the internal algorithm is not specified, and so it and its hashes should not be relied upon over releases". This means that the hashes may change depending on what Rust version OxiPNG is built with, which may cause unexpected collisions or cache misses.
- The [current implementation of `DefaultHasher`](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#3167-3169) uses [SipHash 1-3](https://en.wikipedia.org/wiki/SipHash). Unlike the SHA algorithm used by Git for object hashes, SipHash is a keyed hash function, and the key is set to zero in the used `DefaultHasher::new()` method, meaning [it's relatively easy to cause collisions if you want](https://crypto.stackexchange.com/questions/17996/is-siphash-cryptographically-secure), because the secret permutation key is known. This can cause problems if OxiPNG is exposed to untrusted user input.
- The birthday problem can be generalized to show that, after processing 609 million images with `--cache`, [the chance of a collision happening with 64-bit hashes is comparable to the chance of a full house at poker](https://preshing.com/20110504/hash-collision-probabilities/). This is a lot, but a busy service processing 10 images per second will reach this in ~2 years, and I wouldn't want to be the person debugging such hash collision issues in that production scenario. (This scenario also makes me realize that we should probably have some policy for evicting cache entries; otherwise, the cache can grow indefinitely, and requiring user intervention for cache cleanup in batch jobs is not ideal.)

So... Why take all these chances and add two dependencies (if you count an additional dependency on a fixed hash function) if it's possible to use separate input directories instead?

> No, it handles this case. If you have image A, then optimize it into B, it will cache the hash of B. Then if you copy B 10 times anywhere you'd like and run oxipng recursively, it will skip all instances of B. The only data that is saved are hashes of already optimized images, not hashes of unoptimized images or file paths.

Right, I missed that! Still, I think it's a bit brittle, because the way I see it now, it would not be possible to optimize an already optimized image but with different OxiPNG options when using `--cache` (please correct me if I'm wrong). So from a user perspective, I don't think the ergonomics of `--cache` would be great.

aaronfranke:
> "the internal algorithm is not specified, and so it and its hashes should not be relied upon over releases"

Ah, that's unfortunate. It would be nice to make this more robust.

> This is a lot, but a busy service processing 10 images per second will reach this in ~2 years, and I wouldn't want to be the person debugging such hash collision issues in that production scenario.

That busy service can just not use the `--cache` option.

The target use case here is running `oxipng` on big folders on your system. A service running `oxipng` on arbitrary images would not want to use the `--cache` option, because it's not likely that users will upload already-optimized images to be optimized. After optimizing 609 million images, the file storing the hashes would be 5481 MB in size.

> So... Why take all these chances and add two dependencies (if you count an additional dependency on a fixed hash function) if it's possible to use separate input directories instead?

I don't want to use separate directories, that's too much manual work dealing with them or writing scripts. I want to optimize everything in-place, like this PR allows me to do.

> it would not be possible to optimize an already optimized image but with different OxiPNG options when using `--cache` (please correct me if I'm wrong).

You are correct. I considered storing the settings used but I didn't want to work on solving a theoretical problem, only real practical problems I actually have (the [problem -> solution](https://docs.godotengine.org/en/stable/_images/best_practices1.png) mindset). My thought is that if users are using the `--cache` option they would have already figured out what settings they want to use.

AlexTMjugador:
> You are correct. I considered storing the settings used but I didn't want to work on solving a theoretical problem, only real practical problems I actually have (the [problem -> solution](https://docs.godotengine.org/en/stable/_images/best_practices1.png) mindset). My thought is that if users are using the `--cache` option they would have already figured out what settings they want to use.

It's fair to say that long-running services shouldn't use `--cache` in the first place, or that you don't want to use scripts. But I think it's likely that users will enable `--cache` because they think it's an "option that makes repeated OxiPNG runs faster" and then wonder why OxiPNG refused to optimize a file they "clearly" want to optimize again. Even if they figured out what settings to use beforehand, they may change their mind over time, and that's not good UX, especially if the software documentation and error messages don't explain what's happening ("file was previously optimized", but why?) or suggest corrective actions.

andrews05:
I was going to comment here but thought it better to post it in https://github.com/shssoichiro/oxipng/issues/549#issuecomment-1753568570 instead.

Feel free to continue discussing technical details, but I don't think we need to get into any debates 🙂

PR #578, opened 2023-11-15
Fix and simplify output file permission copying code

### LINKED ISSUE #576: Failed to set permissions, expected: 100666, found: 100644
Hi,

I'm having an issue where files with permissions different than 644 cannot be optimized using both "--preserve" and "--out".
Here are my system info: Arch Linux 64-bit, Kernel 6.6.1-arch1-1, oxipng `9.0.0-1` from the Arch extra repository.

Here are some steps to reproduce the issue:
1. `chmod 666 image.png`
2. `oxipng -o max image.png --preserve --out image-optimized.png`
> image.png: failed to set permissions, expected: 100666, found: 100644

I tried to create the destination file with correct permissions before running oxipng tool and this worked.
1. `touch image-optimized.png`
2. `chmod --reference=image.png image-optimized.png` 
3. `oxipng -o max "image.png" --preserve --out "image-optimized.png"`

I tried to look in [oxipng/src/lib.rs](https://github.com/shssoichiro/oxipng/blob/master/src/lib.rs) where the error is raised (line 1046) but I sadly don't know much about Rust and can't tell if something is off or if it's my system that's improperly configured.

I haven't seen similar opened issues, sorry if this is a duplicate! 
Also, this is my first time using Github, excuse me if anything is wrong with this report.

Thank you in advance for your help and have a nice day!

AlexTMjugador:
A detailed read to the [Rust documentation for the `fs::Permissions` struct](https://doc.rust-lang.org/stable/std/fs/struct.Permissions.html) and a little digging into its implementation in the standard library have shown that this code didn't work as expected in any platform, and was a bit weird to begin with:

- It first read the permissions from the input file metadata.
- Then it fetched the output file metadata.
- After that, it changed the permissions for that output file metadata.
- It then performed a sanity check that the output file had the expected permissions.

Barring the fact that the sanity check in step 4 is not needed, the overall approach is wrong because setting the permissions in a file metadata struct does not actually persist those changes anywere; it's just an in-memory change only, so these operations were useless. The Rust documentation explicitly mentions that the `set_readonly` method "does not modify the files attributes" [sic], but it's easy to miss that warning and not realize that it also applies to the methods offered by the `PermissionsExt` trait. The code only appeared to work because in most cases the default permissions for new files happen to match the input file permissions, so the sanity check passed.

To fix this, use the `set_permissions` method on `File` to actually set the output file permissions to be the same as the input file permissions, which is both much simpler and robust.

These changes were tested in the context of issue #576, and fix #576.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -21,6 +21,7 @@
 
 #[cfg(feature = "parallel")]
 extern crate rayon;
+
 #[cfg(not(feature = "parallel"))]
 mod rayon;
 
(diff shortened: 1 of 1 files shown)


Jaiyce87:
@AlexTMjugador I just tried this fix on my machine using the `Oxipng binary (x86_64-unknown-linux-gnu)` artifact and it seems to work perfectly! Thank you very much!  😀

AlexTMjugador:
Thank you both, merging this then!

PR #580, opened 2023-11-18
Allow strip and keep to be used together

andrews05:
This is a minor change that allows using both `--strip` and `--keep` at the same time.

E.g. `--strip safe --keep eXIf` will strip chunks while preserving both the ones that aren't "safe" to remove *and* eXIf. Essentially it's a convenience to allow extending the default list used by `--strip safe`.

Specifying chunk names for both options is not permitted, e.g. `--strip eXIf --keep eXIf` will error.

Use of `--strip all` with `--keep` is redundant, but is permitted.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -176,9 +176,7 @@ depth is changed, regardless of any options set.",
             Arg::new("keep")
                 .help("Strip all metadata except in the comma-separated list")
                 .long("keep")
-                .value_name("list")
-                .conflicts_with("strip")
-                .conflicts_with("strip-safe"),
+                .value_name("list"),
         )
         .arg(
             Arg::new("alpha")
@@ -613,14 +611,6 @@ fn parse_opts_into_struct(
         };
     }
 
-    if let Some(keep) = matches.get_one::<String>("keep") {
-        let names = keep
-            .split(',')
-            .map(parse_chunk_name)
-            .collect::<Result<_, _>>()?;
-        opts.strip = StripChunks::Keep(names)
-    }
-
     if let Some(strip) = matches.get_one::<String>("strip") {
         if strip == "safe" {
             opts.strip = StripChunks::Safe;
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
LGTM, thanks!

ace-dent:
@andrews05 - please can you update Docs and builtin help? 😇
It's not clear to me the order of priority /precedence. Can I do `--keep aCEd --strip all` to remove everything but my select chunk?

andrews05:
Heh, yeah I didn't think about that until after I posted the PR. I'll make some changes...

Argument order is irrelevant, your example will keep the selected chunk. Perhaps we could prohibit keep with strip-all to reduce any confusion there?

ace-dent:
> Perhaps we could prohibit keep with strip-all to reduce any confusion there?

I think being able to combine both is useful. Perhaps stating that 'keep overrides strip' in the docs is sufficient?

andrews05:
Hm, what makes you say it could be useful?
Normally invalid combinations would be prohibited. In the case of `--strip all --keep` it’s not so much invalid as redundant, but it would still be preferable to prohibit it rather than having to explain in the docs that keep will override if you do that.

ace-dent:
Perhaps I misunderstood...
```bash
--strip all --keep # Useless!
--strip all --keep aCEd # Has some use!
```

andrews05:
Sorry, I should have put a value in there - it is required to have one, so (1) is indeed already prohibited 😆

But I'm still unclear why you're suggesting `--strip all --keep aCEd` has some use. `keep` is overriding, so `--strip all` is necessarily redundant and you should just use `--keep aCEd`.

ace-dent:
I should read the manual .... 🤦‍♂️ !!
Please ignore my comments, I was thinking of another program...

Yes: a combination of `--strip all` and `keep` should be prevented. 
Sorry again 🙃

andrews05:
Ha, all good 🙂
Not sure what `keep` would do if it _didn't_ strip everything else? 🤔

Anyway, I'll make the changes...

PR #581, opened 2023-11-19
Disallow strip-all with keep

andrews05:
Followup to #580, this prohibits `--strip all --keep <list>` to prevent any confusion as to what this might do. The only valid combo is `--strip safe --keep <list>`. Help for keep has been updated to show it can be used in this way.

Note I removed an error check in strip parsing as it really didn't seem necessary.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -175,6 +175,9 @@ depth is changed, regardless of any options set.",
         .arg(
             Arg::new("keep")
                 .help("Strip all metadata except in the comma-separated list")
+                .long_help("\
+Strip all metadata chunks except those in the comma-separated list. May be combined with \
+'--strip safe' to extend the list of chunks to be preserved.")
                 .long("keep")
                 .value_name("list"),
         )
@@ -622,12 +625,6 @@ fn parse_opts_into_struct(
             let names = strip
                 .split(',')
                 .map(|x| {
-                    if x == "safe" || x == "all" {
-                        return Err(
-                            "'safe' or 'all' presets for --strip should be used by themselves"
-                                .to_owned(),
-                        );
-                    }
                     let name = parse_chunk_name(x)?;
                     if FORBIDDEN_CHUNKS.contains(&name) {
                         return Err(format!("{} chunk is not allowed to be stripped", x));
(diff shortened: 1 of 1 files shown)


andrews05:
Thinking about this a little more, what if instead of strip-safe we had a special term to use with keep?
So it would be `--keep <some_short_term_representing_chunks_preservered_by_strip-safe>` 😆
And then you could do `--keep <term>,eXIf`, which I think would be easier to understand than `--strip safe --keep eXIf`, assuming we can come up with a good term.

The chunks preserved are those that are *not* "safely-removable", as they affect rendering:
cICP, iCCP, sRGB, pHYs, acTL, fcTL, fdAT.

I'm struggling to come up with any good suggestions myself. They could be considered "important" (as opposed to "critical", which are the ones that can't be removed)?

ace-dent:
My 2¢: It's fine as it is and not worth making a breaking change for.

andrews05:
Not suggesting any breaking changes, just adding an alternative way to do strip-safe. If it works well we could possibly consider removing strip-safe in some future version when other breaking changes are made, but certainly not right now.

AlexTMjugador:
> Thinking about this a little more, what if instead of strip-safe we had a special term to use with keep? So it would be `--keep <some_short_term_representing_chunks_preservered_by_strip-safe>` 😆 And then you could do `--keep <term>,eXIf`, which I think would be easier to understand than `--strip safe --keep eXIf`, assuming we can come up with a good term.

I like this idea! I vote we revert #580 and go for this. I had to read the source code and your discussion in detail to understand that combining `--strip` and `--keep` was more confusing than I anticipated, which is definitely not a good sign for UX and not something I think will happen by adding this extension to `--keep` instead.

I suggest the term "safe" for this, it's short and accurate.

andrews05:
It's so tempting to just use safe, but I'm not convinced that it's accurate. The current use of "safe" with strip means stripping all chunks that are "safely-removable". The chunks that are kept then are those that are *not* safe to remove. So it would seem more accurate to use "unsafe", but that obviously sounds bad 😂
I'm not totally opposed to "safe" though, it might work if we word the help right...

AlexTMjugador:
As you say, `--strip safe` removes all the safe-to-remove chunks (more precisely, all the non-critical chunks except for a few defined by the standard that affect rendering), so I don't think it's inaccurate to use that word there if we make that meaning clear in the docs. But maybe it'd be clearer at a first glance if it used the slightly longer `safe-to-remove` form.

About `--keep`, we are looking for a word that represents the set of all the critical chunks, plus standard non-critical chunks that are are known to affect rendering (i.e., are not safe-to-remove). I think the word `presentation` could accurately describe this set, since every chunk in it has the ability to affect how the image is presented by a feature-rich, standards-compliant decoder. (To make the vocabulary used in these two options more consistent, `safe-to-remove` could be replaced with `non-presentation`).

andrews05:
Yeah, presentation would be the right idea. It's quite long though. Maybe "display" could work?

AlexTMjugador:
I think "display" sounds good and is short enough. `visual`, `render` or `key` are a bit more ambiguous in interpretation in my opinion.

PR #582, opened 2023-11-20
If only one filter selected use it for evaluations

andrews05:
The filters used during reduction evaluations are normally set to None & Bigrams, regardless of any options. This PR makes a slight change so that if only one filter is specified in the options, this filter will be used for reduction evaluations too.

This resolves an odd situation affecting lower levels (when --fast is enabled) where you may try to force the filter to a specific value but it actually ends up different because a reduction evaluation was smaller. It's particularly helpful if you're wanting it to be as fast as possible by using `-o0 -f0` which will now exclusively use None instead of trying the slower Bigrams as well.

As another use, you could try to brute force oxipng by iterating each filter separately, though this may not actually achieve anything 😂

[edit] I also pulled the options out into a separate file, though this wasn't relevant to the filter change.


AlexTMjugador:
I like the overall idea of this PR, but it makes me reflect on how we should handle the case of selecting exactly two filters. What are your thoughts on just cloning the input filters for evaluation if there is either one or two?

andrews05:
Yeah, I thought about that for a bit, and then what about cases where there are more than two input filters but still excluding None or Bigrams, and it all got a bit complicated. I ended up doing it this way for two reasons:
- The case of only one filter is more likely/more important than other non-default filter selections.
- The standard 0,7 is actually still a good choice for some combos such 0,9.

AlexTMjugador:
Sounds fair enough to me!

> The standard 0,7 is actually still a good choice for some combos such 0,9.

I guess this observation could be extended to mean that the standard 0,7 is also a good proxy for combos of most other filter combinations? How big is the _some combos_ set here?

andrews05:
Generally speaking I'd say 7 is a good proxy for any heuristic filter (5-9). Hypothetically, it may not be optimal if you only chose non-heuristic filters (0-4). Similarly, 0,7 may not be optimal if your combo excludes 0. I haven't exactly performed any analysis to test these though and I imagine any impact would be pretty small.
These situations would also be quite unexpected with no obvious use-case, so I felt it wasn't worth the complexity required to account for them.

AlexTMjugador (approved):
Okay, I'm convinced, let's move forward with this. Thanks for the PR!

PR #584, opened 2023-11-30
Add `--keep display` equivalent to `--strip safe`

andrews05:
This PR reverts #580 and introduces `--keep display` instead.

See discussion in #581.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -176,7 +176,9 @@ depth is changed, regardless of any options set.",
             Arg::new("keep")
                 .help("Strip all metadata except in the comma-separated list")
                 .long("keep")
-                .value_name("list"),
+                .value_name("list")
+                .conflicts_with("strip")
+                .conflicts_with("strip-safe"),
         )
         .arg(
             Arg::new("alpha")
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -611,6 +619,24 @@ fn parse_opts_into_struct(
...
+            .filter_map(|name| {
+                if name == "display" {
+                    keep_display = true;
+                    return None;
+                }
+                Some(parse_chunk_name(name))
+            })
+            .collect::<Result<IndexSet<_>, _>>()?;
+        if keep_display {
+            names.extend(StripChunks::KEEP_SAFE.iter().cloned());
```
AlexTMjugador, line 635: It may be a good idea to standardize the terms used throughout the code and rename this and similar identifiers to `KEEP_DISPLAY`, as it can make things easier to follow. What do you think about this?

andrews05: Ah, I did mean to do that. Must have forgot. Changed now :)

AlexTMjugador (approved):
LGTM, thanks!

PR #589, opened 2024-01-21
Add pre-commit integration docs

UnknownPlatypus:
Hey !

Following the addition of [pre-commit support](https://github.com/shssoichiro/oxipng/issues/429),  I've updated the docs with a small section to reflect that.

I've added this just above the [trunk integration](https://github.com/shssoichiro/oxipng/pull/486) section. 
Because pre-commit is open source while trunk is proprietary, it felt more appropriate to promote this option.

Let me know if you want me to make any changes.

Cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,19 @@ matter of preference. Multiple short options can be combined together, e.g.:
 `-savvo6` is equivalent to to `--strip safe --alpha --verbose --verbose --opt 6`.
 Note that all options are case-sensitive.
 
+## Git integration via [pre-commit]
+
+Add this to your `.pre-commit-config.yaml`
+
+```yaml
+-   repo: https://github.com/shssoichiro/oxipng
+    rev: v9.0.0
+    hooks:
+    -   id: oxipng
+        args: ["-o", "4", "--strip", "safe", "--alpha"]
+```
+[pre-commit]: https://pre-commit.com/
+
 ## Git integration via [Trunk]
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -65,6 +65,19 @@ matter of preference. Multiple short options can be combined together, e.g.:
...
+## Git integration via [pre-commit]
+
+Add this to your `.pre-commit-config.yaml`
+
+```yaml
+-   repo: https://github.com/shssoichiro/oxipng
+    rev: v9.0.0
+    hooks:
+    -   id: oxipng
+        args: ["-o", "4", "--strip", "safe", "--alpha"]
```
AlexTMjugador, line 77: It looks like the pre-commit configuration file must begin with a `repos` map, not a list:

Syntax error

Can you please change it to read like this?

```suggestion
repos:
- repo: https://github.com/shssoichiro/oxipng
  rev: v9.0.0
  hooks:
    - id: oxipng
      args: ["-o", "4", "--strip", "safe", "--alpha"]
```

AlexTMjugador (commented):
Hi! This is a nice addition to the documentation, thank you! :heart:

README.md
UnknownPlatypus: yes exactly!
I took inspiration from some example config from repositories owned by the pre-commit organisation like [this one](https://github.com/pre-commit/mirrors-mypy) where they generally omit this (I think to make the copy/paste to an existing config slightly easier).

But I agree it's more clear like that for new pre-commit users so I made you suggested change.

AlexTMjugador: Ah, true! The nuance here is in the usage of the "add" verb, which now that I think about it, does indeed suggest that those lines should not be interpreted to be a valid config file as-is. Still, I agree with you that it's better to optimize for ease of understanding by people that's not so familiar with the tool already.

AlexTMjugador (approved):
LGTM, let's merge! :rocket:

Issue #129, opened 2018-07-28
resulting image is corrupted


matthiaskrgr:
````
Processing: ./test_images/png-test-corpus/28-c2-to-c3-change-trns-color.png
    256x256 pixels, PNG format
    3x8 bits/pixel, RGB
    IDAT size = 36402 bytes
    File size = 36477 bytes
Reducing image to 8 bits/pixel, 256 colors in palette
Trying: 8 combinations
Found better combination:
    zc = 9  zs = 0  f = 0        28283 bytes
    IDAT size = 28283 bytes (8119 bytes decrease)
    file size = 29120 bytes (7357 bytes = 20.17% decrease)
The resulting image is corrupted and will not be outputted.
This is a bug! Please report it at https://github.com/shssoichiro/oxipng/issues
The resulting image is corrupted
````
https://github.com/FLIF-hub/benchmarks/blob/master/test_images/png-test-corpus/28-c2-to-c3-change-trns-color.png

shssoichiro:
Fixed in 2.1.2

PR #592, opened 2024-02-18
Reduce size of release binary

### LINKED ISSUE #571: Optimize file size of release binaries
Feature Request / Minor –

The w32 executable binary is ~1,385 KB. While this seems small by contemporary standards, it is far larger than equivalent programs. e.g. `optipng` is ~ 209 KB (~15% the size).

This is relevant when working with certain environments / VM's, as it's convenient to package multiple tools in a 'floppy' image <1.44MB.

UPX compression reduces the binary by a third(!) to ~462KB but should be avoided due to false positives with anti-virus software.
Standard zip compression also achieves a similar result ~610KB.  This suggests there is a reasonable amount of redundancy in the binary.

@AlexTMjugador - is it possible to look at compiler optimizations that don't (signficantly) affect speed but produce a more compact executable? Perhaps there is extra fluff that is unneeded for Windows?

andrews05:
This PR makes 3 changes that together reduce binary size by around 25%:
- Sets lto="fat" in cargo.toml
- Sets panic="abort" in cargo.toml
- Sets location-detail=none in RUSTFLAGS

Closes #571

An unrelated change: I've replaced the zopfli test file with a smaller one that runs much faster, as well as removing the slow test for issue-133 which was related to an older alpha optimisation that is no longer relevant.


ace-dent:
@andrews05 - thanks for implementing this! Is the new image for testing permissively licensed?
Comments based on my recollection of #571 ...

1. You decided to avoid `opt-level=s`, which reduced execution  speed ~10%. Now `release` will use the default `3` (it hasn't been explicitly stated). Seems reasonable.
2. In addition to `lto=fat`, I think it is useful to include `codegen-units=1`. This is missing?
3. I believe `build-std` with `panic_immediate_abort` were also found to be [useful](https://github.com/shssoichiro/oxipng/issues/571#issuecomment-1809996427) ?

Many thanks :-)

andrews05:
The image is just a copy of another one in the test suite (see #129 - it comes from a corpus of png test images).

2. `codegen-units=1` drastically slows down compile time for negligible benefit to file size, so it really doesn't seem worth it.
3. `build-std` is a much more advanced technique that I would definitely not recommend for a typical build process, unless there was a very specific need to reduce file size as much as possible. Since we don't have such a need, I'd much rather leave it up to end users to perform this if it's important to them.

ace-dent:
Totally understand your points but it's a shame to not adopt (3), which seemed to save ~30% binary size.

andrews05:
Yeah, I agree. It's a shame that rust can't avoid the bloat more easily.

AlexTMjugador (approved):
I think these changes look good, thanks!

Given that we're already using a nightly toolchain for the release builds, I don't think going with `build-std` is _that_ difficult, but it surely comes with the tradeoff of slower CI times. We can explore that option in more depth in the future; the changes in this PR are not mutually exclusive with that.

PR #593, opened 2024-02-18
Fix secondary palette sort

andrews05:
A very sad bug fix: The battiato sort was only working if the input image was already indexed


ace-dent:
Thanks. Great to see more work on palette optim!

andrews05:
Heh, this is just bug fix, not really "more work" 😄
But it's an important one and could improve many images.

AlexTMjugador (approved):
Good catch, thanks!

Issue #69, opened 2017-06-23
Get oxipng into linux distributions and Homebrew


shssoichiro:
As `cargo install` is not recommended to be used for installing end-user binaries, we should move to get oxipng listed in Linux distributions and Homebrew.

We may not have immediate control over some of these.

- [ ] Homebrew
- [ ] Ubuntu
- [ ] Debian
- [ ] Gentoo
- [ ] Arch
- [ ] Centos

CounterPillow:
Good luck, rust's static compilation means no distro will want to touch this with a 10 foot pole

shssoichiro:
Ripgrep is at least in Homebrew and Gentoo. For Arch there's the AUR. The other distros will probably be a harder sell.

corppneq:
You could use AppImage: https://github.com/AppImage/AppImageKit/

leo-unglaub:
I have build a package for `Alpine Linux` and `OpenBSD`. I will submit them shortly to the official repositories.

magicgoose:
> Good luck, rust's static compilation means no distro will want to touch this with a 10 foot pole

pngquant (later versions) is written in rust as well, and it's present in many distros: https://pkgs.org/download/pngquant

CounterPillow:
@magicgoose that pngquant must not be the one you're referring to, because the one you linked is written in C99.

magicgoose:
@CounterPillow I'm referring to https://github.com/kornelski/pngquant
It is partially written in Rust. (not completely — my mistake)

Calinou:
Just a heads-up for Windows users: oxipng is now available in [Scoop](https://scoop.sh/) ([commit](https://github.com/lukesampson/scoop/commit/664e135d3f1f11e104ae0b53135b0ff98277b5d0)) :smiley:

aaronfranke:
What's the status of this? What is the recommended way to install oxipng?

adiabatic:
It's been in Homebrew for a while.

CounterPillow:
It's also in the official Arch Linux repositories https://archlinux.org/packages/community/x86_64/oxipng/

Since Debian and Ubuntu ship other rust utilities now (e.g. the excellent ripgrep) I imagine it's only a matter of time before a Debian or Ubuntu contributor takes interest and packages it.

remilauzier:
It's now in fedora. https://src.fedoraproject.org/rpms/rust-oxipng

jn64:
[...]

PR #596, opened 2024-03-05
Feature/manpage

### LINKED ISSUE #526: Add man page for the command-line utility

andrews05:
This PR adds a build script to generate a man page using clap_mangen, as per this example:
https://github.com/sondr3/clap-man-example/blob/main/build.rs

I'm not sure what to actually do with the man file from here, I guess it's up to the packaging process to do something with it?
See https://github.com/shssoichiro/oxipng/issues/69#issuecomment-1963352536

Note I couldn't see a way to include the `DISPLAY` chunk names from the constant as we did before. They're now just hardcoded into the help and will require manually updating if the list changes.

Closes #526


andrews05:
Maybe we can use [cargo-deb](https://github.com/kornelski/cargo-deb) to make a deb package in the github workflow.

musicinmybrain:
> I'm not sure what to actually do with the man file from here, I guess it's up to the packaging process to do something with it? See [#69 (comment)](https://github.com/shssoichiro/oxipng/issues/69#issuecomment-1963352536)

Speaking as the current maintainer of the Fedora and EPEL packages, if the build process generates a man page then that’s all I need. I can easily install it.

andrews05:
@musicinmybrain Excellent, that's what I was hoping.

AlexTMjugador:
This is beautiful, great work! For reference, here's an excerpt of the rendered page:

Rendered man page

It was faster for me to push two extra small changes than explaining them on a PR review, but the summary of what I did is as follows:

- I modified the build script to generate man page files to the `generated/assets` folder instead of `target/assets`. The rationale for this change is that we need to create files outside of `OUT_DIR` anyway, and `OUT_DIR`'s relative location to `target` is purposefully underspecified. The previous logic could fail when using the `--target` for a fresh `cargo build`, and maybe if we do [weird shenanigans to change the target directory](https://doc.rust-lang.org/cargo/guide/build-cache.html), so it's best to leverage the fact that [build scripts are executed from a consistent working directory](https://doc.rust-lang.org/cargo/reference/build-scripts.html#inputs-to-the-build-script) and put stuff into a well-known directory that's ignored by Git.
- I devised a way to bring back the generation of `DISPLAY` chunk names from the source code constant, so we don't need to hardcode them anymore :smile:

AlexTMjugador:
> What do you think about changing the workflow to use cargo-deb for the linux builds?

I think that's a good and doable idea. We can begin distributing the generated .deb ourselves too, so even if we don't make it to the official Debian repositories anytime soon, people can still manage their OxiPNG install with the APT package manager.

PR #598, opened 2024-03-26
Optimize color counting

kornelski:
I've seen a panic in `most_popular_edge_color`, so I've made it more resilient.

### DIFF
--- a/src/png/scan_lines.rs
+++ b/src/png/scan_lines.rs
@@ -27,6 +27,8 @@ impl<'a> Iterator for ScanLines<'a> {
         let (len, pass, num_pixels) = self.iter.next()?;
         debug_assert!(self.raw_data.len() >= len);
         debug_assert!(!self.has_filter || len > 1);
+        // The data length should always be correct here but this check assures
+        // the compiler that it doesn't need to account for a potential panic
         if self.raw_data.len() < len {
             return None;
         }
--- a/src/png/scan_lines.rs
+++ b/src/png/scan_lines.rs
@@ -24,20 +24,24 @@ impl<'a> Iterator for ScanLines<'a> {
     type Item = ScanLine<'a>;
     #[inline]
     fn next(&mut self) -> Option<Self::Item> {
-        self.iter.next().map(|(len, pass, num_pixels)| {
-            let (data, rest) = self.raw_data.split_at(len);
-            self.raw_data = rest;
-            let (&filter, data) = if self.has_filter {
-                data.split_first().unwrap()
-            } else {
-                (&0, data)
-            };
-            ScanLine {
-                filter,
-                data,
-                pass,
-                num_pixels,
-            }
+        let (len, pass, num_pixels) = self.iter.next()?;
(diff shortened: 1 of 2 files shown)


kornelski:
I only have a core dump without the image, so I don't know exactly. 

A common tricky case that may be related is pixels having higher values than palette length. 

It also could have been 1x1 interlaced image.

andrews05:
Thanks for this @kornelski. I'll take a closer look later and try to figure out what would actually cause this, but I'm surprised it can happen since:
1. The image should have already been cleaned from any palette issues at this point.
2. Interlaced images are explicitly excluded from the co_occurrence_matrix sort methods.

AlexTMjugador (approved):
Both of your comments sound good! I think this PR could be merged as-is, as it's not exclusive with actually figuring out what's causing the bad `len`, but I'll leave the final choice to @andrews05, given Andrews is more familiar with this code.

andrews05:
I managed to reproduce a crash at that point. @kornelski was right - it can happen when palette entries are missing.
The `reduced_palette` is supposed to fix any such issues, but it won't if the palette was one entry too short. I'll make up a separate PR with a test case to fix this.

src/png/scan_lines.rs
```diff
@@ -24,20 +24,24 @@ impl<'a> Iterator for ScanLines<'a> {
...
-                data,
-                pass,
-                num_pixels,
-            }
+        let (len, pass, num_pixels) = self.iter.next()?;
+        debug_assert!(self.raw_data.len() >= len);
+        debug_assert!(!self.has_filter || len > 1);
+        if self.raw_data.len() < len {
+            return None;
+        }
```
andrews05, line 32: This is the only change I'm unsure about. The raw data length is validated after decompression, so it should be guaranteed to be correct at this point. If it isn't, then something has gone wrong and surely we would rather crash than return None?
https://github.com/shssoichiro/oxipng/blob/f4e631bce773fc5ef67fae0f75a5fa4ea87f2110/src/png/mod.rs#L145-L148

kornelski: I've added an assert above for crashing. 
This condition is to make compiled code more efficient, because panicking branches inhibit optimizations and add code bloat.

andrews05: Ah, good point, thanks for the explanation. I might just add a comment in there for my own benefit 🙂

src/png/scan_lines.rs
```diff
@@ -24,20 +24,24 @@ impl<'a> Iterator for ScanLines<'a> {
...
-            ScanLine {
-                filter,
-                data,
-                pass,
-                num_pixels,
-            }
+        let (len, pass, num_pixels) = self.iter.next()?;
+        debug_assert!(self.raw_data.len() >= len);
+        debug_assert!(!self.has_filter || len > 1);
+        if self.raw_data.len() < len {
```
andrews05, line 30:
```suggestion
        // The data length should always be correct here but this check assures
        // the compiler that it doesn't need to account for a potential panic
        if self.raw_data.len() < len {
```

PR #599, opened 2024-03-28
Fix/truncated palette

andrews05:
Fixes a crash that could occur if a palette had a missing entry. We now guarantee to always fix palette issues, even if it would be larger (this would add at most 3 bytes to the file size).

See #598


AlexTMjugador (approved):
Looks good to me!

PR #600, opened 2024-03-28
Use GitHub macOS ARM runners for building ARM macOS binaries

AlexTMjugador:
GitHub introduced free macOS ARM runners on January, and my experience using them in other projects to improve CI times and be able to actually run tests on Apple Silicon Macs has been positive. Let's use them in OxiPNG to hopefully speed up CI a bit, and finally be able to run the test suite on AArch64 macOS.


andrews05 (approved):
Nice, looks good!

PR #602, opened 2024-04-01
Add modified zeng palette sorting method

andrews05:
This PR adds the modified zeng ("mzeng") palette sorting method, in addition to the existing luma and battiato methods. Speed is very similar to the battiato method with slightly better results on average.

Resulting sizes from two different image sets (all indexed or able to be indexed):
| | master | PR |
|-|-|-|
| Set 1 | 29,647,156 | 29,555,697 |
| Set 2 | 23,732,133 | 23,570,862 |

Additionally, I've added a new "first colour" heuristic for both the mzeng and battiato methods: We use the most popular colour overall, but only if it covers at least 15% of the image. This provided 13k savings on Set 2 vs the edge colour heuristic (which is still used in the luma sort).

### DIFF
--- a/benches/reductions.rs
+++ b/benches/reductions.rs
@@ -301,6 +301,16 @@ fn reductions_palette_sort(b: &mut Bencher) {
 }
 
 #[bench]
+fn reductions_palette_sort_mzeng(b: &mut Bencher) {
+    let input = test::black_box(PathBuf::from(
+        "tests/files/palette_8_should_be_palette_8.png",
+    ));
+    let png = PngData::new(&input, &Options::default()).unwrap();
+
+    b.iter(|| palette::sorted_palette_mzeng(&png.raw));
+}
+
+#[bench]
 fn reductions_palette_sort_battiato(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from(
         "tests/files/palette_8_should_be_palette_8.png",
--- a/src/reduction/mod.rs
+++ b/src/reduction/mod.rs
@@ -1,6 +1,6 @@
 use std::sync::Arc;
 
-use crate::{evaluate::Evaluator, png::PngImage, Deadline, Deflaters, Options};
+use crate::{evaluate::Evaluator, png::PngImage, ColorType, Deadline, Deflaters, Options};
 
 pub mod alpha;
 use crate::alpha::*;
@@ -132,12 +132,41 @@ pub(crate) fn perform_reductions(
         }
     }
 
-    // Attempt to sort the palette using an alternative method
-    if !cheap && opts.palette_reduction && !deadline.passed() {
-        // Make sure we use the `indexed` var if it exists
-        if let Some(reduced) = sorted_palette_battiato(indexed.as_ref().unwrap_or(&png)) {
-            eval.try_image(Arc::new(reduced));
-            evaluation_added = true;
+    // Attempt additional palette sorting techniques
+    if !cheap && opts.palette_reduction {
+        // Collect a list of palettes so we can avoid evaluating the same one twice
+        let mut palettes = Vec::new();
+        if let ColorType::Indexed { palette } = &baseline.ihdr.color_type {
+            palettes.push(palette.clone());
+        }
+        // Make sure we use the `indexed` var as input if it exists
+        // This one doesn't need to be kept in the palette list as the sorters will fail if there's no change
+        let input = indexed.as_ref().unwrap_or(&png);
+
+        // Attempt to sort the palette using the battiato method
+        if !deadline.passed() {
+            if let Some(reduced) = sorted_palette_battiato(input) {
+                if let ColorType::Indexed { palette } = &reduced.ihdr.color_type {
+                    if !palettes.contains(palette) {
+                        palettes.push(palette.clone());
+                        eval.try_image(Arc::new(reduced));
+                        evaluation_added = true;
+                    }
+                }
+            }
+        }
+
+        // Attempt to sort the palette using the mzeng method
+        if !deadline.passed() {
+            if let Some(reduced) = sorted_palette_mzeng(input) {
+                if let ColorType::Indexed { palette } = &reduced.ihdr.color_type {
+                    if !palettes.contains(palette) {
+                        palettes.push(palette.clone());
+                        eval.try_image(Arc::new(reduced));
+                        evaluation_added = true;
+                    }
+                }
+            }
         }
     }
 
--- a/src/reduction/palette.rs
+++ b/src/reduction/palette.rs
@@ -107,14 +107,14 @@ pub fn sorted_palette(png: &PngImage) -> Option<PngImage> {
     enumerated.insert(0, first);
 
     // Extract the new palette and determine if anything changed
-    let (old_map, palette): (Vec<_>, Vec<RGBA8>) = enumerated.into_iter().unzip();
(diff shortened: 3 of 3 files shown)


andrews05:
I actually will tidy this up a bit, so will convert to draft while I do that...

ace-dent:
@andrews05 - nice! I'm working on an [image set](https://github.com/KolibriOS/kolibrios/pull/22), where I have been brute-forcing the palette order. I'd been keen to test this method too!

andrews05:
Brute-forcing? Wow, that's dedication 😅. How are you doing that?

Oxipng does evaluate the existing palette too. You can disable all palette sorting if you need to with `--np`.

Winterhuman:
I have also been brute-forcing a bunch of OxiPNG permutations (including two `--np` variants) to find the best results, and can agree, it does take a **lot** of dedication: https://gist.github.com/Winterhuman/21d7b148db40ff041f397b07a7aafb83

andrews05:
@Winterhuman nice!

I still think it's odd that you're using pngquant in a lossless mode though, which is not what it was built for. You'd be better off swapping it for one or more other lossless optimisers, such as: PNGOUT, AdvPNG, PNGCrush, ECT, OptiPNG, PngOptimizer, pngwolf etc. Piping oxipng output into ECT, for instance, could be quite effective as it has a more advanced deflate compressor.

Existing png multitools include ImageOptim, FileOptimizer, and @ace-dent's pngslim.

andrews05:
Right, I've updated with much cleaner, less C-looking code. I've also made a change to the first colour to improve savings even more (see updated description at top).

andrews05:
Yeah you're right, the most popular colour and most popular edge colour are usually the same, and even when they aren't the choice between the two often doesn't make much difference.

The difference is likely more around situations where the "most popular" isn't actually that popular and putting it first may actually be detrimental. This could happen e.g. in high-detail or photographic material where there are many colours but none that are particularly prominent. These would also likely be image types where the luma sort would not be effective, which is why it may better to retain the edge colour there. The increased diversity by doing different things in different sort methods may also be beneficial.

I haven't tried the 20% threshold with the edge colour yet. I don't anticipate that would be effective, but can certainly try. Of course, I've spent much time already tweaking and experimenting with various things and I have to stop at some point 😄

I don't have the image set with me right now but I can post some samples later.

ace-dent:
@andrews05 I've spent a few hours thinking about this...
My suspicion is the 20% threshold is really a proxy for whether row filter `0` is effective (and Luma sort is as good as any)... or where there is no single dominant 'background' colour, more complex filters (`3` / `4`) are effective... in which case we shouldn't be tinkering with the algorithms' palette order.... perhaps?

Corpus is going to have a big effect here. For small pixel art / icons, nearly always have filter 0. Larger screenshots often favour more complex filters (mix of all).

andrews05:
I've made some more tweaks: Lowered the threshold to 15%, and removed the final rotation of the array in the mzeng method. Saved around 2k total across both sets.

@ace-dent The 20% was just the result of a handful of trials in 10% increments. Nothing super scientific.
Here's the image the saw the biggest improvement:
BeigePyraBack

ace-dent:
@andrews05 - Thanks! Will look over the image with a coffee :-)

> Lowered the threshold to 15% ...

I'm still thinking about this heuristic... I assume we cannot know the row filters, at the stage we are doing the palette sort? (I appreciate they may strong interrelate, so sanely we might pick which to optimise for first).

ace-dent:
Feature request: is it possible with `-v -v` to log some information about the palette sorting?
Ideally in the 'Eval:' line or a summary line? e.g. `Eval: 8-bit Indexed (256 colors - M-Zeng ordered)...` or `...(256 colors - original order)...`, etc.

andrews05:
@ace-dent Correct, we don't know the filters. The filter most commonly ends up being either None or one of the heuristic ones. Rarely is it exclusively one of the other delta filters, so trying to optimise for anything other than 0 is probably not helpful.

And yeah, it would be nice to have more info on those eval lines which are currently identical. I'll see if I can do something about that after this PR.

ace-dent:
Hi @andrews05 - sorry to necropost...
I have done some extensive trials with oxipng's new palette sorting. In most cases it performs very well, compared to a more brute force approach (testing against ~20 sorting variations per image). Typically it performs equally to my brute method, occasionally performs noticeably better, occasionally worse.

I have attached a small corpus of indexed images (GPL licensed), for future algo tuning. I can provide examples of winning and losing paletted images too, if it helps?

Many thanks.
[KolibriOS-indexed.zip](https://github.com/shssoichiro/oxipng/files/15346936/KolibriOS-indexed.zip)

andrews05:
@ace-dent Nice. Yeah, some examples of losing ones could be interesting, along with what algorithm it lost to and by how much.

ace-dent:
@andrews05 - not sure when I will be able to do forensics on the images, so have attached them as is. Will try to review in future.  

[OXIPNG-pal-fail.zip](https://github.com/shssoichiro/oxipng/files/15367081/OXIPNG-pal-fail.zip)

clouvrier:
@andrews05 could you provide set1/2? (if not possible, privately (gmail, user cssignet)). ty!

andrews05:
@clouvrier Oof, this was over a year ago, I'm not sure what sets those actually were now. What do you need them for?

clouvrier:
@andrews05 i do not need those specifically — i am interested in whatever image you have found that benefit from palette sorting

andrews05:
@clouvrier Hm, well pretty much any indexed image can benefit from palette sorting. The order can have a significant impact on the compressibility of the image data, so it's a matter of finding an order that works well.
Try running oxipng with and without the `--np` flag ("don't change the palette") on any indexed image and compare the results. You could try the image I posted further back in this thread, for instance.
Use `-o3` or higher to enable the advanced sorting algorithms and add `-vv` if you want to see the result of each algorithm.

Is this helpful? I'm not sure if I'm fully understanding what you're after.

clouvrier:
so, could you share the files/sets from your first post?

ace-dent:
@andrews05 - some context: Cedric is working on his own palette tests (but little info is shared). Forum post [here](https://encode.su/threads/3879-Optimizing-palette-order-for-indexed-png-images?p=85718&viewfull=1#post85718).

andrews05:
@ace-dent Thanks, context is definitely helpful 😄

So @clouvrier, you're the author of pingo? It sounds like a great tool!
As I said, I'm afraid I don't recall exactly what those sets were. I have an idea one of the them (which included the single image I posted further up) was from a benchmark suite from libjxl or something similar, but I can't seem to find it now. There's nothing special about them though, they were just a bunch of arbitrary indexed images (although it's possible some of them may have been full colour images that I ran through pngquant first). I also often use images from the test suite on your own website, so some of those may have been included too.

Happy to help but not sure what else I can do, sorry. I'd certainly be interested to know if you find any good sorting algorithms though 🙂 (besides GA or brute force approaches).

clouvrier:
@andrews05 sorry i forgot to answer this. about palette: (sort of) luma, nearest, mzeng, and another "greedy-luma" from me which is not as effective but still fast. unlike oxipng which test all combinations, some of them are triggered by results - e.g. if luma would be better, then it could try greedy. same stuff for WebP

i just see your contributions for oxipng - nice work!

andrews05:
Nice, thanks for that @clouvrier.

PR #605, opened 2024-04-06
Create deb archives on deploy

andrews05:
This PR adds a step to build deb archives using cargo-deb in the deploy workflow, along with some setup steps copied from the oxipng.yml workflow.
This deploys two extra release packages: `oxipng_9.0.0-1_amd64.deb` and `oxipng_9.0.0-1_arm64.deb`. (These are just the default output names produced by cargo-deb, we can change it if we need).
These are only done for the gnu targets, not the musl ones. I'm not really sure what's best practice here but I figured gnu was safest.
Disclaimer: I'm not an expert in GitHub workflows. Please review carefully 🙂

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -76,6 +76,39 @@ jobs:
               tar -vczf "${ARCHIVE_NAME}.tar.gz" "$ARCHIVE_NAME"/*;;
           esac
 
+      - name: Set up Ubuntu multiarch
+        if: matrix.target == 'aarch64-unknown-linux-gnu'
+        run: |
+          readonly DISTRO_CODENAME=jammy
+          sudo dpkg --add-architecture arm64
+          sudo sed -i "s/^deb http/deb [arch=$(dpkg-architecture -q DEB_HOST_ARCH)] http/" /etc/apt/sources.list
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
Your work on the new workflow steps looks good! I see you've built upon the existing ones, so I can't complain :laughing: 

I'm totally fine with the package names. Debian follows the same naming scheme in its repositories, so it's all good.

Also, it's customary for Debian packages containing native code to link against glibc, as long as they declare such dependency in their metadata. This is more space-efficient since multiple packages can share the same upgradeable C library. In my experience, `cargo deb` handles generating the correct dependency declarations for glibc just fine.

andrews05:
I've just pushed an extra commit that it seems I omitted earlier. It removes the rust install and just creates the deb archives using the existing binary.

> Also, it's customary for Debian packages containing native code to link against glibc, as long as they declare such dependency in their metadata. This is more space-efficient since multiple packages can share the same upgradeable C library. In my experience, `cargo deb` handles generating the correct dependency declarations for glibc just fine.

Yeah, I found the AArch64 steps were only required in order for it to work out the dependencies. Hopefully arm runners will be publicly available soon and we can simplify the workflows.

PR #607, opened 2024-04-14
Release/9.1

andrews05:
- Update changelog
- Bump version
- Regenerate manual (note the script required a tweak due to an update of `terminal-size`)

Changes in whitespace are due to previous changes made for mangen.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,16 @@
+## Version 9.1.0
+
+- [Improvement] Add `--keep display` equivalent to `--strip safe`.
+- [Improvement] Add modified zeng palette sorting method, improving optimization of indexed images.
+- [Improvement] If only one filter is specified, guarantee to only use this one.
+- [Improvement] Evaluate low-depth indexed even if low-depth grayscale was already achieved.
+- [Bugfix] Fix battiato palette sorting method not being used if the input was not already indexed.
+- [Bugfix] Fix rare crash caused by a truncated palette.
+- [Build] Reduce size of binaries.
+- [Build] Add man page generation.
+- [Build] Publish deb archives for Linux.
+- [Misc] Bump minimum Rust version to 1.74.0.
+
 ## Version 9.0.0
 
 - [Breaking] Remove `--backup` option. Use `--out` or `--dir` to preserve existing files.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -369,7 +369,7 @@ dependencies = [
 
 [[package]]
 name = "oxipng"
-version = "9.0.0"
+version = "9.1.0"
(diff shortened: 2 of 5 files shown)


AlexTMjugador:
Looks good! Before going forward with this I think I should give the deb archive changes a further look, as I have the impression that since you changed the `cargo deb` invocation to use the `--no-build` parameter it may not be necessary to install any cross-compilation toolchains.

andrews05:
Yeah, the better solution is to wait for GitHub arm runners to be available (I believe they're currently in private beta). But I'm sure the current solution will suffice while we wait.

Just let me know when you're ready and I'll publish the release 🙂

ace-dent:
@andrews05 I'm doing extensive testing with your palette sorting work. It is excellent, so well worth a point release! 🎉

AlexTMjugador:
I've had some time to delve into the `cargo deb` dependencies matter, and it seems my suspicions were right. It turns out that we don't actually need the entire cross-compilation toolchain. [`cargo deb` utilizes dpkg-shlibdeps behind the scenes to automatically resolve dependencies](https://github.com/kornelski/cargo-deb/blob/f66a20474975c09bd19e4312db2a1cfc0a56cdfa/src/dependencies.rs#L25-L37), and all `dpkg-shlibdeps` requires are the target shared libraries, which are conveniently provided by the `libc6-arm64-cross` and `libgcc-s1-arm64-cross` packages. These packages are lightweight and don't necessitate a full-blown multiarch setup. I've just pushed a commit with this enhancement, which I've tested on a x64 host within a clean Docker container.

Let's move forward with the release then! 🎉

AlexTMjugador (approved):

andrews05:
@AlexTMjugador Awesome, that looks much tidier, thanks for reworking that!

andrews05:
@AlexTMjugador We've got a problem with the cargo publish.
`Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.`
https://doc.rust-lang.org/cargo/reference/build-scripts.html#outputs-of-the-build-script

I could run it with `--no-verify` for now if we think that's okay but we should definitely fix this going forward.
From my research, there really isn't any good way to handle generated files right now. Some options:
- Revert your [change to the assets dir](https://github.com/shssoichiro/oxipng/pull/596/commits/d7f6757a7845809af14524dafdcddfe9d2929208)
- Use something like xtasks or cargo-make instead of using build.rs
- Make the `manual.sh` script also generate the manpage instead of using build.rs


Additionally, the deploy script didn't publish the release notes. It seems `tail -n +2` need to be `tail -n +3`, though I'm a little confused - was this ever working? 🤔

PR #608, opened 2024-04-16
Document that no metadata is stripped by default

tobbez:


andrews05:
Hi @tobbez, thanks for the PR.

This sounds good, except `none` isn't actually a real option for `--strip`. If you pass `--strip none` it will actually attempt to strip chunks of type `none`.

If you think it would be helpful, perhaps this a feature you would like to add?

tobbez:
I don't think a dedicated option to disable stripping is needed (unless a preset that turns stripping on is expected to be added in the future).

That doesn't seem particularly likely, and in that case it would be better to just update the documentation instead, as you suggested (just pushed such a change).

andrews05 (approved):
Looks good, thank you!

And you're correct, there's no plans to ever do stripping by default.

PR #612, opened 2024-04-22
Move the location for generated man pages inside target dir

### LINKED ISSUE #611: 9.1.0 Release missing on crates.io
You've created a 9.1.0 release here on GitHub, but I cannot find that release on crates.io yet. Did you maybe miss publishing it?

shssoichiro:
The cargo release process expects that the `build.rs` file will not modify anything outside of the `target` directory. Currently, we are doing this for the man pages, which results in the following error when running `cargo publish`:

```
  Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR.
  Added: /home/soichiro/repos/oxipng/target/package/oxipng-9.1.0/generated
        /home/soichiro/repos/oxipng/target/package/oxipng-9.1.0/generated/assets
        /home/soichiro/repos/oxipng/target/package/oxipng-9.1.0/generated/assets/oxipng.1
```

This change moves the `generated` directory to within `target` in order to resolve this.

Fixes #611 (will need to tag a 9.1.1 and `cargo publish` after)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -104,7 +104,7 @@ panic = "abort"
 [package.metadata.deb]
 assets = [
     ["target/release/oxipng", "usr/bin/", "755"],
-    ["generated/assets/oxipng.1", "usr/share/man/man1/", "644"],
+    ["target/generated/assets/oxipng.1", "usr/share/man/man1/", "644"],
     ["README.md", "usr/share/doc/oxipng/", "644"],
     ["CHANGELOG.md", "usr/share/doc/oxipng/", "644"],
 ]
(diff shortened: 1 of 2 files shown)


ace-dent:
Issue was noted by @andrews05 [here](https://github.com/shssoichiro/oxipng/pull/607#issuecomment-2068215565).

andrews05:
Thanks @shssoichiro, nice to see you're still active here 🙂 
Very unfortunate to have this issue, apologies for the mess we've made.

I think we need to make the assets relative to OUT_DIR though, as in examples such as https://github.com/kornelski/cargo-deb/blob/main/example/build.rs. The cargo-deb read me notes: `You can use target/release/ in asset paths, even if Cargo is configured to cross-compile or use custom CARGO_TARGET_DIR. The target dir paths will be automatically corrected.` So it should be pretty safe to do it this way.

I've been working on making this change, I'm currently just making sure we still get the man page on deploy...

PR #620, opened 2024-05-18
Remove default value for interlace to ensure it is kept when nx is used

### LINKED ISSUE #619: Unexpected --nx --nz behavior. Ignores interlaced images.
Environment: 
Windows 10 22H2

Flags `--nx --nz` should disable all optimizations for pngs right? It works on all images with one weird exception, interlaced pngs causes oxipng to ignore these flags and optimizes the image anyways. 

It seems that the default of `-i 0` automatically overrides everything else. It turns out you have to pass `--nx --nz -i keep` to truly disable all optimizations. The behavior should be so that `--nx --nz` alone keeps the images untouched regardless of them being interlaced or not.

Screenshot

andrews05:
Fixes #619.

Interlacing still defaults to 0 when `--nx` is not specified and can still be explicitly set even if it is (e.g. `-i 0 --nx` will still deinterlace).

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -171,8 +171,9 @@ transformation and may be unsuitable for some applications.")
                 .action(ArgAction::SetTrue),
         )
         .arg(
+            // Note: The default value is not explicitly set here, as it is dependant on the `--nx` flag.
             Arg::new("interlace")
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
LGTM! Apologies for the late review :sweat_smile:

PR #622, opened 2024-06-02
Use MatchOptions::default() to disable case-sensitivity

### LINKED ISSUE #621: Case sensitive names under Windows
oxipng-9.1.1-x86_64-pc-windows-msvc.zip, W10

"oxipng *.png" optimizes *.png and ignores *.PNG

andrews05:
Fixes #621

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -157,7 +157,8 @@ fn collect_files(
 fn apply_glob_pattern(path: PathBuf) -> Vec<PathBuf> {
     let matches = path
         .to_str()
-        .and_then(|pattern| glob::glob(pattern).ok())
+        // Use MatchOptions::default() to disable case-sensitivity
+        .and_then(|pattern| glob::glob_with(pattern, glob::MatchOptions::default()).ok())
         .map(|paths| paths.flatten().collect::<Vec<_>>());
 
     match matches {


AlexTMjugador:
I find this change reasonable for Windows environments. [The Win32 API goes to great lengths to present a case-insensitive view of the file system](https://web.archive.org/web/20130723020137/http://support.microsoft.com/kb/100625), even if the filesystem itself is case-sensitive. Therefore, many users reasonably expect this behavior.

However, in Linux environments, it's customary to always treat file names in a case-sensitive manner. For instance, when I run `stat *.png` on my `zsh` terminal at a Linux box, it ignores files with a `.PNG` extension. So this change would conflict with these expectations.

Most other Unix-like, POSIX-compliant systems should behave similarly to Linux in this regard. However, I’ve read that macOS is an exception, being case-insensitive by default. (I'm not a macOS user, so please correct me if I'm wrong!)

What do you think about modifying this PR to use `#[cfg]` attributes for different glob case-sensitivity based on the target OS? I think this approach would accommodate the different expectations of users on various platforms pretty neatly.

andrews05:
@AlexTMjugador You’re correct, but glob here is only used on Windows so it doesn’t affect other platforms.

AlexTMjugador:
Ah, right, my bad, the function this PR modifies has a `#[cfg(windows)]` just above it that I completely overlooked. Let's merge this then!

PR #623, opened 2024-06-14
add oxiora script for handling OpenRaster files

sbrl:
This script is probably inappropriate for submitting as a pull request to this repository, but I couldn't find a wiki to submit it to instead.

I discovered on the [OpenRaster website](https://openraster.org/) that an OpenRaster image is just a `.zip` containing a bunch of PNGs (and, potentially, SVGs), so I wrote a little script to optimise them. I was surprised at the results, as both [DrawPile](https://drawpile.net/) and [Krita](https://krita.org/en/) are not very good at writing optimal OpenRaster files.

Hopefully this is useful to someone?

### DIFF
--- a/scripts/oxiora.sh
+++ b/scripts/oxiora.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+
+set -e;
+
+###
+## Input handling
+###
+
+# Set flags for calling oxipng later
+if [[ -n "${UNSAFE}" ]]; then FLAGS="${FLAGS} -a"; fi
+
+if [[ -z "${1}" ]]; then
+	echo "Wrapper for oxipng for handling OpenRaster files
+
+Usage:
+    oxiora file_a.ora file_b.ora ... file_n.ora
+
+Environment variables:
+    UNSAFE
+        Set to any value to send the -a flag to oxipng, which allows altering the colour of completely transparent pixels. Alters the content of images, so disabled by default.
+    
+    FLAGS
+        Any additional flags to pass to oxipng.
+";
+	exit 0;
+fi
+
+###
+## Preamble
+###
+
+# Create temporary directory and clean it up afterwards no matter what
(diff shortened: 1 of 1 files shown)


AlexTMjugador:
This is actually an interesting script. I find it a bit "alien" to the actual code of this repository, as I don't think it scales well to include scripts and/or code for every format or container under the sun that happens to use PNG internally, but I can see how it can come in handy and solve optimization needs for more people.

What do you and others think about adding a section to the README with useful script snippets like these? The code could be hosted in a Gist or something similar.

ace-dent:
-1
@AlexTMjugador - while a nice idea, I think it might prove problematic to link to *many* other projects and adds ongoing maintenance... I'd say the suggestion is out of scope. Sorry 🤗

andrews05:
I think it would be more appropriate to post this to DrawPile/Krita or OpenRaster itself, as it seems more relevant to OpenRaster users than general oxipng users.

andrews05:
Also just for reference, [FileOptimizer](https://nikkhokkho.sourceforge.io/?page=FileOptimizer) uses oxipng and can recursively optimise zip files, so it likely works for OpenRaster files.
[Leanify](https://github.com/JayXon/Leanify) can too, though it uses zopflipng instead.

sbrl:
Ah, I thought that it was unlikely this would be accepted. No worries!

@AlexTMjugador: Yeah, I thought as much. It would be cool to have somewhere that lists associated scripts and programs, but I would probably agree that place isn't in this repository.

@andrews05: Good idea, I hadn't thought of that. While I can't find any source code repository for the OpenRaster website, I think I've seen some forums for Krita previously that I can dig up.

@andrews05: FileOptimiser looks cool! Though it does appear to be windows-only.

Thanks for everyone's time!

AlexTMjugador:
Everyone's opinions here are fair enough. Thanks for the interesting PR anyway!

PR #630, opened 2024-07-01
Fix name capitalization

karlhorky:
Hi, thanks for Oxipng!

I saw further down in the readme that Oxipng was being used in the middle of a sentence with a capital O, so I adjusted the other lowercase usages (other than the usages in backticks, which may instead refer to the command line program itself) from "oxipng" to also use "Oxipng"

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,14 +24,14 @@ Oxipng is known to be packaged for the environments listed below.
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/oxipng.svg?exclude_unsupported=1&columns=3&exclude_sources=modules,site)](https://repology.org/project/oxipng/versions)
 
-Alternatively, oxipng can be installed from Cargo, via the following command:
+Alternatively, Oxipng can be installed from Cargo, via the following command:
 
 ```
 cargo install oxipng
 ```
 
 Oxipng can also be built from source using the latest stable or nightly Rust.
-This is primarily useful for developing on oxipng.
+This is primarily useful for developing on Oxipng.
 
 ```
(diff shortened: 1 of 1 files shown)


ace-dent:
Need to check in with @shssoichiro what the naming intention is! :-)

AlexTMjugador:
Oddly enough, I tend to capitalize OxiPNG like that, not as Oxipng or oxipng... :joy:

andrews05:
@AlexTMjugador Personally, I think that makes it look too similar to OptiPNG 😆

ace-dent:
Not sure if it's open to voting(!?)... but my preference is `oxipng` and capitalised at the start of a sentence.

karlhorky:
> `oxipng` and capitalised at the start of a sentence.

Usually mixing like this will lead to style confusion - often it's better to have a single, strong capitalization style for a brand name and stick with it (regardless of where the word is used in the sentence).

However, that's not to say that the brand name for the project / ecosystem can't be styled differently than the brand name for the command line interface, for example:

1. Oxipng (capital O, everything else lowercase) for the brand name for the project / ecosystem / anything else
2. `oxipng` for the brand name for the command line interface program

AlexTMjugador:
I'm still unsure about the preferred capitalization for the project, but I think we can all agree that maintaining consistency within a single document makes it look tidier. Therefore, I'm merging this PR as-is. At the moment, I do not intend this PR to imply that a given capitalization style should be preferred over others, though.

karlhorky:
@AlexTMjugador thanks for the consideration, review and merge!

PR #632, opened 2024-07-09
dep:features

kornelski:
Fixes ambiguity between optional deps and crate features

https://lib.rs/crates/oxipng/features#features-dep-implied

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,11 +83,13 @@ clap_mangen = "0.2.20"
 rustc_version = "0.4.0"
 
 [features]
-binary = ["clap", "glob", "env_logger"]
-default = ["binary", "filetime", "parallel", "zopfli"]
-parallel = ["rayon", "indexmap/rayon", "crossbeam-channel"]
+binary = ["dep:clap", "dep:glob", "dep:env_logger"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -83,11 +83,13 @@ clap_mangen = "0.2.20"
...
 [features]
-binary = ["clap", "glob", "env_logger"]
-default = ["binary", "filetime", "parallel", "zopfli"]
-parallel = ["rayon", "indexmap/rayon", "crossbeam-channel"]
+binary = ["dep:clap", "dep:glob", "dep:env_logger"]
+default = ["binary", "parallel", "zopfli", "filetime"]
+parallel = ["dep:rayon", "indexmap/rayon", "dep:crossbeam-channel"]
 freestanding = ["libdeflater/freestanding"]
-sanity-checks = ["image"]
+sanity-checks = ["dep:image", "dep:rayon"]
```
andrews05, line 90: This shouldn't need rayon. It looks like we're just missing a use line, if you can remove this here I'll make a separate PR to fix it.

andrews05 (changes requested):
Thanks @kornelski, this looks like a helpful change.

PR #634, opened 2024-07-09
Allow sanity-checks without parallel

andrews05:
This adds the use of the rayon shim when parallel is not enabled. See #632.


AlexTMjugador (approved):
Good catch! This works fine, so merging :tada:

PR #635, opened 2024-07-10
Make CI Clippy static analysis checks more robust

AlexTMjugador:
I have identified two potential improvements for how we perform static analysis on our code in our CI pipeline:

- The `giraffate/clippy-action` we currently use has not been updated to Node 20, and GitHub has repeatedly indicated that they will phase out actions that do not support the latest Node versions. [Despite my efforts to help with the update by submitting a pull request upstream](https://github.com/giraffate/clippy-action/pull/87), it has been ignored for months despite its perceived ease of review, raising concerns about the ongoing maintenance of the action. This situation suggests we should explore alternative methods for integrating Clippy with GitHub's UI.
- As evidenced by [PR 632](https://github.com/shssoichiro/oxipng/pull/632), thoroughly testing Rust crates for every possible feature combination is often overlooked due to the tedious nature of the task. Our current CI setup only checks two feature combinations, which is far from comprehensive.

To address the first improvement, these changes drop `clippy-action` entirely in favor of utilizing [GitHub's native CodeQL SARIF (Static Analysis Results Interchange Format) file integration](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github). Since Clippy cannot directly output lints in SARIF, [`clippy-sarif`](https://crates.io/crates/clippy-sarif) is used to convert Clippy's JSON output to SARIF. Additionally, `sarif-fmt` is added to turn SARIF into a human-friendly display format in the workflow run logs.

For the second improvement, let's use [`cargo hack` with the `--feature-powerset` flag](https://github.com/taiki-e/cargo-hack?tab=readme-ov-file#--feature-powerset) to run Clippy for every possible feature combination. This approach strikes a good balance between CI runtime and thoroughness, as the number of feature combinations grows superlinearly with the number of features: running `cargo nextest` for every powerset element would lead to excessively long CI times.

While at it, I have fixed the Clippy lints that were catched by the more exhaustive checks.

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -156,6 +156,7 @@ jobs:
       - name: Upload analysis results to GitHub
         uses: github/codeql-action/upload-sarif@v3
         if: always() && matrix.target == 'x86_64-unknown-linux-gnu'
+        continue-on-error: true
         with:
           sarif_file: clippy-results.sarif
           category: clippy
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -102,25 +102,33 @@ jobs:
       - name: Install nextest
         uses: taiki-e/install-action@nextest
 
+      - name: Install cargo-hack
+        if: matrix.target == 'x86_64-unknown-linux-gnu'
+        uses: taiki-e/install-action@cargo-hack
+
+      - name: Install clippy-sarif
+        if: matrix.target == 'x86_64-unknown-linux-gnu'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: clippy-sarif
+
+      - name: Install sarif-fmt
+        if: matrix.target == 'x86_64-unknown-linux-gnu'
+        uses: taiki-e/install-action@v2
+        with:
+          tool: sarif-fmt
+
       - name: Run rustfmt
         if: matrix.target == 'x86_64-unknown-linux-gnu'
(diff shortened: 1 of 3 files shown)


AlexTMjugador:
The bot comment above is expected due to the usage of GitHub's CodeQL SARIF integration. Please heed it accordingly before this PR is merged.

andrews05 (approved):
This looks great @AlexTMjugador! Running all feature combos should be a big help.

I was also thinking it might be good to use the GitHub arm runners for the arm linux builds now that they're generally available, is that something you would want to do?

AlexTMjugador:
> [edit] Actually I may be wrong about this, it might only be available for teams and enterprises...

Indeed, [that's sadly the case at the moment](https://github.blog/changelog/2024-06-03-actions-arm-based-linux-and-windows-runners-are-now-in-public-beta/). However, in the blog post I link to GitHub said that it expects to "begin offering Arm runners for open source and personal accounts by the end of the year", so if all goes according to plan we can get rid of QEMU soon enough :smile:

andrews05:
Yeah, sounds good! Let's go ahead and merge this.
We've had a few bug fixes, do you think we should do a 9.1.2 release?

AlexTMjugador:
> We've had a few bug fixes, do you think we should do a 9.1.2 release?

Yeah, I deem the current mainline stable enough and those bugfixes may indeed come in handy. Let's do it!

PR #642, opened 2024-09-26
Improved Docker image + publish on `ghcr.io`

AFCMS:
Fix #610

- [x] Native cross-compilation support in Dockerfile. Only `linux/amd64` and `linux/arm64` supported and tested (Tier 1 platform support from Rust)
- [x] Cache support for cargo downloads and compilation results in Dockerfile
- [x] Open Container's [annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md) in Dockerfile 
- [x] GitHub Actions workflow to build the image for both platforms, publishing to ghcr.io on tags and master branch pushes.
- [x] Disable use of GitHub Actions cache for tags build, allow manually triggering the workflow with or without cache.
- [x] [Attestation artifacts](https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds) for builds
- [x] Add basic informations about the use of the Docker image in README

I also changed the WORKDIR from `/src` to `/work`, because if got me confused with the use of the src folder for the project source in the first stage of the Dockerfile. It doesn't impact anything anyways since you can mount the file where you want and have the program options point to it.

**How to test?**
_After checking out the branch and making sure you have QEMU installed to build/test_
```shell
# Docker sadly doesn't have yet the way to have locally a tag with multiple platforms
# despite being able to pass multiple platforms to the build command, so we have to use two distinct tags.
docker build --platform=linux/amd64 --tag test-oxipng-amd:latest --load .
docker build --platform=linux/arm64 --tag test-oxipng-arm:latest --load .

docker run --rm test-oxipng-amd:latest --version
docker run --rm test-oxipng-arm:latest --version

# Run on some files
docker run --rm -it -v $(pwd):/work test-oxipng-amd:latest -a /work/tests/files/apng_file.png
docker run --rm -it -v $(pwd):/work test-oxipng-arm:latest -a /work/tests/files/apng_file.png

# Remove the images
docker image rm test-oxipng-amd:latest
docker image rm test-oxipng-arm:latest
```

For the workflow, see the GitHub Actions logs. If you want to test the ghcr.io publishing you can merge the branch into a fork and see the result.

PR #645, opened 2024-11-19
Move manpages generation to an xtask

AlexTMjugador:
PR #596 brought forward automatic generation of Linux manual pages for Oxipng, which is executed every time Oxipng is built. However, while building manpages on every build is convenient for Oxipng development and doing so didn't catch my attention initially, it introduces noticeable inefficiencies for crates using Oxipng as a library: during their build, Oxipng manpages are also built, even though most dependent crates won't use such artifacts, as they are not considered part of the public Oxipng crate API or even appropriate for non-human consumption.

Moreover, generating manpages depends on `clap`, which is a heavyweight dependency: according to a fresh `cargo build --timings --release` on my development workstation, its `clap_builder` dependency is the third most time consuming unit to build, totalling 1.5 s (out of 11.7 s, or 12.8%). And there is no way for dependent crates to turn this off: [`build-dependencies` cannot be conditional on crate features](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#platform-specific-dependencies). Potentially using other `cfg` hacks to either enable or disable manpage generation is unergonomic, if not outright disallowed. Besides reducing their compilation time cost, dependent crates may also want to trim the size of their dependency tree, avoiding unnecessary dependency downloads in the process.

Therefore, a better solution to conditionally build manpages in a way convenient for both Oxipng maintainers and downstream consumers is needed. My proposal implemented in this PR is to leverage the [`cargo-xtask`](https://github.com/matklad/cargo-xtask) convention to define an auxiliary crate to move the manpage generation logic and dependencies to, which is not part of the `oxipng` crate published on `crates.io`. That way Oxipng maintainers and packagers can still generate manpages at request with ease, without any automation being noticeable to uninterested crate consumers. And as a side benefit, Oxipng maintainers can also benefit from slightly faster iteration times due to the lack of a build script for the main crate.

The new `mangen` xtask can be run at any time with `cargo xtask mangen`. The generated manpages are now available at `target/xtask/mangen/manpages`. Existing deployment scripts were updated accordingly.


Cargo.toml
```diff
@@ -105,7 +102,7 @@ panic = "abort"
 [package.metadata.deb]
 assets = [
     ["target/release/oxipng", "usr/bin/", "755"],
-    ["target/release/assets/oxipng.1", "usr/share/man/man1/", "644"],
+    ["target/release/manpages/oxipng.1", "usr/share/man/man1/", "644"],
```
andrews05, line 105: If I'm understanding everything correctly, this will only work within the deploy workflow, right? If you just run `cargo deb` normally the path will surely not be correct. Will it work if you put the xtask path here and skip the move in the workflow?

andrews05 (commented):
Thanks @AlexTMjugador, great to see this done! I think xtask was always the better solution, I just didn't have any experience with it.

Cargo.toml
AlexTMjugador: You're right! I think it's a good idea to skip the move in the workflow, that way there are less moving parts in this automation :+1:

andrews05 (approved):
Looks good!
Shall we do another release?

AlexTMjugador:
Yes, I think we can proceed with a new patch or minor release. However, I'd like to get #642 merged before doing the release, as it's likely to be a pretty welcome improvement for those using Docker.

XhmikosR:
A minor issue that I'm not sure it's worth opening a new one: the MANUAL.txt file contains trailing spaces when there's more than one paragraph in the help text. Could be tackled in the shell script or in your Rust code :)

andrews05:
@XhmikosR Are the trailing spaces causing a problem?

XhmikosR:
Not a big deal, it's just it contradicts with https://github.com/oxipng/oxipng/blob/master/.editorconfig. Again, minor issue, I just thought I'd mention it :)

andrews05:
Right, thanks for pointing it out but I think I’ll just leave it.

PR #646, opened 2024-11-25
Add list of software using oxipng

andrews05:
This adds a short list of software using Oxipng to the read me. It's not intended to be an exhaustive list, just some ones that I'm aware of that may have broad general interest.

I've also removed the long section about Trunk which I'm not sure is appropriate to include in this read me. Trunk users can find documentation on this from Trunk directly.

Lastly, I've removed the older benchmark which I'm sure has no relevance any more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,46 +93,6 @@ repos:
 ```
 [pre-commit]: https://pre-commit.com/
 
-## Git integration via [Trunk]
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
Looks good to me, thank you!

PR #648, opened 2024-11-26
Break when best possible filter result found

andrews05:
I had this sudden idea for a tiny optimisation: For heuristic filter strategies, skip checking remaining filters if we find we have the best possible result. We can do this for MinSum, Entropy, Bigrams, and BigEnt, but not Brute.
For most images this will have little-to-no impact, but for some such as the test file "palette_1_should_be_palette_1.png" (where most lines are all zeros) we can see a 20% performance improvement at o2 and 10% at o5.

### DIFF
--- a/src/png/mod.rs
+++ b/src/png/mod.rs
@@ -345,9 +345,10 @@ impl PngImage {
         for line in self.scan_lines(false) {
             if prev_pass != line.pass || line.data.len() != prev_line.len() {
                 prev_line = vec![0; line.data.len()];
-                // Calculate the best possible entropy for this pass
+                // Calculate the best possible result for this pass
                 best_possible = match filter {
                     RowFilter::Entropy => ilog2i(line.data.len() as u32 + 1) as i32,
+                    RowFilter::Bigrams => 1,
                     RowFilter::BigEnt => ilog2i(line.data.len() as u32) as i32,
                     _ => 0,
                 }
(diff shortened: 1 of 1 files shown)


src/png/mod.rs
```diff
@@ -383,6 +390,10 @@ impl PngImage {
                                 best_size = size;
                                 std::mem::swap(&mut best_line, &mut f_buf);
                                 best_line_raw.clone_from(&line_data);
+                                if size == 0 {
+                                    // Best possible result
+                                    break;
+                                }
```
AlexTMjugador, line 396: For any filter other than `Entropy` or `BigEnt`, could we initialize the `best_possible` variable to match the corresponding right-hand value in these comparisons? I think that approach would be slightly cleaner, as it would make every "best possible result" check the same and centralize the definition of the best possible entropy in one place.

andrews05: We could do that, it would just require more type casts - Entropy and BigEnt use i32 while MinSum and Bigrams use usize.

AlexTMjugador: Given that `as _` can be used to concisely let the Rust compiler figure out the target data type here, I think I still prefer my suggestion.

AlexTMjugador (approved):
Thanks for the PR, it looks great now! Feel free to merge after rebasing and once CI passes :crossed_fingers:

src/png/mod.rs
AlexTMjugador: That also works! :+1:

ace-dent:
Sorry- if this seems obvious- but how is the _best_ possible result determined?

AlexTMjugador:
> Sorry- if this seems obvious- but how is the _best_ possible result determined?

I think it's determined from the [Shannon's source coding theorem](https://en.wikipedia.org/wiki/Shannon%27s_source_coding_theorem), assuming that the [entropy](http://bearcave.com/misl/misl_tech/wavelets/compression/shannon.html) is as ideal as it can get, which is the case for rows with a single constant value. See also [this random Reddit topic with less mathy discussion](https://www.reddit.com/r/askscience/comments/90xtaw/is_there_a_theoretical_minimum_number_of_bits/) about it.

TPS:
I love the concept of this — thanks, @andrews05 & @AlexTMjugador! — but is there way to determine & disable automatically when this _shouldn't_ be used?  (E.g., perhaps when random/static or photographic images?)  Or is the overhead _truly_ so negligible that's not relevant?

andrews05:
For each delta filter, the heuristic strategies apply the filter to the current line and then run an algorithm to produce a single value. They then choose the filter based on which one produced the "best" value.

MinSum chooses the one with the *smallest* value and for this algorithm the smallest *possible* value for any line is 0.
Bigrams is similar but the smallest possible value is 1.
Entropy and BigEnt choose the *largest* value and the maximum possible value can be calculated from the length of the line.

To be clear, the best possible value is always deterministic and it's not possible for this change to have any impact on the output file size. It's purely a performance gain that will occur if all bytes in the line are zero. (In retrospect, I could have just checked for this up front and picked the None filter automatically. I might play around with that...)

PR #650, opened 2024-11-27
Assume None if the line is all zeros

andrews05:
#648 may have been a bit hasty - I realised afterward that there's a simpler way to achieve the same thing, and include the Brute filter as well.

This reverts #648 and instead just picks None up front if the line is all zeros. This is guaranteed to be the chosen filter for MinSum, Entropy, Bigrams and BigEnt. It's almost certainly true for Brute as well but this is harder to prove. I've tested this across hundreds of images and found no change in output.

### DIFF
--- a/src/png/mod.rs
+++ b/src/png/mod.rs
@@ -360,6 +360,15 @@ impl PngImage {
                 prev_line = line_data;
             } else {
                 // Heuristic filter selection strategies
+
+                if line_data.iter().all(|&x| x == 0) {
+                    // Assume None if the line is all zeros
+                    filtered.push(RowFilter::None as u8);
+                    filtered.extend_from_slice(&line_data);
+                    prev_line = line_data;
+                    continue;
+                }
+
                 let mut best_line = Vec::new();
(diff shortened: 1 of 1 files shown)


AlexTMjugador:
Upon reviewing this again, I wonder why is it that `None` always ends up being the chosen filter? In other words, can't heuristic filter trials achieve what was previously considered the `best_possible` size using other filters? Is this behavior specific to lines composed entirely of zeros? If so, could this concept be extended to other patterns, such as all ones?

andrews05:
For Entropy, Bigrams, and BigEnt, it technically is possible for other filters to achieve `best_possible` but I think it would be such a rare occurrence that it's probably not worth trying to optimise for these cases.
These three heuristics can all achieve `best_possible` if the output line is uniform, including the filter byte. E.g. for Sub to achieve `best_possible` the bytes *after filtering* must consist of all ones.

AlexTMjugador (approved):
I see, sounds good. Thanks for the explanation!

PR #652, opened 2024-11-29
API docs

kornelski:


src/interlace.rs
```diff
@@ -4,10 +4,13 @@ use bitvec::prelude::*;
 
 use crate::{headers::IhdrData, png::PngImage, PngError};
 
+/// Whether to enable progressive rendering. See [`Options`][crate::Options])
 #[repr(u8)]
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub enum Interlacing {
+    /// Makes images load top to bottom.
     None,
+    /// Makes it possible to render partially-loaded images at lower resolution. Usually incrases file sizes.
```
andrews05, line 13: Typo
```suggestion
    /// Makes it possible to render partially-loaded images at lower resolution. Usually increases file sizes.
```

src/options.rs
```diff
@@ -133,7 +136,8 @@ pub struct Options {
     ///
     /// Default: `None`
     pub strip: StripChunks,
-    /// Which DEFLATE algorithm to use
+    /// Which DEFLATE (gzip) algorithm to use
```
andrews05, line 139: We should probably say zlib rather than gzip

andrews05 (changes requested):

andrews05 (approved):

andrews05:
Awesome, thanks @kornelski!

PR #653, opened 2024-11-30
Fix manpage for `--zi`

andrews05:
I don't know if this is a bug in the mangen but it seems to require explicitly setting the value name to display correctly.
Before:
```
       --zi [default: 15]
              Set the number of iterations to use for Zopfli compression.
```

After:
```
       --zi=iterations [default: 15]
              Set the number of iterations to use for Zopfli compression.
```

This does not affect MANUAL.txt or the built-in help.


AlexTMjugador:
Interesting quirk indeed, thanks!

PR #656, opened 2024-12-01
Cleanup tests

### LINKED ISSUE #655: Some test files look dubious from a copyright/license perspective
While considering using the GitHub source archive as an additional source for Fedora’s `rust-oxipng` package due to https://github.com/shssoichiro/oxipng/issues/654, I noticed that several test files, such as [`issue-141.png`](https://github.com/shssoichiro/oxipng/blob/master/tests/files/issue-141.png), look like they may be based on copyrighted material not covered by oxipng’s MIT license.

In addition to any concerns that this might raise for oxipng redistributing these files on GitHub, this is a significant inconvenience for anyone – like a Linux distribution – who might have a reason to distribute source archives exported from GitHub, since these test files would need to be filtered out first.

Please consider auditing the contents of `tests/files` for license status, and attempting to replace those that may have license issues with “synthetic” test files that demonstrate the same issues.

andrews05:
I've replaced a few tests with synthetic images and removed a number that I don't believe are relevant anymore.

I've also removed the #553 test without replacement as I don't know how to synthesise a replacement and the issue is still open anyway.

Going forward I think we should avoid adding any new "issue" tests and instead synthesise something suitable in one of the other collections.

Closes #655.


AlexTMjugador (approved):
I'm curious about how AI models like DALL·E, Stable Diffusion, or even DeepDream would perform in creating technically interesting test images... :joy:

That aside, the changes look great, thanks! It's a bit sad to let some of the older test images go, but I agree they're probably no longer relevant for catching regressions. For any new testsuite images that aren't synthetic, we can always ask issue reporters to confirm their licensing from now on, too.

PR #659, opened 2024-12-03
Add feature flag to use system libdeflate

shssoichiro:
This need was raised in https://bugs.gentoo.org/944285. Can be enabled with `--features system-libdeflate` to use the libdeflate located on the user's system instead of building our own.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -331,6 +331,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2f4ae7b48098016dc3bc64a35605668f0af4425ec1a4a175ce2d0c1129067932"
 dependencies = [
  "cc",
+ "pkg-config",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


AlexTMjugador (approved):
Looks good to me, allowing users to choose which libraries Oxipng links to is a welcome concept. Thanks!

As a side note, the underlying Gentoo packaging issue shouldn't occur with Oxipng 9.1.3. This version updates `libdeflater` to 1.22, and according to the [related PR](https://github.com/gentoo/gentoo/pull/39539), the problem stemmed from the older `libdeflate` version.

AlexTMjugador:
cc @lo48576 in case the information above about internal `libdeflater` updates on Oxipng 9.1.3 is useful in some way for Gentoo packaging.

lo48576:
Thank you for introducing the flag!

I'm aware of updated `libdeflater` dependency in the latest version, but I'll use a patch for v9.1.3 and `system-libdeflate` flag for the later versions (because it is (maybe) more preferred way in Gentoo Linux).

The Gentoo repo has `libdeflate` package (app-arch/libdeflate) but currently the last version added is 1.21, and oxipng 9.1.3 depends on libdeflater 1.22.0, so I'll wait for =app-arch/deflate-1.22 to be added to the Gentoo.

PR #661, opened 2024-12-15
docs: added the defaults for some `Options`

CommanderStorm:
There are some parts of the `Options`-struct that miss the defaults.
I also made sure that `Interlacing` is a link


AlexTMjugador (approved):
Thank you for the PR! It looks good to me.

PR #663, opened 2024-12-28
Remove obsolete flag warning

kornelski:


andrews05 (changes requested):
We'll need to remove the option from `cli.rs` too.

andrews05 (approved):

PR #668, opened 2025-01-14
Improve APNG support

andrews05:
Two improvements to APNG support:
1. fcTL and fdAT chunks are parsed into Frames, allowing us to consolidate multiple fdATs for a frame and re-sequence the chunks. This means we can now recompress all frames, rather than only in the case where there's a single fdAT for the frame.
2. Following on from this, we can now also re-filter the frames, including alpha optimisation when enabled. We use the same filter that was selected for the IDAT.

Reductions are still disabled.

Results on a set of 32 apngs, optimised with `-sao6`:

source | size
--|--
original | 20,022,193 
master | 19,412,724 (3.0% smaller)
with full recompression only | 16,494,618 (17.6% smaller)
with refiltering | 15,583,118 (22.2% smaller)
apngopt* (default) | 13,520,262 (32.5% smaller)
apngopt + oxipng | 13,341,852 (33.4% smaller)

(There was a particularly large and poorly compressed file in the set which may be skewing the results a little.)

*Note that apngopt does not guard against writing a larger file than the input. The results here could be slightly smaller if I retained the originals where they were better. 

[edit] A second set:
source | size
--|--
original | 165,721,029 
master | 165,048,775 (0.4% smaller)
with full recompression only | 163,422,620 (1.4% smaller)
with refiltering | 162,850,831 (1.7% smaller)
apngopt (default) | 164,919,387 (0.5% smaller)
apngopt + oxipng | 162,474,452 (2.0% smaller)

### DIFF
--- a/src/apng.rs
+++ b/src/apng.rs
@@ -1,0 +1,75 @@
+use std::io::Write;
+
+use crate::{
+    error::PngError,
+    headers::{read_be_u16, read_be_u32},
+    PngResult,
+};
+
+#[derive(Debug, Clone)]
+/// Animated PNG frame
+pub struct Frame {
+    /// Width of the frame
+    pub width: u32,
+    /// Height of the frame
+    pub height: u32,
+    /// X offset of the frame
+    pub x_offset: u32,
+    /// Y offset of the frame
+    pub y_offset: u32,
+    /// Frame delay numerator
+    pub delay_num: u16,
+    /// Frame delay denominator
+    pub delay_den: u16,
+    /// Frame disposal operation
+    pub dispose_op: u8,
+    /// Frame blend operation
+    pub blend_op: u8,
+    /// Frame data, from fdAT chunks
+    pub data: Vec<u8>,
+}
+
+impl Frame {
+    /// Construct a new Frame from the data in a fcTL chunk
+    pub fn from_fctl_data(byte_data: &[u8]) -> PngResult<Frame> {
+        if byte_data.len() < 26 {
+            return Err(PngError::TruncatedData);
+        }
+        Ok(Frame {
+            width: read_be_u32(&byte_data[4..8]),
+            height: read_be_u32(&byte_data[8..12]),
+            x_offset: read_be_u32(&byte_data[12..16]),
+            y_offset: read_be_u32(&byte_data[16..20]),
+            delay_num: read_be_u16(&byte_data[20..22]),
+            delay_den: read_be_u16(&byte_data[22..24]),
+            dispose_op: byte_data[24],
+            blend_op: byte_data[25],
+            data: vec![],
+        })
+    }
+
+    /// Construct the data for a fcTL chunk using the given sequence number
+    #[must_use]
+    pub fn fctl_data(&self, sequence_number: u32) -> Vec<u8> {
+        let mut byte_data = Vec::with_capacity(26);
+        byte_data.write_all(&sequence_number.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.width.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.height.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.x_offset.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.y_offset.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.delay_num.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.delay_den.to_be_bytes()).unwrap();
+        byte_data.push(self.dispose_op);
+        byte_data.push(self.blend_op);
+        byte_data
+    }
+
+    /// Construct the data for a fdAT chunk using the given sequence number
+    #[must_use]
+    pub fn fdat_data(&self, sequence_number: u32) -> Vec<u8> {
+        let mut byte_data = Vec::with_capacity(4 + self.data.len());
+        byte_data.write_all(&sequence_number.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.data).unwrap();
+        byte_data
+    }
+}
--- a/src/error.rs
+++ b/src/error.rs
@@ -9,6 +9,7 @@ pub enum PngError {
     TimedOut,
     NotPNG,
     APNGNotSupported,
+    APNGOutOfOrder,
     InvalidData,
     TruncatedData,
     ChunkMissing(&'static str),
@@ -33,6 +34,7 @@ impl fmt::Display for PngError {
                 f.write_str("Missing data in the file; the file is truncated")
             }
             PngError::APNGNotSupported => f.write_str("APNG files are not (yet) supported"),
+            PngError::APNGOutOfOrder => f.write_str("APNG chunks are out of order"),
             PngError::ChunkMissing(s) => write!(f, "Chunk {s} missing or empty"),
             PngError::InvalidDepthForType(d, ref c) => {
                 write!(f, "Invalid bit depth {d} for color type {c}")
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -167,7 +167,7 @@ impl RawImage {
             .filter(|c| opts.strip.keep(&c.name))
             .cloned()
             .collect();
-        postprocess_chunks(&mut png, opts, deadline, &self.png.ihdr);
+        postprocess_chunks(&mut png, opts, &self.png.ihdr);
 
         Ok(png.output())
     }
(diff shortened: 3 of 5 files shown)


src/error.rs
```diff
@@ -8,7 +8,7 @@ pub enum PngError {
     DeflatedDataTooLong(usize),
     TimedOut,
     NotPNG,
-    APNGNotSupported,
+    APNGOutOfOrder,
```
andrews05, line 11: Hm, is this a BC break? `APNGNotSupported` should have been removed in v9 but it wasn’t.

src/lib.rs
```diff
@@ -712,38 +714,40 @@ fn postprocess_chunks(
...
+        .with_max_len(1)
+        .enumerate()
+        .for_each(|(i, frame)| {
+            if deadline.passed() {
+                return;
+            }
+            let mut ihdr = png.raw.ihdr.clone();
+            ihdr.width = frame.width;
+            ihdr.height = frame.height;
+            if let Ok(data) = PngImage::new(ihdr, &frame.data).and_then(|image| {
```
andrews05, line 737: Should we error here if decompression fails?

src/lib.rs
```diff
@@ -712,38 +714,43 @@ fn postprocess_chunks(
...
+/// Recompress the additional frames of an APNG
+fn recompress_frames(png: &mut PngData, opts: &Options, deadline: Arc<Deadline>) {
+    if !opts.idat_recoding || png.frames.is_empty() {
+        return;
     }
+    // Use the same filter chosen for the main image
+    // No filter means we failed to optimise the main image and we shouldn't bother trying here
+    let Some(filter) = png.filter else {
+        return;
+    };
```
andrews05, line 728: So I'd be interested in opinions on this. As the comment says, we're unlikely to be able to optimise the fdATs if we failed to optimise the IDAT. The exception is if the image had been previously optimised with the current version of oxipng (before this PR): We will fail to optimise the IDAT any further, but could still optimise the fdATs. The problem is that since IDAT optimisation failed, we won't know what filter to use for the fdATs.
There is a workaround though: using the `--force` option will force the IDAT optimisation, thus providing the necessary filter for the fdATs.

src/error.rs
AlexTMjugador: Yes, it technically is, as it's a publicly viewable enum.

[`cargo semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) is a really useful tool to answer this sort of questions:

image

andrews05: Nice, I've added APNGNotSupported back in and it's not showing that issue. However, it is showing a bunch of issues to do with the recently-added `#[must_use]` to a number of functions, should we worry about this?
I wonder if we should this tool to our workflow...

src/lib.rs
andrews05: I've now made this throw any error from `PngImage::new`, which I think this is the safest thing to do for now.

src/error.rs
AlexTMjugador: Personally, I'd ignore the lints related to `#[must_use]`, as they are too pedantic to matter in practice: they only matter under non-default linter configurations that are a bit of a bad idea to blindly set up. Nevertheless, it's interesting that the tool finds out and let us know about these, just in case my opinion is wrong :slightly_smiling_face:

Adding this tool to our workflow would be a good idea. I didn't read too much on how it computes a baseline for the semver checks though, so making sure that it always picks something sane in a workflow (even if it runs from a PR, or from a release tag...) would be great.

andrews05: Great, that’s kinda what I figured. Are we good to go with this PR then?

AlexTMjugador: Before that, I want to give your seemingly good changes a more proper, in-depth look. Please allow for a bit of time until I get around doing that :smile:

andrews05: Ah sorry, I just assumed you already had. No urgency, I appreciate the scrutiny 🙂

src/apng.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    /// Construct the data for a fdAT chunk using the given sequence number
+    #[must_use]
+    pub fn fdat_data(&self, sequence_number: u32) -> Vec<u8> {
+        let mut byte_data = Vec::with_capacity(4 + self.data.len());
+        byte_data.write_all(&sequence_number.to_be_bytes()).unwrap();
+        byte_data.write_all(&self.data).unwrap();
+        byte_data
+    }
+}
```
AlexTMjugador, line 75: Since the byte `Vec`s returned here are only used to write final PNG chunks, and most of the `Frame` fields implement `Copy`, I believe we could refactor at least `fctl_data` to eliminate the intermediate buffer allocation by returning an `impl Iterator<Item = u8>`, and updating `write_png_block` to accept an `impl IntoIterator<Item = u8>`.

That said, I think such a refactor is beyond the scope of this PR. Moreover, while it would reduce heap allocations, it may also increase code size due to the costs associated with generic parameter monomorphization. Consider this more of a brainstorming note rather than a definite proposal for improvement that should be addressed before merging this PR.

src/png/mod.rs
```diff
@@ -112,27 +119,46 @@ impl PngData {
...
+                    if chunk.name == *b"fcTL" || chunk.name == *b"fdAT" {
+                        // Validate the sequence number
+                        if read_be_u32(&chunk.data[0..4]) != sequence_number {
+                            return Err(PngError::APNGOutOfOrder);
+                        }
+                        sequence_number += 1;
+                        if chunk.name == *b"fcTL" && !idat_data.is_empty() {
+                            // Only create a Frame if it's after the IDAT (else store it as an aux chunk)
+                            frames.push(Frame::from_fctl_data(chunk.data)?);
+                            continue;
```
AlexTMjugador, line 141: The [PNG spec](https://www.w3.org/TR/png-3/#structure) allows `fcTL` to appear before `IDAT` to use the image encoded in `IDAT` as the first frame of the animation:

> The static image may be included as the first frame of the animation by the presence of a single [fcTL](https://www.w3.org/TR/png-3/#fcTL-chunk) chunk before [IDAT](https://www.w3.org/TR/png-3/#11IDAT). Otherwise, the static image is not part of the animation.

So, handling such a `fcTL` as a regular auxiliar, ancillary chunk may change the displayed animation if it gets stripped like other ancillary chunks, which does not seem intended since `fcTL` is indeed a chunk critical for display in the context of APNG.

What was your rationale to write this condition this way? :thinking: :slightly_smiling_face:

AlexTMjugador (commented):
As I initially thought, this looks great overall! I have just a couple of comments to share :slightly_smiling_face:

src/lib.rs
AlexTMjugador: Using `--force` to handle that case sounds like a reasonable first approach to me :+1: 

In theory, I guess there's nothing stopping individual `fdAT` frames from being better compressed using a filter different to that of the main `IDAT` image, especially for APNGs with animations featuring abrupt scene changes. However, performing such an exhaustive filter search is, in my opinion, better suited for an additional option or the "brute" mode, as it's likely to be quite slow.

src/lib.rs
andrews05: Not anymore, it's just the way it was working before. It would try to recompress the fdATs but since we weren't consolidating them it would often fail.

src/apng.rs
andrews05: Yeah, this did bother me slightly but I also thought it wasn't worth dealing with right now. Oxipng is not exactly light on memory usage anyway 😅
That's a good thought for a better approach though, I'll keep it in mind.

src/png/mod.rs
andrews05: So basically at the point we're at right now on the journey towards full APNG support, this just seemed like the easiest way of dealing with that. Note that this is after we've performed the keep/strip check so we know fcTLs are being kept here (stripping of all APNG chunks is permitted).

If we did construct a Frame from an initial fcTL then we'd need additional logic both to handle an "empty" frame and to make sure that it gets written in the correct place in the output.

AlexTMjugador: Hmm, alright, that makes sense. We can reasonably improve on this later on, so no worries for now :+1:

AlexTMjugador (approved):

src/lib.rs
```diff
@@ -716,38 +718,44 @@
...
+                    "Recompressed fdAT #{:<2}: {} ({} bytes decrease)",
+                    i,
+                    data.len(),
+                    frame.data.len() - data.len()
+                );
+                frame.data = data;
+            }
+            Ok(())
+        })
+        .collect()
```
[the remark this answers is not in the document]
TPS: @andrews05, @AlexTMjugador:  Is this note from the bot relevant?  (Sorry for the late query.  I thought it'd be addessed before merging.)  My very-inept researching says this might be net-positive?

PR #671, opened 2025-01-29
Tweak eval cmp key for more consistent results

### LINKED ISSUE #649: Program seems to find new compressions to make after multiple runs (?)
See my commits here https://github.com/EnderNon/Wynntils-jpegnoise-fix

So basically what happened was:

- Ran these for initial commit: 
  - `oxipng -o max --strip all --alpha **/*.png` in `common/src/main/resources/assets/wynntils/textures`, 
  - then ran `oxipng -o max --strip all --alpha logo.png` in `common/src/main/resources/assets/wynntils`
  - creating the commit [refactor: compress every image lossless](https://github.com/EnderNon/Wynntils-jpegnoise-fix/commit/912deb350bf33fd2a8f6baf38ffba125f31d4b8a)
- Realised logo.png became slightly dimmer, 
  - so I regenerated file from before I touched it, with `oxipng -o max --strip safe --alpha **/*.png`
  - creating the commit [fix: use '--strip safe' on logo.png instead. Full command: 'oxipng -o max --strip safe --alpha logo.png'](https://github.com/EnderNon/Wynntils-jpegnoise-fix/commit/f1575e60c2961e7b115dc82f8bbfcf8d525e5328)
 - Then a contributor to upstream Wynntils asked to see what happens if you were to run the first command from the first commit again. 
   - It modified some files here, creating this commit [run oxipng again](https://github.com/EnderNon/Wynntils-jpegnoise-fix/commit/709da7cf2aba4b085fd00c5ffea8a6e4ed031143)
 - Running it again afterwards showed in the output that there was no new differences. 
So for some reason, running the command once wasn't enough to achieve max savings - running it twice was necessary. Seems like a bug.

andrews05:
This changes the evaluator to prefer a later image instead of an earlier one in the case of a tie. This gives more consistent results since the baseline is always added last.

I also removed bit depth from the key since the uncompressed size should cover that.

Fixes #649.
(Mostly. I did discover another possible but very rare issue: if multiple colours tie for "most popular edge colour" in the luma sort, the sort may affect which one is actually picked and multiple runs may flip-flop back and forth)

### DIFF
--- a/src/evaluate.rs
+++ b/src/evaluate.rs
@@ -23,7 +23,7 @@ pub(crate) struct Candidate {
     pub idat_data: Vec<u8>,
     pub filtered: Vec<u8>,
     pub filter: RowFilter,
-    // first wins tie-breaker
+    // For determining tie-breaker
     nth: usize,
 }
 
@@ -32,9 +32,9 @@ impl Candidate {
         (
             self.idat_data.len() + self.image.key_chunks_size(),
             self.image.data.len(),
-            self.image.ihdr.bit_depth,
             self.filter,
-            self.nth,
+            // Prefer the later image added (e.g. baseline, which is always added last)
+            usize::MAX - self.nth,
         )
     }
 }


ace-dent:
@andrews05 

> ... prefer a later image ... since the baseline is always added last.

For *my* use, preferring the input image (which has been heavily pre-optimised), in case of a tie, is preferred. 👍

>  ... if multiple colours tie... in the luma sort ...

Nice catch! I still ponder whether there is something smarter we can do here around edge colours. I still haven't quite got my head around potential interactions with filtering (beyond `f0`).... 🤔

AlexTMjugador (approved):
Thanks, LGTM!

PR #672, opened 2025-01-29
Fix size passed to fast filter evaluator

andrews05:
The evaluator uses the `estimated_output_size` internally but we were passing only the idat size in the fast filter evaluator. At level `o2`, or higher levels when using `--fast`, this could cause a slight bias towards the None and Bigrams filters for indexed images. The actual impact to results is incredibly minor but it's definitely incorrect so we should fix it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -509,7 +509,7 @@ fn perform_trials(
                 opts.optimize_alpha,
             );
             if let Some(result) = &eval_result {
-                eval.set_best_size(result.idat_data.len());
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
Thank you!

ace-dent:
Nice fix. Can I check, when all is equal, `f0` would be selected? For some embedded systems, there is a computational benefit of avoiding `f4` & `f5`, if there's no compression benefit.

andrews05:
@ace-dent Yes, we use the lowest numbered filter when there’s a tie.

PR #673, opened 2025-01-29
Don't use most popular edge colour if tied

andrews05:
If multiple edge colours were tied for most popular, the one chosen would depend on the current palette order, which could cause it to change if run a second time.

This further improves determinism for #649. There is still one more issue I've found affecting the images there which I'll try to address in another PR.


AlexTMjugador (approved):
Looks reasonable to me, thanks!

PR #674, opened 2025-01-29
Improve indexed to grayscale conversion

andrews05:
If no palette reduction occurred (or palette reductions were off), a fully transparent palette entry may not have been zeroed to black. By ensuring this gets cleaned in the `indexed_to_channels` transformation, we may able to achieve a grayscale conversion that would otherwise have been RGB.

This is the final piece of the puzzle in #649 to achieve better file sizes on the first run and avoid further changes on a second run.

Results from the images in #649 (combining this PR and #673):
9.1.3 1st run: 2,191,773
9.1.3 2nd run: 2,191,629
PR 1st run: 2,191,380
PR 2nd run: 2,191,380


AlexTMjugador (approved):
Awesome catch! :+1:

PR #676, opened 2025-02-03
Reduction performance tweaks

andrews05:
Two key changes here:
- Removed some defunct reduction benches and added some more alpha reduction benches.
- Changed use of `chunks()` to `chunks_exact()` where appropriate to improve performance. (As a byproduct, this also fixes a potential crash if a malformed file has a PLTE chunk that isn't a multiple of 3.)

### DIFF
--- a/src/reduction/alpha.rs
+++ b/src/reduction/alpha.rs
@@ -17,7 +17,7 @@ pub fn cleaned_alpha_channel(png: &PngImage) -> Option<PngImage> {
     let colored_bytes = bpp - byte_depth;
 
     let mut reduced = Vec::with_capacity(png.data.len());
-    for pixel in png.data.chunks(bpp) {
+    for pixel in png.data.chunks_exact(bpp) {
         if pixel.iter().skip(colored_bytes).all(|b| *b == 0) {
             reduced.resize(reduced.len() + bpp, 0);
         } else {
@@ -46,7 +46,7 @@ pub fn reduced_alpha_channel(png: &PngImage, optimize_alpha: bool) -> Option<Png
     let mut has_transparency = false;
     let mut used_colors = vec![false; 256];
 
-    for pixel in png.data.chunks(bpp) {
+    for pixel in png.data.chunks_exact(bpp) {
         if optimize_alpha && pixel.iter().skip(colored_bytes).all(|b| *b == 0) {
             // Fully transparent, we may be able to reduce with tRNS
             has_transparency = true;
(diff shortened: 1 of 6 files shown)


src/reduction/alpha.rs
```diff
@@ -75,12 +75,15 @@ pub fn reduced_alpha_channel(png: &PngImage, optimize_alpha: bool) -> Option<Png
...
+    for pixel in png.data.chunks_exact(bpp) {
         match transparency_pixel {
             Some(trns) if pixel.iter().skip(colored_bytes).all(|b| *b == 0) => {
                 raw_data.resize(raw_data.len() + colored_bytes, trns);
             }
-            _ => raw_data.extend_from_slice(&pixel[0..colored_bytes]),
+            _ if colored_bytes == 6 => raw_data.extend_from_slice(&pixel[0..6]),
+            _ if colored_bytes == 3 => raw_data.extend_from_slice(&pixel[0..3]),
+            _ if colored_bytes == 2 => raw_data.extend_from_slice(&pixel[0..2]),
+            _ => raw_data.push(pixel[0]),
```
andrews05, line 86: This does end up 30-40% faster in the benchmark, but is there a more generic way to improve performance here?

AlexTMjugador: This reminds me of [performance issues with how Rust emits `memcpy` calls that I've stumbled upon before](https://github.com/rust-lang/rust/issues/92993). What I learned there is that as application programmers we only can come up with these specific constructions to convince the compiler to emit more efficient code, which is at odds with the generality we usually want, so the answer to your question is probably not.

However, I think adding a comment stating why you made specialized branches for a few colored bytes count would help with code maintenance.

andrews05:
The `chunks_exact` makes a big difference to the 16-to-8 benchmarks. (Still not appreciable in actual use though.)
Before:
```
test reductions_16_to_8_bits                        ... bench:     352,770.85 ns/iter (+/- 11,464.83)
test reductions_16_to_8_bits_scaled                 ... bench:     748,591.70 ns/iter (+/- 7,528.53)
```
After
```
test reductions_16_to_8_bits                        ... bench:     200,931.25 ns/iter (+/- 4,184.77)
test reductions_16_to_8_bits_scaled                 ... bench:     116,423.44 ns/iter (+/- 652.00)
```

src/reduction/alpha.rs
AlexTMjugador: Sounds reasonable, feel free to visit this when you want to!

AlexTMjugador (approved):
LGTM, thank you!

PR #681, opened 2025-02-14
Use current version in `pre-commit` example

Chubercik:
Simple enough change, use the current version in the `README.md`'s `pre-commit` example.


AlexTMjugador (approved):
Thank you, a simple change indeed!

PR #692, opened 2025-04-14
Switch Docker image from `alpine` to `scratch`

AFCMS:
Following https://github.com/shssoichiro/oxipng/pull/642

- [x] Fail to build on linting errors
- [x] Switch Docker image from `alpine` to `scratch`

Reduce the image size roughly from 15Mb to 2Mb.

image

It doesn't break the usecase mentioned in the README, as well as people copying the binary in a Docker step.

I have considered making the image rootless but since it's doing file modifications it requires overwriting the user to bring back root in the README's usecase. The image would benefit of if using it only with stdin/stdout. Not sure it's worth it.

I have been working on enabling cross compilation for non-Tier 1 platforms, but no ETA yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,7 +31,7 @@ RUN --mount=type=cache,target=/root/.cargo/git/db \
     xx-verify /src/target/$(xx-cargo --print-target-triple)/release/oxipng && \
     cp /src/target/$(xx-cargo --print-target-triple)/release/oxipng /src/target/oxipng
 
-FROM alpine AS tool
+FROM scratch AS tool
 
 LABEL org.opencontainers.image.title="Oxipng"
 LABEL org.opencontainers.image.description="Multithreaded PNG optimizer written in Rust"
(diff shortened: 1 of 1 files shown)


AlexTMjugador (approved):
Looks great to me, thank you! ❤️

I was wondering if you've also considered using [distroless base images](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md)? They're not quite as barebones, as they include extra essential files like `/etc/passwd`, which, although not currently used by Oxipng, could be helpful for future expansions of its feature set. Plus, they reportedly have similarly small base image sizes, of around 2 MiB.

AlexTMjugador:
You make a very good point about it being easier to extend a Docker image than to trim it down. And you're right, Oxipng doesn't do any network communications or timezone handling, so it doesn't require those supporting files. I'm also unsure about how Docker handles `/tmp`, but I don't think it's important either because Oxipng does not create any temporary files at the moment. Let's go ahead and merge this as is!

PR #697, opened 2025-04-24
Bump version to 9.1.5

andrews05:


AlexTMjugador (approved):
Neat, let's do this release! :tada:

andrews05:
@AlexTMjugador Something's gone wrong with the deploy. I think it tried to create the release multiple times?

andrews05:
I re-ran the workflow and it worked 🤷‍♂️

AlexTMjugador:
After reviewing the CI workflow file and recent runs, I believe the flakiness you've noticed is likely due to race conditions with the GitHub Releases API, as discussed in https://github.com/softprops/action-gh-release/issues/445#issuecomment-2407940052.

I've pushed a commit that works around this suspected issue by running each matrix job sequentially. This does increase the total execution time, but hopefully not by much: each job takes about 30 seconds, so eight jobs should complete in around four minutes, which sounds reasonable for an occasional workflow. (At the moment, some jobs take around a minute because of `cargo-deb` installation, but I plan to optimize this by switching to `cargo-binstall`, which completes in negligible time.)

andrews05:
@AlexTMjugador Nice work. Arm runners are finally available in public preview now, might be worth using them - maybe it will resolve the deb issue mentioned in your latest commit.

AlexTMjugador:
Thanks for the reminder, @andrews05! I have actually migrated the CI workflows of other projects to ARM runners, and the results have been positive: the workflows were both simplified and sped up.

For Oxipng, I decided to hold off for now due to the perceived increased likelihood of needing a cross-compilation setup to support exotic architectures that users might request. (Oxipng is the only project I maintain where people have asked for 32-bit Windows builds, for example.) However, with each passing day that such requests don't materialize, I'm becoming less convinced that the tradeoff is worthwhile, so I'm open to eventually making the switch if things continue that way.

ace-dent:
@AlexTMjugador - this user with odd/ esoteric w32 build requirements, thanks you! 🙏

PR #707, opened 2025-06-22
Change repository name reference + Docker registry URL

AFCMS:
All the urls in READMEs and such caused a redirect when opening, but most importantly the URL of the Docker registry was now wrong (will only have old versions of the software).

I wonder if it would be useful to mention that old versions of the software can be found at the old URL in the README?

https://github.com/users/shssoichiro/packages/container/package/oxipng

**Also new packages are private by default on GitHub, so the `ghcr.io/oxipng/oxipng` package is private at the moment (the CI passes but nobody outside the team can see/access the resulting package).**

image

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -89,7 +89,7 @@ jobs:
               # so make the binary world-executable to meet user
               # expectations set by preceding releases.
               # Related issue:
-              # https://github.com/shssoichiro/oxipng/issues/575
+              # https://github.com/oxipng/oxipng/issues/575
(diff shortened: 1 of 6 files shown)


shssoichiro (approved):

shssoichiro:
Thanks, I've made the package public. It might not be a bad idea to note the location of previous builds as well. :thinking:

AFCMS:
Added the note about older images, plus linked directly to the packages pages (to make it easy to see available tags).

AlexTMjugador (approved):

AlexTMjugador:
Good catches, I think we can merge this PR :+1:

PR #713, opened 2025-08-02
Reduce memory usage of fast evaluation

andrews05:
This is an experiment I started a while ago before life happened. It reduces memory usage of fast evaluation (-o2 and lower), bringing it inline with normal (slow) evaluation. It does this by not retaining the filtered image data of the evaluations, but instead retaining the row filters that were used in each line so it can be quickly re-filtered when required. This does incur a tiny performance penalty, but it's negligible even at `-o0`.
Although this is fully functional, there are a few rough spots in the code so I'm just opening it as a draft for now.

```
PR -sao6
       70.08 real       614.36 user         2.66 sys
          2024243200  maximum resident set size
          1839339520  peak memory footprint

PR -sao2
       11.16 real        60.85 user         1.33 sys
          1982562304  maximum resident set size
          1842173824  peak memory footprint

PR -sao2 -t1
       55.11 real        53.85 user         0.76 sys
           429457408  maximum resident set size
           245008064  peak memory footprint

master -sao6
       67.70 real       616.07 user         2.72 sys
          2043379712  maximum resident set size
          1838340416  peak memory footprint

master -sao2
       11.53 real        60.63 user         1.25 sys
          2753396736  maximum resident set size
          2283741440  peak memory footprint

master -sao2 -t1
       54.29 real        53.55 user         0.72 sys
           626311168  maximum resident set size
           305252544  peak memory footprint
```

Note that this involves some refactoring of `RowFilter` and the new `FilterStrategy`. These are breaking changes so it will ultimately be destined for v10. One advantage to this new structure is it opens the door for future changes such as allowing the Brute strategy to take a parameter for the number of lines.

### DIFF
--- a/benches/filters.rs
+++ b/benches/filters.rs
@@ -13,7 +13,10 @@ fn filters_16_bits_filter_0(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from("tests/files/rgb_16_should_be_rgb_16.png"));
     let png = PngData::new(&input, &Options::default()).unwrap();
 
-    b.iter(|| png.raw.filter_image(RowFilter::None, false));
+    b.iter(|| {
+        png.raw
+            .filter_image(FilterStrategy::Basic(RowFilter::None), false)
+    });
 }
 
 #[bench]
@@ -21,7 +24,10 @@ fn filters_8_bits_filter_0(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from("tests/files/rgb_8_should_be_rgb_8.png"));
     let png = PngData::new(&input, &Options::default()).unwrap();
 
-    b.iter(|| png.raw.filter_image(RowFilter::None, false));
+    b.iter(|| {
+        png.raw
+            .filter_image(FilterStrategy::Basic(RowFilter::None), false)
+    });
 }
 
 #[bench]
@@ -31,7 +37,10 @@ fn filters_4_bits_filter_0(b: &mut Bencher) {
     ));
     let png = PngData::new(&input, &Options::default()).unwrap();
 
-    b.iter(|| png.raw.filter_image(RowFilter::None, false));
+    b.iter(|| {
+        png.raw
+            .filter_image(FilterStrategy::Basic(RowFilter::None), false)
+    });
 }
 
 #[bench]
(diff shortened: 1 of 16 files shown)


benches/filters.rs
```diff
@@ -13,15 +13,21 @@ fn filters_16_bits_filter_0(b: &mut Bencher) {
     let input = test::black_box(PathBuf::from("tests/files/rgb_16_should_be_rgb_16.png"));
     let png = PngData::new(&input, &Options::default()).unwrap();
 
-    b.iter(|| png.raw.filter_image(RowFilter::None, false));
+    b.iter(|| {
+        png.raw
+            .filter_image(FilterStrategy::Basic(RowFilter::None), false)
```
andrews05, line 18: Using the basic filters is more verbose now. There are a couple of options for improving this if we feel the need:
1. Implement `From<RowFilter>` on `FilterStrategy` so we can write:
`RowFilter::None.into()`
3. Add constants on `FilterStrategy` like `NONE = Self::Basic(RowFilter::None)`, so we can write:
`FilterStrategy::NONE`

src/evaluate.rs
```diff
@@ -17,14 +17,17 @@ use rayon::prelude::*;
...
+};
 
 pub(crate) struct Candidate {
     pub image: Arc<PngImage>,
     pub data: Vec<u8>,
     pub data_is_compressed: bool,
     pub estimated_output_size: usize,
-    pub filter: RowFilter,
+    pub filter: FilterStrategy,
+    pub filter_used: FilterStrategy,
```
andrews05, line 30: These two are confusing. I should come up with clear names and add descriptive comments.

andrews05: Perhaps a better option would be to just make `Predefined` take a string description of where it came from.

For a radically different idea: Maybe the filters can just be different implementations of a trait, with all the heuristic logic contained within the implementation. This could even allow library implementers to create their own strategies. I'm not sure I'm prepared to do such a refactor right now though.

benches/filters.rs
andrews05: I've implemented the second option for now, but open to suggestions.

andrews05:
The predefined is causing a regression with APNGs, where it uses the same row filters it generated for the first frame for every subsequent frame, rather than re-running the heuristic strategy. I'll have to investigate how to fix this.

src/evaluate.rs
andrews05: Okay, I've reverted to the two properties on Candidate, as we need to retain the input filter for APNG frames.

andrews05:
> The predefined is causing a regression with APNGs, where it uses the same row filters it generated for the first frame for every subsequent frame, rather than re-running the heuristic strategy. I'll have to investigate how to fix this.

Fixed by reverting to two filter properties on `Candidate` instead of just keeping a string name on `Predefined`.

andrews05:
Thanks heaps @AlexTMjugador!

Issue #658, opened 2024-12-03
`--zi` options with parameter higher than 255?


dustdfg:
I know I am crazy but if 255 is an artificial restriction, I want to set it to 1024 or something like it. For three reasons:

1. I am crazy
2. Optimize for any cost
3. Want to look how long it will take (I've already tried 255)

Is it possible to edit one line in code, recompile custom binary and run with 1024?

andrews05:
Hi @dustdfg
See #640 where `--zi` was added. The reason for the restriction is that the API currently uses a U8 and raising it would be a breaking change, so it's been deferred until the next major release (no timeline for that yet though).
If you want to change it yourself, take a look at the diff in the above PR. You'll need to change one reference of "255" to something higher, and also change all references of `NonZeroU8` to NonZeroU64`.

dustdfg:
Thanks. It worked. I've run it with parameter 10240 just for fun and saved 2 bytes in comparison to already optimized image with 255 :rofl: :rofl: :rofl: But it worked!

AlexTMjugador:
Out of curiosity, how much time did it take to run 10240 iterations with respect to "just" 255? :joy:

TPS:
> Out of curiosity, how much time did it take to run 10240 iterations with respect to "just" 255? :joy:

& on what spec'ed image (dimensions & color depth, mostly)?

dustdfg:
![godot_icon](https://github.com/user-attachments/assets/df7c8828-d873-4d88-bee4-8e6cc5b22506)
It was a [Godot icon](https://github.com/godotengine/godot) previously optimized by setting `--zi 255` to **3683 bytes**. When I ran `--zi 10240` it gave me 2 bytes difference. It took more than compiling Godot for this one image. When I ran `--zi 255` for all png files in Godot it also took more than compiling Godot. I can say that running `--zi 10240` was at least 34 times slower than running `--zi 255` for all Godot icons

I've also ran it on another icon (also from godot):

![icon_foreground](https://github.com/user-attachments/assets/ee8158b8-fad4-4dbf-8372-546e5b36db56)
It also previously was optimized with `--zi 255`. And running it with `--zi 102400` (**102400** it is not a mistake there are two zeroes in the end!) it didn't have any effect but took 6-9 hours

TPS:
Is there an iteration counter displayed saying _which_ round produced the savings?

[...]

PR #639, opened 2024-07-21
Note that higher levels are not guaranteed to be better

andrews05:
Closes #638

Issue #638, opened 2024-07-19
`-o max` is worse than `-o 2` in some cases


zzyyyl:
version v9.1.2

Test image 1 (1030 Bytes): ![test](https://github.com/user-attachments/assets/f3ed700b-7fa3-4bcc-a7f8-09740cc73900)

```sh
> oxipng -o max -s -Z test.png
test.png: Could not optimize further, no change written

> oxipng -o max -s test.png
test.png: Could not optimize further, no change written

> oxipng -o 2 -s test.png
1028 bytes (0.19% smaller): test.png
```

Test image 2 (147 Bytes): ![RoguelikeRecruitSupportElite0](https://github.com/user-attachments/assets/5e8bae5f-6a2b-4a24-996e-92d6ede21716)

```sh
> oxipng -o max -s -Z test.png
test.png: Could not optimize further, no change written

> oxipng -o max -s test.png
test.png: Could not optimize further, no change written

> oxipng -o 2 -s test.png
140 bytes (4.76% smaller): test.png
```

ace-dent:
Further reading: 
https://github.com/ebiggers/libdeflate/issues/241#issue-1420436397
https://github.com/shssoichiro/oxipng/issues/545
https://github.com/shssoichiro/oxipng/issues/568#issuecomment-1764127487

andrews05:
Hi @zzyyyl
As noted in the references above, this is indeed possible on occasion and is actually pretty normal behaviour in the world of lossless data compression. The differences here are a matter of bytes and shouldn't be anything to be concerned about.

As this has come up a couple of times, I'll make a PR to note this in the read me...

AlexTMjugador:
> As noted in the references above, this is indeed possible on occasion and is actually pretty normal behaviour in the world of lossless data compression. The differences here are a matter of bytes and shouldn't be anything to be concerned about.

While I agree that this is a very reasonable expectation to have in general, I'm curious about whether these images have some technical property whose analysis could result in improvements to optimization heuristics. Did you suspect or find out anything of that sort, @andrews05?

andrews05:
I didn't look at the first image but in the second one it selected 1-bit at -o2 and 8-bit at -o6. This can happen because the evaluation compression level changes from 7 to 8, which gives a more accurate selection most of the time but still isn't perfect of course.

PR #720, opened 2025-08-19
Allow brute filter configuration

andrews05:
This PR allows the Brute filter parameters to be configured rather than using fixed constants. It also tweaks the presets to take advantage of these new options:
o3: level=1, lines=3 (fractionally worse/faster)
o4: level=1, lines=4 (same as before)
o5: level=4, lines=4 (fractionally better/slower)
o6: level=5, lines=8 (slightly better/slower)

The parameters can be explicitly set in both the API and the CLI.

### DIFF
--- a/src/png/mod.rs
+++ b/src/png/mod.rs
@@ -513,14 +513,14 @@ impl PngImage {
                         let mut compressor =
                             Compressor::new(CompressionLvl::new(level.into()).unwrap());
                         let limit = filtered.len().min((line.data.len() + 1) * num_lines);
-                        let capacity = compressor.zlib_compress_bound(limit);
+                        let capacity = compressor.deflate_compress_bound(limit);
                         let mut dest = vec![0; capacity];
 
                         for f in try_filters {
                             f.filter_line(bpp, &mut line_data, &prev_line, &mut f_buf, alpha_bytes);
                             filtered[line_start..].copy_from_slice(&f_buf);
                             let size = compressor
-                                .zlib_compress(&filtered[filtered.len() - limit..], &mut dest)
+                                .deflate_compress(&filtered[filtered.len() - limit..], &mut dest)
                                 .unwrap_or(usize::MAX);
                             if size < best_size {
                                 best_size = size;


TPS:
> An open question is whether we should continue the existing pattern of short long-form options like `--bl` or do them properly like `--brute-lines`.

The great thing is, since you're _adding_ options, do whatever (all of the following obv imho):
 - The consistency part comes later, when _maintaining_ them: don't willy-nilly _change/remove_ them w/o providing some sort of backwards compatibility, to avoid breaking stuff downstream
- Short options for commonly used switches, long 1s for the more obscure.  That _should_ satisfy anyone.
 - _Especially_ on Windows terminals, _how_ to use each kind is constant, e.g., does 1 use a space, equals, colon, &c between a switch & parameter, & what delimiters are _required_ (none, quotes, double-quotes, &c).


Sorry for ranting.  Not experiencing this w/ `oxipng` per se, but I hate having to search through a tool's longhelp scanning for a ½-recalled option, just to go "how on earth was I supposed to remember _that_ ?!"

andrews05:
> * The consistency part comes later, when _maintaining_ them: don't willy-nilly _change/remove_ them w/o providing some sort of backwards compatibility, to avoid breaking stuff downstream

Exactly, which is why it's important to get it right now so that we don't decide later that we made the wrong choice but can't change it until v11 due to being a BC break.

> * Short options for commonly used switches, long 1s for the more obscure.  That _should_ satisfy anyone.

These are definitely more obscure options. Proper long form* may indeed be the right choice.

> * _Especially_ on Windows terminals, _how_ to use each kind is constant, e.g., does 1 use a space, equals, colon, &c between a switch & parameter, & what delimiters are _required_ (none, quotes, double-quotes, &c).

I think we already cover this well enough 🙂

*I say "proper" long form to distinguish from some of our existing options like `--zc` which are technically long-form by virtue of being more than one character and requiring a double dash, but obviously aren't very long. I'm not sure this is good practice. The purpose of long form is to convey the meaning of the option right in the name.
That said, something like `--zopfli-iterations-without-improvement` seems a bit too much of a mouthful, so abbreviations like `--ziwi` may be appropriate in some cases.

TPS:
> *I say "proper" long form to distinguish from some of our existing options like `--zc` which are technically long-form by virtue of being more than one character and requiring a double dash, but obviously aren't very long. I'm not sure this is good practice. The purpose of long form is to convey the meaning of the option right in the name.

This _is_ confusing to me & likely others.  AfaIk, `oxipng` is the only app that has double-dash short-forms.  Would it cause mayhem to _also_ accept single-dash for all short forms?

andrews05:
This is the distinction:

Short form:
- Single character
- Single dash
- Can be combined together, e.g. `-s -a -v` can be `-sav`

Long form:
- Multiple characters
- Two dashes
- Cannot be combined

So despite being very "short", `--zc` and others like `--nx`, `--nz` etc are technically long form, and can't be made short form without reducing them to a single character. Many programs, especially older ones, do not follow this convention and just always use single dashes regardless, but this is not considered good practice today.
It might be better for e.g. `--nx` to be written as `--no-reductions`, but I'd be reluctant to make changes like that just for the sake of it.

ace-dent:
Hey @andrews05 I got lost in the pedantry of naming here... rather than the bigger change! I don't really understand the intent here... 😅 

1. Currently what default settings are used for `brute`? Is it `zc=12` (highest libdeflate setting) and `lines=1`?
2. Is your lines setting to effectively skip rows? So test every other row? Or...?
3. If the intent is to increase speed, will you be performing benchmarking to make sure we don't regress  compression here? 

I'm a niche user who wants everything maxed out, with no regard for time... but .... Lower levels of libdeflate favour images that compress well with RLE / Huffman only, etc. ... It may be beneficial to run with a low and a high setting for libdeflate...?

andrews05:
@ace-dent Since this is a draft, I didn't initially bother adding details to the top post (was going to add it later). I've now updated it with a little explainer. See also the long help for more info on the brute options.

I know libdeflate did have a problem where it wasn't trying Huffman-only strategies but that has since been resolved. I'm not aware of any remaining issues there?

src/main.rs
```diff
@@ -205,6 +204,21 @@ fn parse_opts_into_struct(
...
+            if let Some(new_lines) = matches.get_one::<usize>("brute-lines") {
+                *num_lines = *new_lines;
+            }
+            if let Some(new_level) = matches.get_one::<i64>("brute-level") {
+                *level = *new_level as u8;
+            }
+        }
+        new_filters.insert(f);
+    }
+    opts.filter = new_filters;
```
andrews05, line 220: This feels a bit awkward but I'm not sure if there's a better way to do it. Values in an index set aren't mutable like they are in a vec.

ace-dent:
Thanks @andrews05.

> ... I know libdeflate did have a problem where it wasn't trying Huffman-only strategies ...

It's more a 'feature', where higher levels favour more complex compression strategies, which can prove less efficient in edge-cases. e.g. #638. Hence for exhaustive testing, I still test `--zc 1-12`.

ace-dent:
I still don't really get these new parameters... hopefully it will all become apparent in time! 😅

andrews05:
For each line, the brute method tries each filter and compresses the result to see which one is smallest. On it's own this isn't very effective though, so we include a few previous lines as context when compressing. The lines parameter is the total number of lines we compress at once.

Winterhuman:
@andrews05 Quick question. When it says "Higher values are not necessarily better", does that mean using higher line or brute levels can yield worse compression than lower levels, or just that it won't yield anything better most of the time?

TPS:
@Winterhuman We've already had examples of this raised, e.g. #638 → #639.  I believe that text is to clarify that unpredictable result.

Winterhuman:
Ah, so that's true of these line & brute levels too, gotcha. Thanks for clarifying!

andrews05:
So the Brute filter, like any of the other heuristic filters, is just an attempt to find filters that compress well. It works reasonably well, but a good outcome is not guaranteed.
From my own testing and observations, increasing the parameters can yield better results *on average*, but only to a point, and yes, it can be worse for some images. At `-omax` it's using level=5, lines=8. Beyond this you're essentially just adding random variance into the mix which may find a good result for one image, but not others. On one of my test sets I did find overall improvement with lines=16, but that didn't seem to hold true on other test sets. See also https://github.com/oxipng/oxipng/issues/730#issuecomment-3310133672 where I got a good result on one image with level=5, lines=6, but everything else I tried was worse.

You may like to experiment for yourself and see what you find, but at the end of the day these are advanced settings that most users shouldn't be touching.

[edit] One might assume that maxing it out with level=12, lines=&lt;height of your image&gt; would give the best result. This is not the case though, and to be honest I don't know why. (It would also be incredibly slow if you did this)

ace-dent:
@andrews05 - Does this make sense for a brute force script? ...

```sh
oxipng --zopfli --zi 9000 --ziwi 300 --filters 0-8 "${png_file}"
oxipng --zopfli --zi 9000 --ziwi 300 --filters 9 --brute-level 1 --brute-lines 4 "${png_file}"
oxipng --zopfli --zi 9000 --ziwi 300 --filters 9 --brute-level 12 --brute-lines 8 "${png_file}"
```

Rationale: I want to test `Brute` filter with small and larger context windows (4 vs 8 lines).

andrews05:
If you really want to brute force it you'll need to permute all combinations of levels and lines 😂
But if you're wanting to be more pragmatic and just try a few options, I'd probably stick with level 5 and just vary the lines, e.g. 4, 8, 12, 16, or even more if want to experiment.
In my own tests, I did not find that higher levels were beneficial. And as noted above, fiddling with the parameters is often just a matter of adding random variance that may or may not work well for any given image. There's no real science here unfortunately.

Winterhuman:
(Currently working on all those permutations btw. Stay tuned)

ace-dent:
Thanks @andrews05. And @Winterhuman I look forward to your results.
I *do* have a theory why higher line numbers may hinder compression... but need to check what the code is actually doing.

Winterhuman:
The updated version of Sisyphus is now available with the new brute permutations (`--brute-level {1..12}` & `--brute-lines {2..16}`): https://gist.github.com/Winterhuman/21d7b148db40ff041f397b07a7aafb83

Be warned: up to 73000+ trials are created; a 96-byte favicon took nearly 3.5 minutes. The script tries to truncate & symlink identical & invalid trials where it can, but expect high CPU and memory (`--max-procs` can be specified to reduce or increase the time/computation cost; the default is 8).

TPS:
_(@Winterhuman Did you see https://github.com/oxipng/oxipng/issues/658 for further "inspiration"?)_

Winterhuman:
Actually, hold on, does `--brute-level`/`--brute-lines` actually do anything when `--zopfli` is specified? Just double checking in case I can save on permutations

andrews05:
@Winterhuman Yes, but you probably want to do what @ace-dent was doing above, running filters 0-8 on their own and then permuting the brute options with only the brute filler. And feel free to go above 16 lines, I seem to recall one image got an improvement for me at around 128 lines. You probably want to cap it to the height of the image though.

andrews05:
Also, you may be interested in checking out #749. It should remove the need to permute the -n flags.

Winterhuman:
@andrews05 I've updated the script with your feedback!

1. The `--filters 0-8` and `--filters 9 --brute-level --brute-lines` variations are now separate.
2. The script now limits `--brute-lines` to the source image's height; and also, the script now has its own `-b|--brute-lines` option to set the maximum `oxipng --brute-lines` value to try up to (the default is still 16).

PR #724, opened 2025-08-25
Update runners, rust, and dependencies

andrews05:
#719 is failing tests due to requiring a newer version of rust than we currently specify. This PR updates to 1.85.1 and sets the edition to 2024.
I've also updated dependencies and runner images, using the ubuntu arm runner and removing the need for qemu.

Closes #719.


andrews05:
Note some of my PRs will conflict with each other. I'll rebase and fix them as they get merged.

AlexTMjugador (approved):
Thanks, LGTM! I'll miss having the multiarch and QEMU setup in CI, since it was a handy remainder to myself on how to cross-compile and test Rust binaries from x64 to AArch64, which is something I need to do somewhat often. But now that ARM runners are available, I think it's best for us to use them :smile:

PR #732, opened 2025-09-21
Add max_uncompressed_size

### LINKED ISSUE #411: Resource limits for RAM, dimensions, and CPU time
I use oxipng to process untrusted files from uploaded users, and we currently rely on pod level resource limits in order to protect against PNG bombs. Is it possible for there to be resource limits by oxipng itself, ala ImageMagick?

andrews05:
Adds an option to skip files when the uncompressed IDAT (or the file size itself) exceeds a certain size. This provides an effective way to limit resource usage in constrained environments.

Closes #411.


AlexTMjugador (approved):
LGTM!

For anyone casually looking at this PR, please note that it is _not_ intended to be an ideal and watertight solution for limiting OxiPNG's resource usage in every situation. If you need a high level of assurance in resource usage bounds for security purposes, it's still best to use operating system features for it, such as `cgroups` on Linux (which can be used in easier ways through containers and `systemd-run`, as mentioned in the linked issue). This PR is best viewed as an _usually_ effective way to bound resource use, but there may be some edge cases, especially with weirdly formatted PNGs, that may warrant further measures tailored for your use case.

andrews05:
@AlexTMjugador well said!

ace-dent:
@andrews05 - genuinely not trying to be pedantic... but... is `MB` here actually `MiB` or truly `MB`, etc. ?
https://github.com/oxipng/oxipng/blob/aebddf443a7d1625e133d26e36669f9037ca94e7/src/cli.rs#L396

Presumably the 'hidden' raw option doesn't need a units suffix... it's always bytes and therefore unambiguous. How much does including `parse-size` increase the compiled binary size? I personally prefer a minimal unit approach, to avoid any doubt with metric -vs- binary units.

I like the example of what is a reasonable value, to guide the User. I wonder if it's more common to have 32bit, alpha channel images?

andrews05:
@ace-dent I believe MB will be 1,000,000, but it supports MiB as well if you prefer. It's surprisingly flexible, you can do things like "14.2e+8" which will parse as 1,420,000,000. And yes, the default when no unit is provided is bytes.

In practice, I imagine users wanting to make use of this feature are unlikely to be too concerned about the exact value they provide.

[edit] Missed your question about binary size. A quick check before and after this PR shows about a 1% difference.

ace-dent:
Thanks @andrews05. This is a really great addition! 🙏

PR #734, opened 2025-10-03
Small tweaks

### LINKED ISSUE #728: Improve documentation for `-fix` option
I wonder if the current help text can be improved?

`--fix  Disable checksum validation`

`--fix  Do not perform checksum validation of PNG chunks. This may allow some
          files with errors to be processed successfully.`

Is this accurate and correct? My understanding based on #5 (but not code review), is that it will **_repair_** CRC errors (unless we have forbidden re-encoding the IDAT)...?

Should it say ~ `Attempt to process and repair files with checksum errors`?

BTW, I still consider this risky; I believe it's better to reject any broken file. Even pngcheck [removed](https://github.com/pnggroup/pngcheck/blob/6419815370bb808d238b3b78eeac13811d0abb3e/README-303#L37-L39) `-f` to tolerate broken files!! However, server side users have voiced it's an useful feature.

### LINKED ISSUE #733: Documentation for Entropy filter is misleading: it actually selects the filter with lowest entropy
The docs for the entropy filter option state that the selection criteria is the highest Shannon entropy:

https://github.com/oxipng/oxipng/blob/14532e8bf54e75f5c280017c1e93fa2d323b94b7/MANUAL.txt#L122

However, the formula for Shannon entropy actually contains a negative sign (per Wikipedia):

Image

The code leaves off this negation, and ends up instead computing the negation of the shannon entropy (plus a constant factor). That doesn't cause any correctness issue because the same filter is still chosen. However, it means that what looks like a maximization is actually minimizing the Shannon entropy.

This makes sense from a logical perspective. The Shannon entropy is an estimate of how many bits it takes to encode a value. The fewer bits required, the more compressible a given input is. Thus the filter with the lowest Shannon entropy is more likely to be the better pick.

andrews05:
- Fix docs for entropy filter (fixes #733)
- Improve docs for `--fix` (fixes #728)
- Add tests for alpha optimisation
  - Following a recent change to the code here, I realised we don't actually have any tests covering alpha optimisation. I've adjusted some of the test images to include lines with both leading transparent and all transparent pixels, and added dedicated tests for them.
- Change x86_64 macOS build to use the macos-15 runner
  - There is a new macos-15-intel runner available but we don't actually need an intel runner to build x86_64. (As support for x86_64 winds down, we can probably drop these builds eventually - homebrew is the preferred installation method for macOS anyway)


AlexTMjugador (approved):
Nice! Good catch on a x86_64 macOS runner not being needed to build and run binaries for that target... Now we know those runners have Rosetta 2 in place :smile:

PR #741, opened 2025-11-17
Always remove iDOT

### LINKED ISSUE #740: Correctly handle `iDOT` chunk
It seems the Apple parallel block is being accepted:
https://github.com/w3c/png/issues/530#issuecomment-3293051476

We should make sure Oxipng handles correctly. Probably just remove an invalidated chunk? or.... Possibly modify with correct offsets, if IDATs have not been changed?

andrews05:
Closes #740

### DIFF
--- a/src/headers.rs
+++ b/src/headers.rs
@@ -410,4 +410,16 @@ pub fn postprocess_chunks(aux_chunks: &mut Vec<Chunk>, ihdr: &IhdrData, orig_ihd
             !invalid
         });
     }
+
+    // Remove iDOT which will necessarily be invalid after successful optimization
+    aux_chunks.retain(|c| {
+        let invalid = &c.name == b"iDOT";
+        if invalid {
+            trace!(
+                "Removing {} chunk as it no longer matches the IDAT",
+                std::str::from_utf8(&c.name).unwrap()
+            );
+        }
+        !invalid
+    });
 }


AlexTMjugador (approved):
Nice! :tada:

This PR made me wonder whether we should handle by default unknown chunks marked as unsafe to copy (e.g., those with a capital letter in their fourth byte) by dropping them from the generated PNG. This would more closely follow the [PNG specification](https://www.w3.org/TR/png-3/#table52), but of course, reality doesn't always align with the spec, and a behavior that's technically more correct might cause regressions with unusual chunks found in the wild...

In any case, that's a topic best explored in a separate PR if needed. Thank you!

andrews05:
> reality doesn't always align with the spec

This is very much true 😄
However, I also don't think oxipng classifies as an "editor", so the usual rules about safe-to-copy don't necessarily apply. I note the all of our "display" chunks except `pHYs` are not marked as safe-to-copy, but since we're just re-writing the file rather than "editing" it in any meaningful way, they obviously are safe to copy (and in fact desirable in most cases).
On the other hand, bKGD, sBIT and hIST are true to form and not safe to copy regardless. At the end of the day, it seems like the safe-to-copy flag was well intentioned but not super practical as the correct behaviour is always more nuanced.

This reminds me, I need to update the help to note that iDOT will always be stripped...

TPS:
> but of course, reality doesn't always align with the spec, and a behavior that's technically more correct might cause regressions with unusual chunks found in the wild…

@AlexTMjugador Indeed, the [writeup](https://www.hackerfactor.com/blog/index.php?/archives/895-Connecting-the-iDOTs.html) I [linked in the referring issue](https://github.com/oxipng/oxipng/issues/740#issuecomment-3543732894) mentions that even `iDOT`'s capitalization isn't correct according to the spec.

Issue #523, opened 2023-06-21
"Qualifying round" optimization


Pr0methean:
It would probably improve performance with very little risk of suboptimal compression, if we had the option to discard filtering strategies whose result was e.g. more than 10% larger than the smallest when processed with fast compression, but then use more thorough compression to make the most of a clear winner or resolve a dead heat. This is especially true when processing a lot of images in parallel. The fast compression could also be used to prioritize the most promising filter strategies when under deadline pressure in sequential mode.

Pr0methean:
It's possible this could be implemented as a variant on the fast-evaluation option, where all 10 strategies were evaluated with and without possible reductions. In my app, Brute is best more often than Bigrams, so I'd want to test it when choosing the best reduction.

andrews05:
I'll just note that #509 was recently merged which greatly reduces the losses from fast-evaluation. If you're wanting more speed, I'd definitely recommend using it.

andrews05:
Just to comment further here, this is something I'd like to pursue for a future release (though no promises as to when I'll get to it).

The idea would basically be to combine reduction and filter evaluations into one, with `numReductions * numFilters` fast trials. Then pick some number of top candidates to run full trials of, with thresholds dependent on the `-o` level.

TPS:
https://github.com/shssoichiro/oxipng/pull/640#issuecomment-2278812572 seems to be a similar idea.

PR #742, opened 2025-11-17
Add placeholder options for future development

andrews05:


andrews05:
This relates to #523. I started work on it a few months back and got about halfway but have struggled to find time to complete it. As it's a breaking change, I've been holding off on releasing v10 in the hopes of getting it finished but perhaps we can just add in the required options as placeholders to get v10 out and then complete the development later.

AlexTMjugador:
Hmm, I reckon we could simply release v10 as it is if necessary and add these options once they're actually implemented.

Within reason, I don't believe we should feel any commercial or emotional pressure to delay a major version bump until the bump "feels bigger" like this. Oxipng is mostly used as an "end-user library", meaning it typically doesn't sit deep in transitive dependency trees (from what I've skimmed on crates.io, most dependents are final applications, so ecosystem fragmentation isn't a significant concern when compared to, for example, `serde` or `syn`). We also aim to make breaking changes easy to handle, thanks to thorough documentation and straightforward migration paths, so while downstream consumers won't be able to upgrade by just running `cargo update`, it's not that big of an ask for them to modify the major version in the `Cargo.toml` file and a few other things. What do others think about this?

andrews05:
Hm, yeah that's certainly a fair point. I guess I wasn't thinking so much from a point of view of wanting it to "feel bigger", but more to avoid frequent major version bumps that may require changes to people's code or workflows. But you're right, it's not necessarily a concern that we should worry about.

TPS:
Maybe leave this to a `.5` release that's bit less abrupt than a major version, once it's all implemented?

andrews05:
@TPS That would be fine, _if_ we add the placeholder options in now (though it would probably be a `.1`). Otherwise the new options constitute a breaking change and would require a v11 release.

Realistically, API users shouldn't be constructing the `Options` entirely from scratch - they should be extending the defaults or one of the presets. But the fact that they _can_ construct it from scratch means that any additions are a breaking change.
@AlexTMjugador It might be sensible if we just mark `Options` as non-exhaustive? Then we could freely introduce new options at any time.

[edit] Actually, non-exhaustive doesn't seem work as I was hoping. It prevents you from doing e.g.:
```rs
Options {
    force: true,
    ..Default::default()
};
```
It would be sad to lose this way of constructing the options. (Although there is a [non_exhaustive macro](https://docs.rs/non-exhaustive/latest/non_exhaustive/) that people could use if they really wanted to do that.)

PR #744, opened 2025-11-22
Apply some Clippy lints

Luracasmus:
Let me know if I should add these to the `Cargo.toml`.

All `cargo test` tests pass on the latest x86_64-unknown-linux-gnu nightly.

### DIFF
--- a/src/deflate/mod.rs
+++ b/src/deflate/mod.rs
@@ -1,8 +1,9 @@
 mod deflater;
+use std::{fmt, fmt::Display};
+
 pub use deflater::{crc32, deflate, inflate};
 
 use crate::{PngError, PngResult};
-use std::{fmt, fmt::Display};
 
 #[cfg(feature = "zopfli")]
 mod zopfli_oxipng;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -673,7 +673,7 @@ fn copy_permissions(metadata_input: &Metadata, out_file: &File) -> PngResult<()>
 }
 
 #[cfg(not(feature = "filetime"))]
-fn copy_times(_: &Metadata, _: &Path) -> PngResult<()> {
+const fn copy_times(_: &Metadata, _: &Path) -> PngResult<()> {
     Ok(())
 }
 
--- a/src/sanity_checks.rs
+++ b/src/sanity_checks.rs
@@ -44,7 +44,7 @@ fn load_png_image_from_memory(png_data: &[u8]) -> Result<Vec<RgbaImage>, image::
         decoder
             .apng()?
             .into_frames()
-            .map(|f| f.map(|f| f.into_buffer()))
+            .map(|f| f.map(image::Frame::into_buffer))
             .collect()
     } else {
         DynamicImage::from_decoder(decoder).map(|i| vec![i.into_rgba8()])


andrews05:
Hi @Luracasmus, thanks for your PR.
These changes all seem... fine... but they're also fairly inconsequential, so I'm curious to know why you're wanting to do this?
Note that without adding the rules to cargo, they won't be enforced for any future changes. But if we do add them to cargo, I'd want a good reason (for each rule). So far we've been mostly happy with clippy's defaults (though there are some extras enabled in lib/main).

Luracasmus:
I just went through all the suggestions from the Clippy profile I usually use (mostly pedantic) and applied what felt reasonable, also getting to know the project a bit. A lot of it is essentially pointless but makes the code a bit more consistent and idiomatic.

**missing_const_for_fn:**
This could make the library more versatile and might allow constifying more things in the future. Afaik there's no downside to it. Specifying it as a rule would prevent us from missing things that could be const.

**redundant_clone:**
This could improve performance, but the compiler probably already does this. There is still no reason to clone, though.

**option_if_let_else, manual_let_else, map_unwrap_or:**
These prevent manually implementing standard library functions and purpose-built syntax. The compiled code is probably the same, but it often makes things shorter and imo it looks better. (In some places I also moved duplicated code out of both branches to instead be executed after the convergence, which could be done even without the switch to built-in functions.)

**useless_let_if_seq:**
This is mostly a style change to make the code simpler and more idomatic but in this case it also moved an assignment into just one of the branches. That would probably have been optimized away from the other branch, though.

**if_not_else:**
This might technically reduce the amount of operations being done, but it probably gets compiled to the same thing in practice. Maybe readability could be better without it in some cases?

**redundant_closure_for_method_calls:**
As Clippy puts it, it’s unnecessary to create a closure calling a method when you can just call the method. It's probably inlined and causes no change to the compiled code, though.

**cloned_instead_of_copied**:
There is no reason to clone types which implement `Copy` when `clone` is just implemented by copying.

**unnecessary_semicolon, use_self, ignored_unit_patterns, unseparated_literal_suffix, semicolon_if_nothing_returned:**
These just improve consistency and maybe make the code adhere a bit more to common style practices. They shouldn't affect the library surface or built binary.

andrews05:
Thanks for the info there. May I ask, are you planning on diving deeper into the code and possibly making further contributions? I.e. would these changes be beneficial to you personally as you work oxipng?
I don't mean to sound ungrateful (and certainly no offence intended), I'm just hesitant about accepting personal code style preferences from every passerby who happens to be checking out the code.
Some of these seem potentially good, while others I'm less keen on. Perhaps we could cherry pick some of them. @AlexTMjugador what are your thoughts ?

Luracasmus:
I do want to find more things to contribute since Oxipng is a project that I use a lot personally but I can't guarantee if I will. These changes don't matter much for that though and I split them up so that they would be cherry-pickable. It's perfectly fine by me and understandable if you want to do that or not merge at all. The people who have written the thing should decide how it looks lol, not some random guy

AlexTMjugador:
I think these changes look good as-is.

As stated above, they are not _necessary_ for any functional aspect of the codebase at the moment, but personally I consider disabled-by-default Clippy lints as authoritative and illustrative of the "standard" style of Rust, envisioned by the Rust development team. In my view, the main reason why pedantic lints not pertaining to the restriction category are disabled is due to them being too noisy, not yet properly implemented, or more circumstantial, any of which don't really indicate that it is a bad idea to heed their apparent intent when applicable anyway.

I also found the choice of lints tasteful, considering relatively robust and useful lints only. To name a lint that I'd find more uncomfortable despite being well-intentioned: rustc has a neat `unused_crate_dependencies` lint, which at face value is obviously a good thing to lint about, yet in practice [it can't be used reliably with most Cargo projects](https://github.com/rust-lang/rust-clippy/issues/4341#issuecomment-982989607).

So, let's add these lints to the `Cargo.toml` and merge this :rocket:

AlexTMjugador (approved):

Issue #574, opened 2023-11-13
Support NO_COLOR option for CLI


ace-dent:
Add support for NO_COLOR env var; see https://no-color.org/

TPS:
It seems a bit complex?  IIUC, this is looking to make command-line color text output _always_ be opt-in (via other environment variables or command-line options) than opt-out.

The good news is that there are Rust CLI crates (listed on the site) that already support this, so, if 1 of these is _already_ used, maybe very little additional support is needed other than perhaps the specific supported argument to turn color text on?

andrews05:
This already work for me, have you tried it?

TPS:
@andrews05 That's excellent news indeed!

ace-dent:
@andrews05 test on macOS with `env 'NO_COLOR=1'`. Not working for me.

andrews05:
@ace-dent Note that `env` only applies when executing a command on the same line, as in `env NO_COLOR=1 oxipng ...` (or just `NO_COLOR=1 oxipng ...`).
To retain the var for your current session use `export NO_COLOR=1`.

ace-dent:
@andrews05 Thanks! User error! 😄 
Closed.

PR #745, opened 2025-11-23
Add colors to --help/-h

starsep:
~I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.~
User can disable colors via `NO_COLOR=1` environment variable

Output of `oxipng --help`:
| before | after | `NO_COLOR=1` |
| - | - | - |
| oxipng-before | oxipng-colors | oxipng-no-color |

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -1,9 +1,17 @@
 use std::{num::NonZeroU64, path::PathBuf};
 
+use clap::builder::Styles;
+use clap::builder::styling::{AnsiColor, Effects};
 use clap::{Arg, ArgAction, Command, builder::ArgPredicate, value_parser};
 use parse_size::parse_size;
 
 include!("display_chunks.rs");
+
+const STYLES: Styles = Styles::styled()
+    .header(AnsiColor::Green.on_default().effects(Effects::BOLD))
+    .usage(AnsiColor::Green.on_default().effects(Effects::BOLD))
+    .literal(AnsiColor::Cyan.on_default().effects(Effects::BOLD))
+    .placeholder(AnsiColor::Cyan.on_default());
 
 pub fn build_command() -> Command {
     // Note: clap 'wrap_help' is enabled to automatically wrap lines according to terminal width.
@@ -15,6 +23,7 @@ pub fn build_command() -> Command {
         .version(env!("CARGO_PKG_VERSION"))
         .author("Joshua Holmer <jholmer.in@gmail.com>")
         .about("Losslessly improve compression of PNG files")
+        .styles(STYLES)
         .arg(
             Arg::new("files")
                 .help("File(s) to compress (use '-' for stdin)")


TPS:
> I used [clap-cargo](https://github.com/crate-ci/clap-cargo) as it's easy to use and provides sensible default styles.

@starsep Per https://github.com/oxipng/oxipng/issues/574, `OxiPNG` already has color/`NOCOLOR` support via `anstream`, so maybe doesn't need `clap-cargo` also?  Or maybe replace it, if sensible?  Keeping dependencies minimal & optional is a bit of secondary goal around here, evidentally.

starsep:
Actually `clap-cargo` wasn't used for `NO_COLOR` support. It only provided style. I removed it

`NO_COLOR` seems handled by clap, unfortunately `NOCOLOR` doesn't seem to work.
There is `color` argument for `clap` builder which perhaps could be set based on environment variable

andrews05:
Hi @starsep, thanks for your PR.
As I recall, we used to have coloured output for the help under a previous version of `clap` (v2 or 3), and then clap v4 changed the default output to not be coloured anymore. There's currently an [open discussion](https://github.com/clap-rs/clap/discussions/5881) on the matter where they're reconsidering the default colours and styles.
Your styles here do look pretty sensible, though I wonder if it would be good to wait for clap's guidance here?

starsep:
I was aware of clap changes but not of this discussion, thanks!
From my perspective lack of colors in v4 was a UX regression hence my PR :smile: 

Default `clap` provided style would be nice but I don't consider it a blocker.
Migrating to it once it's released should be easy

andrews05:
@starsep Is your screenshot still accurate after your latest change? I notice the styles you've set aren't identical to [clap_cargo](https://docs.rs/clap-cargo/latest/src/clap_cargo/style.rs.html#10). Did you prefer the non-bright colours?

Btw, if you feel strongly about colours it might be good if you add your feedback to the clap discussion, seeing as they're looking for input.

starsep:
I will add the feedback, I just need to gather my thoughts :smile:.

I have strong feelings about having colors, not so strong about particular values as long as everything is readable.

I wasn't 100% sure whether I updated the image but looks the same.
New one generated using `freeze -x "./target/debug/oxipng --help" --theme monokai --font.size 4 --padding 1` and `oxipng`ed:

517833915-a4ee13d1-5920-4253-acf6-23481ac1c80f

andrews05:
Yeah, I don't have strong feelings either. I'm just wondering if the "bright" colours used in clap-cargo might be better for readability.

TPS:
A stronger contrast is helpful in low-light or -vision (a11y) scenarios, if 1 is already overriding user's TUI setup a bit via color.

AlexTMjugador:
Most terminal themes adjust the ANSI colors requested by programs as needed so that they are somewhat readable anyway when considering the likely surrounding colors. For example, here is the colored output of `cargo --help` when viewed under both an usual dark theme, and a more unusual light, solarized theme:

image

image

Therefore, as Monokai is a reasonably well-calibrated theme, I think chances are that it'll look also fine in most other well-done themes, and the responsibility of ensuring proper contrast is moved to either theme authors or end-users selecting the theme they feel the most comfortable with.

In my view, now that accessibility concerns are set aside, choosing the most beautiful color palette for Oxipng is bound to be an endless debate where everyone will very reasonably have their own preference, so I don't think we should aim to resolve that in this PR. Instead, I think that we could be more pragmatic and consider that _some_ color is arguably better than _no_ color, because the [usage of different colors has been shown to help shape interactions of people with different parts of an UI](https://en.wikipedia.org/wiki/Color_theory#Color_harmony_and_color_schemes), and we want that here to make it easier for people to read the documentation. (After all, most people can't really imagine a code editor that doesn't do any syntax highlighting with different colors, right?)

So, let's merge this! :rocket:

AlexTMjugador (approved):

ace-dent:
@AlexTMjugador @andrews05 @starsep 
Testing the CI [artefact](https://github.com/oxipng/oxipng/actions/runs/19990616685/artifacts/4786125173) under macOS.

Screenshots attached (system dark and light themes). 
My 2¢ – The colorised version is slightly harder to read with the light theme... but then I'm not sure what a better alternative could be!?... 🤷🏼‍♂️

Screenshot 2025-12-06 at 19 35 42

Screenshot 2025-12-06 at 19 35 09


Workaround for macOS users: ` env NO_COLOR=1 oxipng -h `... but then we lose the **bold** formatting for parameters :-(

andrews05:
@AlexTMjugador you're posting screenshots from cargo, but as noted earlier the colours set here aren't identical. Are you sure we don't want to match? If cargo's colours are commonly used, it might be good for consistency.

@ace-dent I'm afraid I don't have much helpful to say on the matter, other than colours are hard. It's true that the colours don't stand out as much on the light theme there but I think it's probably acceptable. People who really struggle with colours will likely have already tailored their theme to work for them.

AlexTMjugador:
> @AlexTMjugador you're posting screenshots from cargo, but as noted earlier the colours set here aren't identical. Are you sure we don't want to match? If cargo's colours are commonly used, it might be good for consistency.

Besides Cargo, I've also tested a random selection of CLI apps I had readily available (Docker, Hyperfine, yt-dlp, vim, psql, oggenc, btop, k9s, nix) in my KDE Breeze terminal theme and the result were basically far from consistent:

- Most apps just didn't bother coloring their `--help` output on an interactive terminal at all.
- `btop` and Hyperfine displayed the same blue accent colors.
- Cargo was the only CLI that displayed green-ish colors as shown above.
- `nix --help` bothered with an admittedly pretty (and unique) mix of red, blue and yellow accents. (And it also piped the output through the `less` pager, making it easier to navigate! Maybe some future Oxipng PR or issue requests this at some point...)

So I don't think there is anything close to a widely-observed standard in what colors CLI apps should use for "manual-like" output, and thus Oxipng picking a somewhat random choice here is basically equivalent to what other apps do.

> It's true that the colours don't stand out as much on the light theme there but I think it's probably acceptable. People who really struggle with colours will likely have already tailored their theme to work for them.

Yeah, the light terminal themes usually come across as worse in contrast, in my opinion. Some people prefer that, though. I'd say that particular themes looking worse is not anything we should try to fix on the Oxipng side of things, because the precise colors used are highly dependent on the precise terminal configuration, and thus asking the terminal to print reasonably different colors is the best we can do (and are already doing): they may end up being completely different anyway.

andrews05:
Well I think _someone_ ought to come up with a standard for everyone to use 😄

TPS:
> Well I think _someone_ ought to come up with a standard for everyone to use 😄

@andrews05 You have successfully incanted… the obligatory xkcd!
Standards

PR #750, opened 2025-12-07
Tiny collection cleanup/optimization

Luracasmus:
See commit descriptions.

~⚠️ adds a very small additional dependency. **arrayvec** is used and trusted by rustdoc, clippy and rustc itself. **tinyvec**, which also provides an `ArrayVec` type, is an alternative, **or just keeping the Vec but still using `extend_from_slice` etc.** I understand if you don't want to add a dep just for this.~

## Testing

All cargo tests pass with default and all features. ~Testing the CLI with some generic unoptimized PNG and APNG images with the default configuration gives bitwise identical results to those from Oxipng 9.1.5.~

~I have not ran any benchmarks, but I would expect performance to be a tiny (probably not measureable unless explicitly targeted in benchmarks) bit better in all cases, unless `ArrayVec::try_extend_from_slice` is somehow slower than what was used before despite most of its checks probably happening at compile time here. An `ArrayVec<u8, 20>` is the same size as an empty `Vec` on a 64-bit system.~

### DIFF
--- a/src/apng.rs
+++ b/src/apng.rs
@@ -1,5 +1,3 @@
-use std::io::Write;
-
 use crate::{
     PngResult,
     error::PngError,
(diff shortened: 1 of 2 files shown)


andrews05:
Apologies for grilling you again, but... convince me why this is good 😄

My understanding of structures like ArrayVec is they're generally only helpful in specific cases and if you don't fall into those cases then there's really no point. In our case, this is not in a hot loop so I would be surprised if a performance difference could be observed even with a targeted benchmark.

Ref: https://www.reddit.com/r/rust/comments/180gnhg/should_you_really_ever_use_arrayvec_or_smallvec/

andrews05:
I just ran a quick bench on the `fctl_data` function.
master: 19ns per iteration
PR: 33ns per iteration
So we're actually seeing a regression here.

We could just do your first commit for `extend_from_slice`? You can probably apply that to `output` as well.

Luracasmus:
> Apologies for grilling you again, but... convince me why this is good 😄

If it's slower then it isn't. My bad ;-; I find it strange though since we're barely storing more data on the stack, but I should have waited to get my PC working to bench this locally. I'll revert it and just do the switch to `extend_from_slice`. Sorry for bothering you with untested microoptimizations.

PR #752, opened 2025-12-09
Use `PngResult` where applicable

Luracasmus:
~This brings `PngError` from 40 to 32 bytes on a 64-bit system, and should only have a small cost in some cases when a `ReadFailed` or `WriteFailed` is created, which should be very rarely.~

Also uses `PngResult` everywhere instead of the type it's defined to. I don't think (?) this should be a breaking change. If it is then maybe it's not worth doing.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -15,9 +15,9 @@ pub enum PngError {
     InvalidData,
     InvalidDepthForType(BitDepth, ColorType),
     NotPNG,
-    ReadFailed(Box<str>, std::io::Error),
+    ReadFailed(String, std::io::Error),
     TruncatedData,
-    WriteFailed(Box<str>, std::io::Error),
+    WriteFailed(String, std::io::Error),
     Other(Box<str>),
 }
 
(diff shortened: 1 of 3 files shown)


andrews05:
Thanks @Luracasmus. The `PngResult` looks fine but the `Box` is surely a breaking change, can we leave this out? (It might actually be better to make it a `Box<Path>`, but again that will need to wait until the next major)

Luracasmus:
Absolutely, I'll revert it. Do you have some kind of system/policy for when major and minor releases happen? `Box<Path>` would make more sense except [when we just write "stdout"](https://github.com/oxipng/oxipng/blob/master/src/lib.rs#L254) (because that's not a valid/special path afaik?), but maybe it's still better. We could also have an enum of path and string or path and a unit variant named `StdOut`

andrews05:
> Do you have some kind of system/policy for when major and minor releases happen?

Not really. We follow semver but there's no schedule as such. Major releases generally happen we have a new feature or notable improvement that requires a breaking change. (E.g. when I finish #749)

> `Box<Path>` would make more sense except [when we just write "stdout"](https://github.com/oxipng/oxipng/blob/master/src/lib.rs#L254) (because that's not a valid/special path afaik?), but maybe it's still better. We could also have an enum of path and string or path and a unit variant named `StdOut`

Oh yeah, I forgot about that. There is the `OutFile` enum but then we won't know the path if it's set to same as input. String will be fine.

PR #758, opened 2025-12-28
Show summary of results

### LINKED ISSUE #678: Summary option to display total percentage and byte reduction
I'm wanting to move away from ECT due to a lack of updates and questionable decisions regarding defaults, however the per-file statistics of oxipng makes for quite a cluttered window and rather hard to see how much of an overall improvement it had. `-q` stops the flood of file paths, but doesn't display anything at all.

I thought I'd ask about a `-S` or `--summary` that would print similar to ECT, allowing for quick comparisons. Something like
```
Processed 107 files
Could not optimize 12 files further
Saved 192.09MB out of 774.25MB (24.81% Smaller)
```

andrews05:
This PR changes oxipng's output to show a summary of results at the end, which is printed to `stdout`.
The default error log level has been lowered from `Info` to `Warn`, meaning `stderr` now prints only errors and warnings. The previous per-file messages are suppressed unless `-v` is specified (`-v` can now be specified up to three times to reach max verbosity).

Example output:
```
Files processed: 289/289   
Input size: 49.9 MiB (52312208 bytes)
Output size: 39.1 MiB (41046190 bytes)
Total saved: 10.7 MiB (21.54%)
(42 files could not be optimized further)
```
When not using verbose mode, the "Files processed" line will be updated after each file processed as a live progress indicator.

Closes #678.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,9 +6,10 @@ use std::num::NonZeroU64;
 use std::{
     ffi::{OsStr, OsString},
     fs::DirBuilder,
-    io::Write,
+    io::{IsTerminal, Write, stdout},
     path::PathBuf,
     process::ExitCode,
+    sync::atomic::{AtomicUsize, Ordering::AcqRel},
     time::Duration,
 };
 
@@ -69,17 +70,28 @@ fn main() -> ExitCode {
         )
     };
 
-    let parallel_files = matches.get_flag("parallel-files");
-    let results: Vec<OptimizationResult> = if parallel_files {
-        files
-            .into_par_iter()
-            .map(|(input, output)| process_file(&input, &output, &opts))
-            .collect()
+    let is_verbose = matches.get_count("verbose") > 0;
+    let print_summary = !matches.get_flag("quiet") && !matches!(out_file, OutFile::StdOut);
+    let print_progress = print_summary && !is_verbose && stdout().is_terminal();
+    let total_files = files.len();
+    let num_processed = AtomicUsize::new(0);
+    if print_progress {
+        print!("Files processed: 0/{}...", total_files);
+        stdout().flush().ok();
+    }
+    let process = |(input, output): (InFile, OutFile)| {
+        let result = process_file(&input, &output, &opts);
+        if print_progress && matches!(result, OptimizationResult::Ok(_, _)) {
+            let value = num_processed.fetch_add(1, AcqRel) + 1;
+            print!("\rFiles processed: {}/{}...", value, total_files);
+            stdout().flush().ok();
+        }
+        result
+    };
+    let results: Vec<OptimizationResult> = if matches.get_flag("parallel-files") {
+        files.into_par_iter().map(process).collect()
     } else {
-        files
-            .into_iter()
-            .map(|(input, output)| process_file(&input, &output, &opts))
-            .collect()
+        files.into_iter().map(process).collect()
     };
 
     // Collect stats
@@ -104,7 +116,7 @@ fn main() -> ExitCode {
     }
 
     // Print summary
-    if !matches.get_flag("quiet") && !matches!(out_file, OutFile::StdOut) {
+    if print_summary {
         let in_bytes = format_bytes(total_in, true);
         let out_bytes = format_bytes(total_out, true);
         let saved = total_in - total_out;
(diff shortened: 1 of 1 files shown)


andrews05:
I moved the stdin determination from `collect_files()` to `main()`. This allows us to:
a. Error out if trying to use stdin with `--dir` (previously it would write a file named "-").
b. Easily determine when stdout is being used.

Thoughts for future BC breaks:
- Remove `OutFile::Path {path:None}` from the API, which represents "same as input". This doesn't really make sense if used in conjunction with `InFile::StdIn` and generally makes it more annoying to work with.
- Allow `--out -` for consistency with how stdin currently works, removing the `--stdout` option.

ace-dent:
@andrews05 - this is a _really_ useful feature. If it's a non-breaking change... might we get a 10.1.0 release?... 🙏

andrews05:
@ace-dent Yeah, we can probably do 10.1 after the json PR lands.

src/main.rs
```diff
@@ -35,57 +36,133 @@ fn main() -> ExitCode {
...
+        .unwrap()
+        .cloned()
+        .flat_map(apply_glob_pattern)
+        .collect();
+    #[cfg(not(windows))]
+    let inputs: Vec<_> = matches
+        .get_many::<PathBuf>("files")
+        .unwrap()
+        .cloned()
+        .collect();
```
AlexTMjugador: While I realize this code has been like this beforehand, now that we're touching it, I think I'd prefer to push the conditional compilation down the fluent call chain as much as possible, like this, so it's less repetitive:

```rust
    let inputs: Vec<_> = matches
        .get_many::<PathBuf>("files")
        .unwrap()
        .cloned()
        .flat_map(|file| {
            #[cfg(windows)]
            {
                apply_glob_pattern(file)
            }
            #[cfg(not(windows))]
            [file]
        })
        .collect();
```

src/main.rs
```diff
@@ -35,57 +36,133 @@ fn main() -> ExitCode {
...
+    let total_files = files.len();
+    let num_processed = AtomicUsize::new(0);
+    if print_progress {
+        print!("Files processed: 0/{}...", total_files);
+        stdout().flush().ok();
+    }
+    let process = |(input, output): (InFile, OutFile)| {
+        let result = process_file(&input, &output, &opts);
+        if print_progress && matches!(result, OptimizationResult::Ok(_, _)) {
+            let value = num_processed.fetch_add(1, SeqCst) + 1;
```
AlexTMjugador: It's probably better to use an `AcqRel` ordering here, it's likely to be more than good enough already.

`SeqCst` is unnecessarily strict for a counter that's updated in the same thread it is read, and the single total modification order guarantee is mostly useful in MPMC situations only. It's likely that even `Relaxed` is mathematically correct here, but given that it provides no meaningful performance difference on common architectures compared to `AcqRel`, unlike `AcqRel` vs. `SeqCst` (the latter of which is more likely to require fencing instructions, even though [it may not even affect codegen in this case](https://rust.godbolt.org/z/rP7W5nTj7)), I'd put the somewhat vague goalpost of balanced safety in that middle point rather than `SeqCst`.

src/main.rs
```diff
@@ -463,3 +537,42 @@ fn process_file(input: &InFile, output: &OutFile, opts: &Options) -> Optimizatio
         }
     }
 }
+
+/// Format byte counts as IEC units to 3 significant figures.
+fn format_bytes(count: isize, include_raw: bool) -> String {
+    const K: isize = 1 << 10;
+    const M: isize = 1 << 20;
+    const G: isize = 1 << 30;
```
AlexTMjugador: I'd feel slightly safer if we used `i32` here rather than `isize`. You never know when someone may want to port Oxipng to a platform with 16-bit pointers...

But jokes aside, I think numeric types with fixed widths are easier to reason about here.

TPS: *Somebody*'s guaranteed going to try to compile this for VAX, a Mars mission, or anything equally out of our collective imaginations.

src/main.rs
AlexTMjugador: Hmm, I do indeed think that's prettier than before, but I still have somewhat of an adversion towards declaring the same variable with different statements for different platforms. Feel free to choose whatever approach you prefer here though, both proposals here are better than before anyway.

A third way could be just renaming and implementing `apply_glob_pattern` for non-Windows platforms too, but doing nothing, as those platforms don't need us to do glob expansion.

AlexTMjugador (approved):
LGTM, thank you for working on this! :rocket:

PR #761, opened 2025-12-29
Add json output option

### LINKED ISSUE #583: Feature request: Machine-readable CLI output
Would be great to have a `--json` flag, or something similar. As it stands, if one wants to integrate Oxipng into another CLI tool (like I'm trying to do at my company), it's not really possible to easily parse the CLI output and figure out what images were optimized vs skipped.

Example output might be something like:

```json
{
  "result": "success",
  "images": [
    { "path": "/path/to/image.png",
      "original_size": 13232,
      "optimized_size": 11323,
    },
    ...
  ]
}
```

Could pipe more and more into these objects over time (like timing data, or what optimizations were applied, etc.)

andrews05:
This PR adds a new `--json` option to output results as JSON.
The JSON is constructed manually to avoid dependency bloat. `serde_json` is just used in tests to validate the output.

Example output (actual output will not be "pretty" formatted):
```json
{
  "results": [
    {
      "input": "dir/image.png",
      "status": "success",
      "output": "dir/optimised.png",
      "insize": 5000,
      "outsize": 3000
    },
    {
      "input": "dir/file.txt",
      "status": "error",
      "error": "Not a png file"
    }
  ]
}
```

Closes #583.

(This builds upon #758 which will need to be merged first)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,46 +38,22 @@ name = "zopfli"
 required-features = ["zopfli"]
 
 [dependencies]
-zopfli = { version = "0.8.3", optional = true, default-features = false, features = ["std", "zlib"] }
-rgb = "0.8.52"
+bitvec = "1.0.1"
+clap = { version = "4.5.53", optional = true, features = ["wrap_help"] }
+crossbeam-channel = { version = "0.5.15", optional = true }
+env_logger = { version = "0.11.8", optional = true, default-features = false, features = ["auto-color"] }
+image = { version = "0.25.9", optional = true, default-features = false, features = ["png"] }
 indexmap = "2.13.0"
 libdeflater = "1.25.0"
 log = "0.4.29"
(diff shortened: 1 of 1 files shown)


Summertime:
Would it be feasible to add information about the compression level (`--zc`) / filter (`-f`) combinations tried (probably not due to heuristic filters and `--fast`), and the winning combination?

andrews05:
@Summertime Wouldn't you already know the options you passed in?
Unfortunately at the point we output the JSON, we don't have access to information such as the winning combo. (Getting access to that would require careful consideration as it would involve breaking changes to the API.)

AlexTMjugador:
> (This builds upon https://github.com/oxipng/oxipng/pull/758 which will need to be merged first)

(Public note to self: let's review this after #758 gets merged, as it is likely to cause merge conflicts or need a rebase that requires manual intervention anyway.)

Issue #757, opened 2025-12-18
oxipng vs optipng


larsenv:
How does this compare to optipng?

TPS:
@larsenv https://github.com/oxipng/oxipng#history is an official answer.

Though, @ 10 years on & v10, maybe it's time for an official retrospective.  What do you think, @shssoichiro, @andrews05, @AlexTMjugador, & other @OxiPNG crew?  Do y'all want to gather such together & perhaps update the above section?

andrews05:
Indeed, while Oxipng may have started out as an OptiPNG clone, today the architecture very different and the similarities are only superficial. OptiPNG is a great program but it's limited in what it can do because it's tied to libpng and zlib.

Comparing the two, the short answer is that Oxipng can optimise better and faster. It has stronger transformations, stronger filtering, and stronger deflate compression. Multi-threading makes Oxipng very fast, but even in single-threaded mode the lower levels will probably still be better and faster than what OptiPNG can do.

ace-dent:
The benefit of OptiPNG (for me), is the access to some very specific zlib [settings](https://optipng.sourceforge.net/optipng-7.9.1.man1.html) (like Window Size). This may be useful for experimentation. In practice, the deflate compression engines used by Oxipng (Zopfli and libedeflate) will nearly always outperform zlib.

Note- @ctruta is quite active moving png projects into the GitHub [pnggroup](https://github.com/pnggroup).

@andrews05 - nice reply +1. Perhaps worth adding some of the text to the document @TPS linked to? 🙏

TPS:
> Note- @ctruta is quite active moving png projects into the GitHub [pnggroup](https://github.com/pnggroup).

@ace-dent I think it's great for software that's essentially abandoned for the PNG Group to take it over (e.g., https://github.com/pnggroup/libpng-see-issue-2/issues/13), but I've some doubt that's relevant to `OxiPNG` for some _long_ time yet.

ace-dent:
@TPS - Sorry, not sure if you are joking?... my note was for general interest; that Truta is still active. Not that this is relevant to Oxipng.

shssoichiro:
[...]

Issue #730, opened 2025-09-11
A few benchmarks and comparisons


demetris:
Hey, all!

I published on my blog some benchmarks I ran recently:

[op111.net - oxipng and friends: A comparison of PNG optimization tools](https://op111.net/posts/2025/09/png-compression-oxipng-optipng-fileoptimizer-cwebp/)

I don’t know if there is anything useful there for the people working on oxipng but I thought I’d share anyway.

The tests were tracked with hyperfine, and all raw data are available in a txt file.

I was curious to see how close oxipng gets to FileOptimizer, so I tested with FileOptimizer too.

Let me know if you have any questions or suggestions. I’d be happy to run additional tests if they can be helpful for the oxipng devs, or format any parts of the results to be included in any part of the oxipng project.

Cheers,

Demetris

ace-dent:
Interesting corpus. Could you create a zip archive with all test files, and make available on your blog? 🙏

TPS:
@demetris Really nice job!  Would you be interested in adding other modern formats, such as `jxl` & `avif` for complete lossless comparision?

Also, if you feel like a followup for _lossy_ formats, iIrc, all of the moderns have lossy modes, too, which would be very interesting compared against the state-of-the-art `jpg` & lossy `png` compressors.

andrews05:
@demetris Nice work! I would definitely recommend using both `--alpha` and `--strip safe` for all files to be more inline with what FileOptimizer will do. Strip makes a difference on e.g. the MediaWiki logo.

demetris:
Thank you for the feedback, all! :-)

@ace-dent That was the initial plan but then I got a bit worried because I don’t know if I am allowed to that. But I will try do something (next week).

@TPS I will do that! At least I want to do it because I am curious.

Only for lossless though. Lossy is difficult. (How do you determine the level to set each tool at for perceptual parity or perceptual transparency? It can be done, but you cannot do it as a single person.)

@andrews05 Today I learned. :-D

I always use FileOptimizer with *Copy metadata* checked, so I thought I was comparing like for like. Looking at the source code of FO, it seems it uses oxipng with `--strip safe` when copying metadata, and with `--strip all` when not copying metadata. Also it always uses `--alpha` for oxipng. Please correct me someone if I am wrong.

I ran all the oxipng tests again with `--strip safe` and I will also add a note about the settings I used, and about `--strip safe`. Thank you for making me look into that! :-)

andrews05:
[...]

PR #767, opened 2026-01-10
Update Readme

andrews05:
Makes some updates to the read me, following comments in #757.

@shssoichiro Let me know if you're okay with the changes to history. There are two things I wanted to do here:
1: Make it clear that Oxipng isn't just a rust version of OptiPNG (i.e. don't assume it will give similar results).
2: Remove any interpretation that there might be any conflict.
Happy to revert some of that if you prefer though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,3 +157,8 @@ Other contributions (such as improving documentation or translations) are also w
 ## License
 
 Oxipng is open-source software, distributed under the MIT license.
+
+## Benchmarks
+
+An independent benchmark conducted by @demetris is linked here with permission:\
+[oxipng and friends: A comparison of PNG optimization tools](https://op111.net/posts/2025/09/png-compression-oxipng-optipng-fileoptimizer-cwebp/)


TPS:
@andrews05 Maybe, rather than outright removing benchmarks, it's possible to move them to a separate file (linked here), & regenerate every release via the build process, using some set of test [A]PNGs?  It really helps differentiate `oxipng` from from other PNG optimizers, especially `optipng`.

andrews05:
@TPS A good benchmark would require a good set of test images that would need to included in the repo here, which is problematic for a few reasons (see #655 for instance). There would be a lot of work involved and it's largely tangential to the basic changes here.
(I removed the benchmark because I don't think it's particularly helpful without the amount saved, and amount saved isn't particularly helpful without a wide image set. Plus, if we're saying oxipng's architecture is quite different to optipng, it doesn't really make sense to directly compare the same levels.)

AlexTMjugador:
> A good benchmark would require a good set of test images that would need to included in the repo here, which is problematic for a few reasons (see #655 for instance). There would be a lot of work involved and it's largely tangential to the basic changes here.

What do you think about linking to an external benchmark, such as https://css-ig.net/benchmark/png-lossless? Since it's external, we avoid the complications of creating and maintaining our own benchmarks, and it's also likely to be less biased (intentionally or not) toward presenting our program in an overly favorable light.

The only courteous thing to do on our end would be to ask the authors whether they're comfortable with us linking to their benchmarks. We could also include a brief disclaimer noting that we don't control the accuracy of such external benchmarks and similar standard caveats.

Edit: there is also issue #730 mentioning further third-party benchmarks.

andrews05:
> Edit: there is also issue #730 mentioning further third-party benchmarks.

Yeah, that might be a good idea. @demetris What do you think about us linking to your post in our read me?

TPS:
> What do you think about linking to an external benchmark, such as https://css-ig.net/benchmark/png-lossless?
> 
> The only courteous thing to do on our end would be to ask the authors whether they're comfortable with us linking to their benchmarks.

I believe this is @ctruta's, who's recently [chimed in](https://github.com/oxipng/oxipng/issues/757#issuecomment-3675976579) on a semi-related topic.

demetris:
@andrews05 Would be happy to have the oxipng README link to my benchmark!

And if there is anything in the post or in the data that folks here feel would make sense to include directly in the README, feel free to do so.

Also, if there are any additions people would like to see in the benchmark, like more data or a few more images to make it more representative, please say so in the issue where I shared the benchmarks (#730) and I will look into it for the next update of the benchmarks.

ace-dent:
*Re. benchmarks*

A user that is seeking fast performance is surely going to conduct their own independent benchmarks- with their own settings and dataset. The inclusion in the project README seems superfluous, IMHO. Linking to other projects and resources adds long-term debt, to guard against future link hijacking, etc.

AlexTMjugador:
> A user that is seeking fast performance is surely going to conduct their own independent benchmarks- with their own settings and dataset. The inclusion in the project README seems superfluous, IMHO. Linking to other projects and resources adds long-term debt, to guard against future link hijacking, etc.

While I agree with the overall sentiment here, and I also run benchmarks myself, I still see value in projects mentioning or conducting benchmarks. They serve as a signal of where others have placed the performance ballpark for a given piece of software, and that information is especially convenient during the initial survey of available alternatives, helping narrow down a set of options worth trying. Besides, historically, linking to external resources has not been a significant maintenance burden, and when links become stale, they're ideal candidates for low-effort (but welcome!) external PRs to update them.

Edit: in light of the comments above, I'd say we could mention @demetris benchmarks in the README then :slightly_smiling_face:

AlexTMjugador:
Alright, so from my side, this PR would be good to merge once we add a mention of, for example, @demetris’ benchmarks to the README.

I'm also adding @shssoichiro as a reviewer, as we'd appreciate his input on the more historical questions raised by @andrews05 in the initial PR message 🙂

andrews05:
> Alright, so from my side, this PR would be good to merge once we add a mention of, for example, @demetris’ benchmarks to the README.

Yup, still planning to get on to that...

andrews05:
Right, finally found a moment to add the benchmark link.

AlexTMjugador (approved):
Alright, since there haven't been any further comments here for a while, I'd say there are no visible concerns about merging this PR, so let's go ahead and merge it! 🚀

PR #769, opened 2026-01-24
v10.1

andrews05:
New release.
(I don't think we need to wait for the read me updates to proceed here)


AlexTMjugador:
Let's do this, thanks 👍

ace-dent:
Thanks @andrews05 and the team! Some really solid features. v10 noticeably helped me save bytes on highly optimised input with `--ziwi` and the recent release makes batch processing far nicer with summary output! Great stuff!!! 🔥

PR #777, opened 2026-02-22
Remove crossbeam-channel dependency

JakubOnderka:
[Since Rust 1.67](https://blog.rust-lang.org/2023/01/26/Rust-1.67.0/#std-sync-mpsc-implementation-updated), `std::sync::mpsc` is based on crossbeam-channel crate, so doesn't make sense to include additional dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -160,15 +160,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "crossbeam-channel"
-version = "0.5.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "82b8f8f868b36967f9606790d1903570de9ceaf870a7bf9fbbd3016d636a2cb2"
-dependencies = [
- "crossbeam-utils",
-]
-
-[[package]]
 name = "crossbeam-deque"
 version = "0.8.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


TPS:
> [Since Rust 1.67](https://blog.rust-lang.org/2023/01/26/Rust-1.67.0/#std-sync-mpsc-implementation-updated), `std::sync::mpsc` is based on crossbeam-channel crate

\>3ya!  Have they kept them synced to each other so there's really no difference?

AlexTMjugador:
> Have they kept them synced to each other so there's really no difference?

Yeah, as far as I can see after some skimming through they have been kept in sync for user-relevant aspects such as performance. You can compare the [commit history of its implementation in `std`](https://github.com/rust-lang/rust/commits/main/library/std/src/sync/mpmc) with the [corresponding history in `crossbeam-channel`](https://github.com/crossbeam-rs/crossbeam/commits/master/crossbeam-channel) to see for yourself.

And the remark in `crossbeam-channel`'s README claiming better performance than `std`'s channel [has not been updated since at least 2018](https://github.com/crossbeam-rs/crossbeam/commit/d9b1e3429450a64b490f68c08bd191417e68f00c#diff-3c737a49994decff92ce83e82171be56cec7724309b2f1f729d3ab1bb4359b39), which is quite a long time ago.

In any case, passing evaluation results across channels is extremely fast compared to the heavy computational workload of the trials themselves. Even if this change introduced a slight performance regression, Amdahl's law suggests it would not be worth worrying about: the measurable benefits of fewer dependencies and a smaller binary size are likely to outweigh any minor regressions, even if they did occur.

So, good catch, and thanks for the PR! 🎉

Issue #286, opened 2020-08-16
Allow specifying a minimum threshold for writing optimized files


aaronfranke:
By default, `oxipng` will only decide to not write a file if the optimized file is equal to (or, presumably in theory, greater than) the size of the original file. This can be forced so that files are always written, but there is currently no way to go in the other direction and specify a minimum compression ratio.

The reason I want this feature is that, if a file can only be compressed by a few bytes, it's not worth doing so for files stored in Git, since the benefit is small and it comes with the downside of increasing the size of the Git repo. I'm only interested in files that can see a significant amount of compression (I've noticed some can be reduced to have the size, for example).

I would like to specify a minimum compression threshold, so that if a file was only able to be compressed by, for example, less than 1 percent, then `oxipng` would not save the compressed file, but if the file was compressed by ex: 2% then it would be saved.

TPS:
Is it possible to denote this in different ways, e.g., %age (`1%`, as above) or just a straight # of bytes (`512`, `1K`, &c)?

aaronfranke:
@TPS Percentage would be ideal, or a ratio (so `0.01` for 1%).

PR #784, opened 2026-02-26
feat: add minimum gain threshold for writes

kyc001:
Split from the previous combined PR so this feature can be reviewed independently.

## Summary
- add `--min-gain <value>` CLI option (accepts percentages and byte sizes)
- thread optional `MinGain` through CLI and optimization paths
- add `optimize_with` / `optimize_from_memory_with` APIs
- treat below-threshold savings as no-change for write behavior
- update docs and tests for min-gain parsing/behavior

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -1,4 +1,10 @@
-use std::{fs, fs::File, io::prelude::*};
+use std::{
+    fs,
+    fs::File,
+    io::prelude::*,
+    path::PathBuf,
+    time::{SystemTime, UNIX_EPOCH},
+};
 
 use oxipng::*;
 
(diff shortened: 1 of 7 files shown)


kyc001:
Addressed the review feedback in 6139e83 by removing the unnecessary `set_readonly(false)` cleanup call in `tests/lib.rs`.

tests/lib.rs
```diff
@@ -87,3 +93,88 @@ fn optimize_srgb_icc() {
...
+fn temp_path(name: &str) -> PathBuf {
+    let nanos = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .expect("clock should be after epoch")
+        .as_nanos();
+    std::env::temp_dir().join(format!("oxipng-{name}-{}-{nanos}.png", std::process::id()))
+}
+
+#[test]
+fn min_gain_bytes_skips_in_place_write() {
```
andrews05: It would probably make more sense to put these in `flags.rs` rather than `lib.rs`. I would also make these a lot simpler:
1. Read the input file.
2. Call `optimize_from_memory` with your options.
3. Make a single assertion based on the insize and outsize of the result (i.e. equal if skipped, less than if not).

andrews05 (changes requested):
Hi @kyc001, thanks for the PR.
I've made a few comments, but I think the overall approach seems sensible.

Out of interest, what's your use-case for this feature?

andrews05:
@kyc001 I'm going to mark this as draft for now, but feel free to pick it up again any time.

vlakoff:
Refs #286.

andrews05:
Closing as stale

PR #785, opened 2026-02-26
Make five small changes to the updated README.md

### LINKED ISSUE #778: A few small things in the updated README
Hey, all! Congrats on the new README!

Just wanted to mention a few small things I noticed. I am not making a PR, since they are not technically wrong stuff (typos etc.), but I will be happy to make a pull request if you ask me to.

Last sentence in the APNG support section:

> prior to running Oxipng

I would use *before* there. Means exactly the same thing, is simpler, is shorter, is fully native English.

> if you are migrating from OptiPNG, please check the help before using.

Unless this is idiomatic English I am not familiar with, there is something missing. I would change that to something like *before using Optipng* or just *before use*.

Last sentence in Usage:

> Similarly, using Zopfli compression (`-z`) is not guaranteed to always be better than without.

I would replace *without* with something like *default compression* or *the default compression algorithm*.

Benchmarks section:

> An independent benchmark conducted by @demetris is linked here with permission:

This is strange. I mean, it is great for Demetris to be mentioned by handle in the oxipng README (visibility is good for lots of things), but I do not like how it looks. No other person is called out by name or handle in the README, not even @shssoichiro.

I would change that to something like:

*An independent benchmark is linked here with permission by the author:*

Last thing:

I think the License section should move lower and become the last section of the README.

Cheers!

demetris:
- CHANGES prior to running TO before running
- CHANGES before using TO before use
- CHANGES better than without TO better than default compression
- CHANGES conducted by @demetris TO with permission by the author
- MOVES License section below Benchmarks, to make it the last section of the README

Fixes #778

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,14 +76,14 @@ All options are case-sensitive.
 \* Note that oxipng is not a brute-force optimizer. This means that while higher optimization levels
 are almost always better or equal to lower levels, this is not guaranteed and it is possible in
 rare circumstances that a lower level may give a marginally smaller output. Similarly, using Zopfli
-compression (`-z`) is not guaranteed to always be better than without.
+compression (`-z`) is not guaranteed to always be better than default compression.
 
 ## APNG support
 
 Oxipng currently only supports limited optimization of animated PNGs (APNGs). It can perform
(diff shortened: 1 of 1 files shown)


ace-dent:
@demetris lgtm!

Issue #790, opened 2026-03-09
Way to force greyscale instead of indexed?


jonnyawsom3:
Some programs only detect greyscale images if the color type is set accordingly, or can do memory reductions by handling it as single-channel.

AFAIK, currently you can only disable transformations to/from Greyscale, but not force it. Meaning some optimized images no longer register as Greyscale, with no easy way to make it into a Greyscale PNG.

Maybe `--ng` could be changed from only `on/off`, to like `-i` with `off/on/keep`? `off` disabling Greyscale (always Indexed), `on` always making it greyscale if possible and `keep` using current behaviour.

jonnyawsom3:
Would probably also solve #704, it's the only related issue I could find.

andrews05:
Some ideas I've been tossing around:
Replace `--ng` with `--grayscale <prefer|avoid|nochange>` (like what you suggested but with different terms).
Add `--prefer <indexed|grayscale>`.
Expand `--np` to not convert to or from indexed, similar to `--ng`.
(I'm using "prefer" rather than "force", as we don't force colour images to grayscale - you'll need imagemagick or similar to do that.)

TPS:
@andrews05: @jonnyawsom3's version of switches seems much more backward-compatible?

jonnyawsom3:
`--ng` with `prefer|avoid|copy` would be simplest, but it wouldn't fit the shorthand for 'No Gray' anymore... Maybe just `--gray`?
But I think expanding `--np` would make sense too, since it was the first thing I tried assuming that creating a new pallete would count as changing one.

ace-dent:
Hmmm.... personally I use ImageMagick and other png optimisers, so generally have control over the whole pipeline... but if you are _only_ using Oxipng, I can see why this might be useful...

My 2¢: Not loving any of the suggestions so far... it's not clear what the default would be set to? 🤷🏼‍♂️

As noted, depending on an individual image, indexed may outperform greyscale (/+alpha) for compression. Currently I use a brute force [script](https://github.com/ace-dent/dex98/blob/4c38c56c2d3bb5efa2f2b4bfdc3d0abbf1dc7580/utils/processimages.sh#L84-L90), to first try with the input color mode, before allowing reductions.

andrews05:
Of the suggestions so far, the only one that isn't a bc break is expanding `--np`. I'd be happy to go ahead with that.

TPS:
[...]

PR #802, opened 2026-03-27
Prevent conversion to indexed when palette reduction is off

andrews05:
~~Fixes 790.~~

@jonnyawsom3 Give the build artefact a test and let me know if this does what you need.

@AlexTMjugador I'm tempted to revert the previous change we made c71fd44 that prevents conversion _from_ grayscale. It's not intuitive that "no grayscale" would work that way and less flexible in the amount of control you have. If you need to prevent conversion from grayscale, you could still use `--nc` or `--np` with this new change. What do you think?

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -298,6 +298,8 @@ The default value depends on the optimization level preset.")
         .arg(
             Arg::new("no-palette-reduction")
                 .help("Do not change color palette")
+                .long_help("\
+Do not convert to indexed and do not modify an existing color palette.")
                 .long("np")
                 .action(ArgAction::SetTrue),
         )
--- a/src/reduction/mod.rs
+++ b/src/reduction/mod.rs
@@ -118,7 +118,7 @@ pub(crate) fn perform_reductions(
     // Attempt to reduce to indexed
     // Keep the existing `png` var in case it is grayscale - we can test both for depth reduction later
     let mut indexed = None;
-    if opts.color_type_reduction && !deadline.passed() {
+    if opts.color_type_reduction && opts.palette_reduction && !deadline.passed() {
         if let Some(reduced) = reduced_to_indexed(&png, opts.grayscale_reduction) {
             // Make sure the palette gets sorted (but don't bother evaluating both results)
             let new = Arc::new(sorted_palette(&reduced).unwrap_or(reduced));


jonnyawsom3:
Due to previous runs already making the images indexed, this doesn't help unfortunately. It seems like Indexed and Grayscale have the same size for my particular files, but because Indexed was chosen first, it won't change to Grayscale without increasing the bitdepth back to fullcolor and then re-optimizing.

andrews05:
Right, I wondered if that might be the case. This may be a sensible change regardless though.

AlexTMjugador:
> I'm tempted to revert the previous change we made [c71fd44](https://github.com/oxipng/oxipng/commit/c71fd44454213469c7712036976bb4ffea152a86) that prevents conversion _from_ grayscale. It's not intuitive that "no grayscale" would work that way and less flexible in the amount of control you have. If you need to prevent conversion from grayscale, you could still use `--nc` or `--np` with this new change. What do you think?

I think preventing conversions from grayscale to other color types can be useful for working around quirks like [this one](https://stackoverflow.com/questions/31312645/java-imageio-grayscale-png-issue). However, if this can already be achieved with `--nc`, then `--nc` seems like the more generally useful option for preventing color type changes people should use, and I don't see a strong reason to overload `--ng` specifically for blocking conversions from grayscale. (Unless people are processing many PNGs at once, some grayscale and others not, in which case they might want to allow RGBA-to-indexed conversions but disallow conversions from grayscale, for example? Even in that scenario it would be more manageable to provide a CLI option for defining a list of allowed color type transformations, with `--nc` acting as a shortcut for that more precise option, though.)

andrews05:
@jonnyawsom3 Apologies, I was just re-reading the issue thread and I think I misread your comment about this change. I thought you were saying the `--np` would work for you, but I realise you just meant it might be good to that _as well_.
So let's go ahead here and keep thinking about how we might offer more control over colour types.

@AlexTMjugador Great comments. Let's take all this into consideration when we think about the bigger picture in #790, as it may impact how we go about it.

TPS:
> a small simple directed graph of allowed color type transformations

What we ancients used to call a flowchart, essentially?

ace-dent:
@AlexTMjugador :
> ... `--nc` seems like the more generally useful option for preventing color type changes ...

+1. I like this generic approach, but have the luxury of feeding in different color types using ImageMagick.

AlexTMjugador:
> What we ancients used to call a flowchart, essentially?

Yes, it's a fancy, math and computer science friendly way of referring to a flowchart :slightly_smiling_face: 

(Trees are also a type of graphs, but don't tell anyone about that.)

PR #804, opened 2026-03-28
Improve bigrams performance

andrews05:
Switch bigrams to use a u8 vec rather than a bit array and avoid reallocating each line.
Bit arrays are memory efficient but less performant. The vec requires 64k memory (8x more than before) but it should be acceptable. It gives a noticeable speed improvement at lower levels.

Bigrams benchmark: 28% faster (now faster than Entropy)
Level 0: 20% faster
Level 1: 10% faster
Level 2: 3% faster

We could do something similar for BigEnt but it would require 4x more memory again and it isn't used as heavily as Bigrams, so real world gains may be negligible.

### DIFF
--- a/src/png/mod.rs
+++ b/src/png/mod.rs
@@ -1,6 +1,5 @@
 use std::{fs, path::Path, sync::Arc};
 
-use bitvec::bitarr;
 use libdeflater::{CompressionLvl, Compressor};
 use log::warn;
 use rustc_hash::FxHashMap;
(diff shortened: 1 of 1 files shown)


TPS:
> We could do something similar for BigEnt but it would require 4x more memory again and it isn't used as heavily as Bigrams, so real world gains may be negligible.

@andrews05 You gave relative performance but not memory №s for Bigrams, & vice versa for BigEnt.  Are you able to give both more specifically for each, especially for worst case?  I usually run OxiPNG on limited hardware, & really want to understand ramifications.  Alternatively, is there a different type that would preserve memory _**&**_ performance?

andrews05:
Bigrams uses a fixed amount of memory. It was previously 8k, with this PR it's 8x more at 64k. It's a pretty small difference in the bigger scheme of things.
I haven't done any performance testing of BigEnt. Bigrams is used for evaluations - it can run multiple times per image and so performance can have a noticeable difference at lower levels. BigEnt is not used for evaluations - it will only run once when enabled so I don't expect the difference to be noticeable, especially as it's only on by default at levels 3 and up.
BigEnt's memory usage is currently variable depending on how many distinct bigrams are present in each line. Changing it to work similarly to this PR would require exactly 256k memory.

TPS:
& these are per-comparison?  So running in || on multi-thread, these would multiply?  That doesn't seem bad  @ all, especially since I expect the pooling takes it into account when allocating.

andrews05:
Yes, multiple threads performing a bigrams filter at the same time would multiply the memory.

PR #812, opened 2026-04-17
Warn when Zopfli --ziwi exceeds --zi

vlakoff:
While manually optimizing PNG files via the CLI, I was using Zopfli and progressively increasing the `--ziwi` value to find the best compression threshold. Since I had already set a specific `--zi` value, I eventually reached a point where `--ziwi` exceeded it without me noticing, making the improvement check redundant.

This PR adds a warning when `--ziwi` is strictly greater than `--zi` to inform the user that the threshold is unreachable. While this specific case happened during manual fine-tuning, this warning will be useful for any user who might inadvertently set an inconsistent Zopfli configuration.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -430,6 +430,16 @@ fn parse_opts_into_struct(
         let iterations_without_improvement = *matches
             .get_one::<NonZeroU64>("iterations-without-improvement")
             .unwrap_or(&NonZeroU64::MAX);
+
+        if iterations_without_improvement > iteration_count
+            && iterations_without_improvement != NonZeroU64::MAX
+        {
+            warn!(
+                "--ziwi ({}) is higher than --zi ({}) and will never be reached.",
+                iterations_without_improvement, iteration_count
+            );
+        }
+
         opts.deflater = Deflater::Zopfli(ZopfliOptions {
             iteration_count,
             iterations_without_improvement,


andrews05:
Thanks @vlakoff, this looks like a sensible change!

The only thing I'd suggest is that we don't really need the test since it's a trivial change that has no impact on operation.

vlakoff:
Fair enough! It's a bit of a shame though, as I was quite happy to provide a test right from the start for once. 😭

I've just removed it to keep the PR focused on the warning logic as suggested. For the sake of posterity, here was the test code:

<details>
<summary>Integration test for the warning message</summary>

```rust
#[test]
#[cfg(feature = "zopfli")]
fn zopfli_ziwi_exceeds_zi_warning() {
    let path = "tests/files/zopfli_ziwi_warn.png";
    let output = Command::new(env!("CARGO_BIN_EXE_oxipng"))
        .args(["--zopfli", "--zi", "10", "--ziwi", "20", path])
        .output()
        .expect("Failed to run executable");
    assert!(output.status.success());
    let stderr = String::from_utf8_lossy(&output.stderr);
    assert!(
        stderr.contains("--ziwi (20) is higher than --zi (10)"),
        "Warning message missing from stderr. Found: {}", stderr
    );
}
```

Note: `zopfli_ziwi_warn.png` was a copy of `zopfli_mode.png`.
</details>

Thanks for the feedback!

vlakoff:
By the way, I should mention that I'm comparing against `NonZeroU64::MAX` to detect if the parameter was not specified by the user. This means the warning won't trigger if a user manually sets that exact value (18,446,744,073,709,551,615).

I figured this is such an edge case that it's safe to ignore. I tried refactoring the logic to handle it differently, but it made the code significantly less readable for no practical gain.

andrews05:
> Fair enough! It's a bit of a shame though, as I was quite happy to provide a test right from the start for once. 😭

Ha, sorry 😄. I just think things might get out of hand if we needed a test for every warning or other message that we emit.

> By the way, I should mention that I'm comparing against `NonZeroU64::MAX` to detect if the parameter was not specified by the user. This means the warning won't trigger if a user manually sets that exact value (18,446,744,073,709,551,615).
> 
> I figured this is such an edge case that it's safe to ignore. I tried refactoring the logic to handle it differently, but it made the code significantly less readable for no practical gain.

Yup, I think it's a perfectly sensible way to handle it.

andrews05 (approved):

andrews05:
@AlexTMjugador It looks like the merge queue is re-running the same tests that the PR has already run. Can we do anything about that?

AlexTMjugador:
> It looks like the merge queue is re-running the same tests that the PR has already run. Can we do anything about that?

I elaborated more about this in https://github.com/oxipng/oxipng/pull/815#pullrequestreview-4157279042, but in short, I don't think there's anything we can do about it, as merge queues are designed to rebase commits onto updated temporary branches.

PR #815, opened 2026-04-21
Improve CI efficiency: Add merge_group support and concurrency limits

### LINKED ISSUE #814: Improve CI workflows for GitHub Merge Queue
@AlexTMjugador It looks like the merge queue is re-running the same tests that the PR has already run. Can we do anything about that?

_Originally posted by @andrews05 in https://github.com/oxipng/oxipng/issues/812#issuecomment-4277550071_

---

To address this, we should refine how our workflows are triggered to better align with GitHub's native Merge Queue feature.

#### 1. Primary Fix: Formalizing the `merge_group` trigger
Currently, our workflows (like `oxipng.yml`) do not explicitly listen for the `merge_group` event. As a result, when the Merge Queue creates a temporary merge commit, GitHub Actions uses the generic `push` event as a **fallback** to trigger the CI.

**The Proposal:**
Explicitly add the `merge_group` event to the `on:` section of our workflows and restrict the `push` event to the `master` branch.

* **Why it matters:** Adding `merge_group` allows GitHub to explicitly identify these runs as part of the queue logic rather than standard branch activity. This ensures that required status checks are correctly mapped to the Merge Queue context, preventing redundant triggers and ensuring the CI behaves predictably within the modern GitHub workflow.
* **Clarification for contributors:** We still run the full test matrix for the final validation, but this change provides the clarity needed for GitHub to manage the transition from PR to "Merged" without overlapping unnecessary runs.

#### 2. Additional Improvement: Concurrency Control
As a separate improvement, we could implement a `concurrency` policy with `cancel-in-progress: true`.

* **What it does:** If a PR is updated (e.g., a new commit or a force-push) while a previous CI run is still processing, GitHub will immediately cancel the old, now-obsolete run to start the new one.
* **The Benefit:** This is purely to save runner minutes and clear the "Actions" tab of builds that no longer matter. It ensures that for any given PR or Merge Group attempt, only the most recent version is consuming resources.

---

#### Proposed Implementation (Example for `oxipng.yml`)

```yaml
name: oxipng

on:
  push:
    branches: [ master ] # Restrict push to the main branch
  pull_request:
    types: [ opened, synchronize ]
  merge_group: # The proper way to trigger CI for the Merge Queue
  workflow_dispatch:

# Bonus: Prevent wasting resources on obsolete runs
concurrency:
  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
  cancel-in-progress: true
```

vlakoff:
This PR addresses #814 by optimizing the GitHub Actions workflows for the native Merge Queue feature and streamlining event triggers.

### Changes

* **Merge Queue Support**: Added the `merge_group` trigger to `oxipng.yml` and `docker.yml` to support the native GitHub queue logic.
* **Push Restriction**: Restricted the `push` trigger to the `master` branch and version tags. This avoids duplicate runs when a Pull Request is already open.
* **Concurrency Control**: Added `concurrency` groups with `cancel-in-progress: true` to all workflows. Obsolete runs will now be automatically cancelled when a new commit is pushed.
* **Simplified PR Triggers**: Removed explicit `types` from `pull_request` events to include the default `reopened` trigger.

### Benefits

* Ensures smooth integration with the GitHub Merge Queue.
* Removes redundant "Push" runs that overlap with "Pull Request" events.
* Automatically cancels stale builds, providing faster feedback on the latest commits.

### Note for Contributors

**CI runs are no longer automatically triggered on simple pushes to non-master branches.**

* The test suite will trigger automatically once a Pull Request is opened (including Draft PRs).
* Manual runs can still be initiated via the **workflow_dispatch** button in the Actions tab for any branch.

**Note:** I have since reverted the **Push Restriction** based on the discussion below. CI will still trigger on all branch pushes to allow testing before a PR is opened. While this means some redundancy remains, the new **concurrency control** will still handle canceling any stale builds.

Fixes #814

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -7,6 +7,8 @@ on:
       - 'v*.*.*'
   pull_request:
   merge_group:
+    types:
+      - checks_requested
   workflow_dispatch:
     inputs:
       use_cache:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -36,7 +36,7 @@ jobs:
       context: . # Use local checkout to support GitHub Merge Queue
       output: image
       platforms: linux/amd64,linux/arm64
-      push: ${{ github.event_name != 'pull_request' && github.event_name != 'merge_group' }}
+      push: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
       cache: ${{ github.event_name != 'workflow_dispatch' || inputs.use_cache }}
       # We don't sign the image because ghcr.io doesn't support OCI Referrers API yet, and the fallback relies on pushing new tags on each build
       # https://github.com/docker/github-builder/issues/109#issuecomment-3885082486
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,14 +18,18 @@ on:
         default: true
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}
+  group: >-
+    ${{ github.workflow }}-
+    ${{ github.event.pull_request.number ||
+        (github.event_name == 'merge_group' && format('mq-{0}', github.event.merge_group.head_ref)) ||
+        github.ref }}
   cancel-in-progress: true
 
 jobs:
   build:
     uses: docker/github-builder/.github/workflows/build.yml@v1
-    # Exclude 'push' events from the Merge Queue to avoid a race condition where
-    # the 'push' event might cancel the 'merge_group' event via the concurrency group.
+    # Exclude 'push' events on Merge Queue branches, as they are already
+    # handled by the 'merge_group' event.
     if: (!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))
     permissions:
       contents: read
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,9 +6,7 @@ on:
     tags:
       - 'v*.*.*'
   pull_request:
-    types:
-      - opened
-      - synchronize
+  merge_group:
   workflow_dispatch:
     inputs:
       use_cache:
@@ -16,6 +14,10 @@ on:
         required: true
         type: boolean
         default: true
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   build:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,7 +18,7 @@ on:
         default: true
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -24,6 +24,9 @@ concurrency:
 jobs:
   build:
     uses: docker/github-builder/.github/workflows/build.yml@v1
+    # Exclude 'push' events from the Merge Queue to avoid a race condition where
+    # the 'push' event might cancel the 'merge_group' event via the concurrency group.
+    if: (!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))
     permissions:
       contents: read
       packages: write
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -18,11 +18,7 @@ on:
         default: true
 
 concurrency:
-  group: >-
-    ${{ github.workflow }}-
-    ${{ github.event.pull_request.number ||
-        (github.event_name == 'merge_group' && format('mq-{0}', github.event.merge_group.head_ref)) ||
-        github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -30,6 +30,7 @@ jobs:
       id-token: write
       attestations: write
     with:
+      context: . # Use local checkout to support GitHub Merge Queue
       output: image
       platforms: linux/amd64,linux/arm64
       push: ${{ github.event_name != 'pull_request' }}
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -36,7 +36,7 @@ jobs:
       context: . # Use local checkout to support GitHub Merge Queue
       output: image
       platforms: linux/amd64,linux/arm64
-      push: ${{ github.event_name != 'pull_request' }}
+      push: ${{ github.event_name != 'pull_request' && github.event_name != 'merge_group' }}
       cache: ${{ github.event_name != 'workflow_dispatch' || inputs.use_cache }}
       # We don't sign the image because ghcr.io doesn't support OCI Referrers API yet, and the fallback relies on pushing new tags on each build
       # https://github.com/docker/github-builder/issues/109#issuecomment-3885082486
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -9,11 +9,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: >-
-    ${{ github.workflow }}-
-    ${{ github.event.pull_request.number ||
-        (github.event_name == 'merge_group' && format('mq-{0}', github.event.merge_group.head_ref)) ||
-        github.ref }}
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -19,8 +19,14 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 60
 
-    # Prevent tags and in-repo PRs from triggering this workflow more than once for a commit
-    if: github.ref_type != 'tag' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork)
+    # Prevent tags and in-repo PRs from triggering this workflow more than once for a commit.
+    # We also exclude 'push' events from the Merge Queue (gh-readonly-queue/) because they
+    # trigger alongside the 'merge_group' event. Without this, a race condition in the
+    # concurrency group could cause the 'push' event to cancel the 'merge_group' event.
+    if: >-
+      github.ref_type != 'tag' &&
+      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork) &&
+      (!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))
 
     strategy:
       fail-fast: false
@@ -154,8 +160,14 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
-    # Prevent tags and in-repo PRs from triggering this workflow more than once for a commit
-    if: github.ref_type != 'tag' && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork)
+    # Prevent tags and in-repo PRs from triggering this workflow more than once for a commit.
+    # We also exclude 'push' events from the Merge Queue (gh-readonly-queue/) because they
+    # trigger alongside the 'merge_group' event. Without this, a race condition in the
+    # concurrency group could cause the 'push' event to cancel the 'merge_group' event.
+    if: >-
+      github.ref_type != 'tag' &&
+      (github.event_name != 'pull_request' || github.event.pull_request.head.repo.fork) &&
+      (!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))
 
     steps:
       - name: Checkout source
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -3,10 +3,12 @@ name: oxipng
 on:
   push:
   pull_request:
-    types:
-      - opened
-      - synchronize
+  merge_group:
   workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
 
 jobs:
   ci:
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -4,6 +4,8 @@ on:
   push:
   pull_request:
   merge_group:
+    types:
+      - checks_requested
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -9,7 +9,11 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}
+  group: >-
+    ${{ github.workflow }}-
+    ${{ github.event.pull_request.number ||
+        (github.event_name == 'merge_group' && format('mq-{0}', github.event.merge_group.head_ref)) ||
+        github.ref }}
   cancel-in-progress: true
 
 jobs:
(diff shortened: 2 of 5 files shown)


vlakoff:
I've added a commit to refine the `merge_group` trigger logic.

According to the [GitHub Actions documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#merge_group), the `merge_group` event triggers on all activity types by default. This may include the `destroyed` action, which occurs when a PR is removed from the queue or a group is invalidated.

As detailed in the [GitHub Webhooks documentation](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=checks_requested#merge_group), receiving a `destroyed` event should ideally be used to cancel existing checks, not start new ones. Without specifying `types: [checks_requested]`, the CI could start redundant runs for groups that are no longer relevant.

**Benefits of this change:**

* **Efficiency:** Prevents wasting CI minutes on "zombie" runs when the queue is reorganized or a PR is removed.
* **Precision:** Ensures the test suite only kicks in when a merge group actually requests status checks.
* **Future-proofing:** Aligns the workflows with GitHub's explicit recommendations for CI efficiency.

.github/workflows/docker.yml
```diff
@@ -6,9 +6,9 @@ on:
     tags:
       - 'v*.*.*'
   pull_request:
```
shssoichiro, line 8: Is this empty key valid? This seems likely to be a mistake.

.github/workflows/oxipng.yml
```diff
@@ -2,12 +2,18 @@ name: oxipng
 
 on:
   push:
+    branches:
+      - master
   pull_request:
```
shssoichiro, line 5: Same empty key here.

shssoichiro:
Clarifying question: Does this still automatically rerun the pipeline for a pull request when a user pushes changes to it? Assuming yes, I think these are good improvements.

.github/workflows/docker.yml
vlakoff: This syntax is actually valid; leaving the key empty simply enables the trigger with its default configuration. We could use `pull_request: {}` to be more explicit, but it's functionally identical and I opted for brevity.

vlakoff:
Yes, it will! Even with the `push` trigger restricted to `master`, the `pull_request` trigger includes the `synchronize` event by default. This event specifically triggers a new run every time a user pushes new commits to an existing PR.

By removing the global `push` trigger, we avoid running the same pipeline twice (once for the branch push and once for the PR synchronization). This saves CI minutes and prevents having duplicate entries for the same commit in the Actions history.

andrews05:
Thanks @vlakoff! The changes look pretty sensible to me. I’ll leave it to @AlexTMjugador  to sign off on it though.

AlexTMjugador (changes requested):
> Yes, it will! Even with the `push` trigger restricted to `master`, the `pull_request` trigger includes the `synchronize` event by default. This event triggers a new run every time a user pushes new commits to an existing PR.

We do want pushes to non-`master` branches to trigger CI, though. Since pushing directly to `master` is blocked when CI checks do not pass for a commit, I often push changes to another temporary branch, waiting for checks to pass there, and then merge the commit back into `master`.

Moreover, I'm not convinced these changes achieve this stated benefit:

> - Removes redundant "Push" runs that overlap with "Pull Request" events.

As the GitHub documentation you referenced explains, there are multiple CI triggers relevant during a PR merge:

- `merge_group`, which fires when a PR is added to the merge queue.
- `push`, which also fires when a PR is added to the merge queue, since the merge queue creates a temporary branch from the mainline with the PR commits to ensure CI runs against the latest code.

So CI still needs to pass again in both cases, and there doesn't seem to be a general way to reuse a successful CI run from the PR branch. In fact, the commit hash for [the PR-triggered CI in one of your previous PRs](https://github.com/oxipng/oxipng/actions/runs/24641270123) (`71fb5346`) differs from that of the [merge-queue push-triggered CI](https://github.com/oxipng/oxipng/actions/runs/24646089624) (`e167b04`), due to the rebasing done by the queue to test against an updated base.

> - **Simplified PR Triggers**: Removed explicit types from pull_request events to include the default reopened trigger.

Why should CI run when a PR is reopened? Reopening a PR doesn't change its branch, so it's unclear to me what value there is in triggering CI again.

That said, I see more merit in your changes to the `concurrency` setting, as avoiding CI runs for stale commits makes intuitive sense. Still, it's worth double-checking that the chosen concurrency group key works well across all relevant events, not just PRs (for example, `push` events also fire for tags).

Did you verify that your changes produce the intended effects?

.github/workflows/docker.yml
AlexTMjugador: Indeed, [this is valid YAML syntax](https://stackoverflow.com/a/64462925/9366153), and GitHub replaces an undefined value with a default.

.github/workflows/oxipng.yml
AlexTMjugador: See https://github.com/oxipng/oxipng/pull/815#discussion_r3126320736 :slightly_smiling_face:

vlakoff:
Thank you for the feedback. I've updated the PR to address your points:

* **Branch Pushes:** I understand your workflow of testing on temporary branches before opening a PR. I have reverted the restriction, so CI will now trigger on all pushes. While this means the redundancy between `push` and `pull_request` triggers will persist, it ensures that you can still validate code on any branch at any time.
* **The "reopened" Trigger:** I've kept this trigger because the **GitHub merge commit** (the temporary ref CI actually tests) is a snapshot of the PR merged into the _current_ `master`. If a PR is reopened after a period of inactivity, the previous status check is technically testing against an obsolete base. Triggering on `reopened` ensures the code still integrates correctly with the latest mainline.

vlakoff:
Regarding the redundant runs, I’ve looked into ways to further consolidate the `push` and `pull_request` triggers using concurrency groups, but it seems we’ve reached the limit of what can be done without compromising the workflow.

Using a shared key (like `${{ github.head_ref || github.ref }}`) would eliminate the redundancy, but it risks canceling the 'virtual merge commit' test in favor of a simple branch test (or vice-versa), which isn't ideal for proper PR validation since they test different contexts.

So, while the current concurrency setup mainly helps by canceling _stale_ runs from multiple rapid pushes, it's likely the best middle ground for avoiding redundant work without making the CI logic too opaque or risky. The initial redundancy when a PR is opened is a 'necessary evil' to guarantee both branch stability and integration integrity.

vlakoff:
Regarding the concurrency group, a solution could potentially be:
`${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}`

This would make the logic more explicit for the Merge Queue context.

vlakoff:
I've pushed a commit to refine the concurrency group logic.

By using `${{ github.workflow }}-${{ github.event.pull_request.number || github.event.merge_group.head_ref || github.ref }}`, we ensure better precision for the Merge Queue.

**Why this change?**

* **For Pull Requests:** It continues to use the PR number, correctly cancelling stale runs when new commits are pushed.
* **For Merge Queue:** Using `merge_group.head_ref` provides a stable identifier for the queue's temporary branches. This prevents "zombie" runs if the queue is reordered or invalidated, as GitHub will now correctly identify and cancel obsolete checks within the queue context.
* **For Master/Tags:** It safely falls back to `github.ref` as before.

This makes the CI more robust and avoids wasting resources on redundant or outdated merge group evaluations.

vlakoff:
Regarding the switch to default `pull_request` triggers, which now includes `reopened`: I'm not entirely sure this will always achieve the goal of testing against an updated `master`. GitHub often reuses the previous virtual merge commit (`refs/pull/X/merge`) when a PR is reopened, unless a new commit is pushed or a manual update is triggered. At worst, though, it would simply result in a superfluous run.

However, this is largely mitigated by the addition of the **Merge Queue** (`merge_group` trigger). This is the proper way to ensure the code is validated against the latest `master` before merging, regardless of how long the PR has been open.

vlakoff:
Regarding the `reopened` trigger: while the virtual merge commit isn't always refreshed by GitHub upon reopening, using the default triggers fits the intended architecture:

* **Role of Merge Queue:** The `merge_group` trigger handles the final validation against the latest `master`. This allows the PR-level CI to serve as an early feedback loop rather than the definitive integration gate.
* **Concurrency management:** The `cancel-in-progress: true` setting minimizes the impact of redundant runs. Any unnecessary build is automatically terminated if a new commit is pushed.
* **Standardization:** Using default triggers follows standard GitHub Actions behavior. In projects without a Merge Queue, stale PRs are typically addressed through "Require branches to be up to date" protections to ensure code correctness before merging.

Given the rarity of `reopened` events, the potential for an updated status check likely outweighs the overhead of an occasional redundant run. The combination of simplified triggers, Merge Queue support, and concurrency control provides a balanced setup for the repository's needs.

vlakoff:
I have updated the PR to restore CI triggers on all branch pushes as @AlexTMjugador requested.

I've kept my other changes (concurrency, `merge_group` support, and default PR triggers) as they remain relevant.

AlexTMjugador (approved):
Thank you for addressing the feedback.

I'm still not entirely convinced that triggering CI on both `push` and `merge_group` events for temporary merge queue branches is the right approach. The strongly worded guidance against relying on `push` triggers without clearly explaining why on the GitHub docs puts me off somewhat. That said, we can better monitor how it behaves after merging and adjust as needed.

vlakoff:
### Explanation of CI Failure

The Docker build failed during the Merge Queue run because the `docker/github-builder` was attempting to fetch the source code using the temporary Git ref `refs/heads/gh-readonly-queue/...` directly from the remote repository. These internal references are often not yet visible or accessible via a standard remote fetch.

### The Solution

I have updated `docker.yml` to explicitly set `context: .`. This forces the builder to use the local files already checked out by the GitHub runner in the previous step, rather than trying to clone the temporary branch from the internet. This is the recommended approach for supporting GitHub Merge Queues in containerized workflows.

vlakoff:
Following the previous change to use the local context for Docker builds (to support the Merge Queue), I have updated `.dockerignore` and `.gitignore` with the following improvements:

* **Build Performance:** Since we now use `context: .`, the runner sends the local workspace to the Docker daemon. I've excluded `/.git/` (~54MB) and `/target/` to ensure this transfer is near-instantaneous and doesn't leak history into image layers.
* **Precision & Safety:** I have adopted a more explicit syntax using leading slashes (`/`) to anchor paths to the root and trailing slashes (`/`) to specifically target directories. This prevents accidental exclusion of files with similar names in sub-directories.
* **Organization:** Entries are now sorted alphabetically for better readability and easier maintenance.

vlakoff:
> I'm still not entirely convinced that triggering CI on both `push` and `merge_group` events for temporary merge queue branches is the right approach. The strongly worded guidance against relying on `push` triggers without clearly explaining why on the GitHub docs puts me off somewhat. That said, we can better monitor how it behaves after merging and adjust as needed.

A solution might be to add some condition, like:

```
if: github.event_name != 'push' || !startsWith(github.ref, 'refs/heads/gh-readonly-queue/')
```

"Run this job for every type of event, except when a `push` happens on a branch created by the Merge Queue."

vlakoff:
### Fixing potential race conditions and event collisions in Merge Queue

I pushed a commit to refine the CI triggers and ensure stability when using the native Merge Queue.

#### The Issue:

When a PR enters the merge queue, GitHub triggers both a `merge_group` event and a `push` event. Even with a refined concurrency group, these two distinct events might resolve to the same concurrency key because they both reference the same temporary `gh-readonly-queue/` branch.

#### The Race Condition:

Since these events fire simultaneously, they create a race condition. If the redundant `push` job resolves to the same key and starts just after the `merge_group` job, the `cancel-in-progress: true` policy might unexpectedly terminate the official `merge_group` run to favor the `push` run. This can lead to lost context and inconsistent CI results.

#### The Solution:

I have added an explicit filter: `(!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))`.

* **Reliability:** It guarantees that only the `merge_group` event is processed for queue branches, protecting it from being canceled by a duplicate `push` event.
* **Efficiency:** It prevents the CI from wasting resources by not spawning two identical environments for the same commit.
* **Maintainability:** I reformatted the `if` condition in `oxipng.yml` using the `>-` (folded block scalar) syntax to make the logic clearer and easier to update in the future.

vlakoff:
First of all, I apologize for the verbosity of my comments in this PR.

I wanted to explain the **domino effect** behind these changes. My goal was to provide official support for the `merge_group` event while improving efficiency with `concurrency` groups. However, combining these two features introduced a potential technical conflict:

* **The Duplicate Run:** When a PR enters the queue, GitHub triggers both a `merge_group` event and a `push` event for the same temporary branch.
* **The Race Condition:** Since both events might resolve to the same `concurrency` key with `cancel-in-progress: true`, they compete for a single slot. This creates a race condition where one event could prematurely cancel the other. If the official `merge_group` run happens to be the one canceled, it breaks the merge queue's logic.
* **The Solution:** The `if` conditions I added specifically filter out these redundant `push` events on `gh-readonly-queue/` branches. This ensures that only the official `merge_group` job is processed, guaranteeing CI stability.

With the addition of the local Docker context to support temporary refs, the workflow should now be robust.

vlakoff:
I'm concerned that `cancel-in-progress: true` will still kill the `merge_group` run. GitHub evaluates concurrency at the workflow level **before** the job-level `if` conditions.

If a `push` event is registered after a `merge_group` event, GitHub will kill the active CI and then skip the new one because of the `if`, leaving the queue with no results.

While adding `${{ github.event_name }}` to the concurrency key would isolate and protect the Merge Queue, it breaks efficiency for standard PR development. When a user pushes a commit to an open PR, GitHub fires both a `push` and a `pull_request (synchronize)` event. If we separate them by event name in the concurrency group, they will no longer cancel each other, and we'll end up running the full CI twice for every push.

vlakoff:
Pushed yet another commit. To be honest, this is starting to get a bit tedious; the logic is fine, the only real pain point is how concurrency groups behave.

I've updated the keys to properly isolate `merge_group` events using a prefix. This ensures that the internal `push` events triggered by the queue can no longer cancel the actual `merge_group` runs.

vlakoff:
It's a mess. And the official documentation is actually spurious, which really doesn't help.

I can't even find out if  `push` events are triggered on merge queue updates.

In doubt, let's keep the condition, which can't harm:
```
if: (!(github.event_name == 'push' && startsWith(github.ref, 'refs/heads/gh-readonly-queue/')))
```

What really p***** me off are actually the concurrency groups. I'm close to suggesting ditching them. Or something including `github.event_name` for strong separation, but potentially preventing less duplicates. e.g.:
```
${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || ... }}
```

Some refs:
* https://github.com/orgs/community/discussions/69342
* https://github.com/go-vela/community/issues/1061
* https://github.com/orgs/community/discussions/46757

vlakoff:
The concurrency group keys in `oxipng.yml` and `docker.yml` were unnecessarily complex, using a chain of fallbacks (`pull_request.number || merge_group.head_ref || github.ref`) to distinguish between event types.

By including `github.event_name` in the key, `github.ref` alone is sufficient as the second component — each event type already produces a naturally distinct ref:

* `push` on a branch → `refs/heads/feature-x`
* `pull_request` → `refs/pull/42/merge`
* `merge_group` → `refs/heads/gh-readonly-queue/...`

This also cleanly prevents the race condition between the `merge_group` event and the phantom `push `event that GitHub fires on the same temporary queue branch: they now resolve to structurally different keys and cannot cancel each other, without needing the `mq-` prefix workaround.

The `if` condition filtering `push` events on `gh-readonly-queue/` branches is unrelated to concurrency management — it simply prevents a pointless job from being queued at all for those events.

vlakoff:
I know I’ve said this before, but we should finally be good to go now.

To summarize the changes:

* **Merge Group Support:** I added the `merge_group` event and handled the necessary side effects:
    * Added a condition to skip redundant runs on `gh-readonly-queue/` branches. I’m not 100% certain it's strictly required by GitHub's internal logic, but it's safer to have it to prevent duplicate workflows.
    * Switched Docker to a "local context" and updated `.dockerignore` to optimize transfer speeds and build reliability.
    * Excluded the Merge Queue from pushing Docker images, moving to a "whitelist" approach for better control and reliability.
* **Concurrency Groups:** Added as a "bonus" for better resource management. I designed them to avoid any "unexpected key overlap," making the logic simple and safe. This is particularly useful when pushing multiple updates to the same PR.
* **Standardized PR Events:** Now using default types for `pull_request`. This adds the `reopened` type; it’s more standard and favors reliability—better to have a redundant run than to miss a necessary one.

AlexTMjugador (commented):
Thank you again for your PR! However, I'm afraid it got somewhat out of hand after iterating on it.

To move forward with it, I've cherry-picked elements from your PR that I agree are net improvements, such as the addition of concurrency groups, ignore file tweaks, and replacing a usage of `${{ env.GITHUB_SHA }}` with `${{ github.sha }}`. I've also added you as a co-author of the cherry-pick commit so that you also get some credit for it :slightly_smiling_face: 

Most notably, I didn't carry over the `merge_group` changes because I still don't see any benefit in them and, perhaps most importantly, they introduce additional maintenance burden: as you've discovered, GitHub's documentation on them is not great, the merge queue works well with the `push` event anyway, and there is no realistic chance of either of those changing anytime soon.

vlakoff:
Thank you @AlexTMjugador for taking the time to review this PR and cherry‑pick the useful parts into commit 4e80dd0.

I'll admit this PR went through a lot of iterations — perhaps more than it should have! The `merge_group` support turned out to introduce a cascade of side effects (race conditions, Docker context issues, concurrency key collisions) that required one fix after another. That complexity is ultimately what made the PR hard to review and confidently merge as a whole.

That said, I'm glad the process allowed us to surface some genuinely useful improvements: the concurrency groups, the `github.sha` fix, and the ignore file cleanups. And looking at the concurrency keys you've chosen (especially the SHA‑based approach for `oxipng.yml`), I think your implementation is actually cleaner and more robust than what I had landed on — so the end result feels very solid to me.

Thanks again for the co‑authorship credit, and for pushing this forward!

----

One small thing that might still be worth considering, and which is actually independent of the `merge_group` topic: the whitelist approach for the Docker image push condition (commits ccbd2f9, 09cb2da). Switching from:

```yaml
push: ${{ github.event_name != 'pull_request' }}
```

to an explicit allowlist:

```yaml
push: ${{ github.event_name == 'push' || github.event_name == 'workflow_dispatch' }}
```

is a more defensive practice — any future trigger added to the workflow would default to *not* pushing an image, which is the safe behavior. The current exclusion‑based condition is functionally equivalent today, but less robust as the workflow evolves.

PR #820, opened 2026-06-03
Improve palette sorting

andrews05:
Changes:
1. Implemented a new variation of the Zeng algorithm ("ezeng") to be used instead of the "modified" version.
2. Implemented a further refinement of the result using pairwise swaps.
3. Retired the Battiato method which is now just noise in comparison with the new method (it actually hurts some images due to evaluation mis-predictions).

Some results at o4:

| | Master Savings | PR Savings | Master Time | PR Time |
|-|-|-|-|-|
| Set 1 | 2.23 MiB (9.01%) | 2.51 MiB (10.15%) | 12.8s | 12.7s |
| Set 2 | 4.65 MiB (15.82%) | 4.98 MiB (16.93%) | 15.0s | 15.3s |

Ezeng is a more expensive but notably more effective version, developed with some accidental AI assistance while trying to implement the original zeng method. Despite being slower it is still fast enough to have negligible impact on total runtime. However, the pairwise swaps are even slower still and this does impact runtime a little.

```
test reductions_palette_sort_mzeng                  ... bench:   1,635,055.23 ns/iter (+/- 126,507.20)
test reductions_palette_sort_ezeng                  ... bench:   7,996,658.30 ns/iter (+/- 86,015.49)
test reductions_palette_sort_ezeng_with_swaps       ... bench:  47,197,983.20 ns/iter (+/- 339,817.45)
```


TPS:
Is it reasonable @ all to keep _all_ of them, @ least 'til _next_ major version, & perhaps have a flag that'd allow to choose/block certain sorts  for those who want that?

As wide as your dataset is, @andrews05, it's unlikely that it'd cover everything we encounter out here, & this seems like a potentially large breaking change.

ace-dent:
@andrews05 - exciting!
Did you preserve the heuristic I suggested- that palette position 0 should be reserved for the most frequently occurring horizontal-edge colour (i.e. common row start and end colour)? 
In your testing, do you count number of wins -vs- loss on a per file basis, rather than just net improvement? This might expose regressions. I wonder if the new approach is universally better?

ace-dent:
> ... Retired the Battiato method ... (it actually hurts some images due to evaluation mis-predictions)...

Is that due to `fast` evaluation of the palette variations? So we get stuck in a local minima early? I have wondered if the palette variations are within a threshold (~5% ?), we should consider no clear winner and continue to test into the next stage?

TPS:
While on the subject, I believe there's a switch to skip all this & just use whatever palette sort that already exists.  Is _that_ sensible to include as a null/non-sort _by default_ in this mix?  I think it'd be fascinating to see how a previous tool's heuristic (e.g., `pngquant` or `rdopng`) stacks up against the 1s above in the default trials, especially @ `-o max`.

andrews05:
I've just pushed an improvement to the pairwise swaps to get even better results. The figures have been updated above.

I guess I didn't spell it out, but these are quite significant gains in the area of palette sorting. The ezeng method (+ swaps) is far better than mzeng and battiato combined.
These algorithms are all trying to do the same thing—minimise entropy in the image data—so if one method is extremely good at doing that then we don't really need others. The battiato method was already providing minimal gains over the mzeng method and now its value drops to almost nothing, below even the basic luma sort. I.e. it's no longer pulling its weight. See #514 where I mention the lack of value in 120 other methods.

Note we still do the luma sort as well as evaluate the input palette if there was one.

> Did you preserve the heuristic I suggested- that palette position 0 should be reserved for the most frequently occurring horizontal-edge colour (i.e. common row start and end colour)?

I'm continuing to use the same "popular colour" heuristic from mzeng in ezeng (see #602).

> I have wondered if the palette variations are within a threshold (~5% ?), we should consider no clear winner and continue to test into the next stage?

This is what #749 is doing.

ace-dent:
@andrews05 - quick test results on latest artefact. 
For some highly optimized images, that are 'palette sensitive', this new sorting yields ~1% improvement! Nice!
Does `zeng` work for 4 colour palettes (2bpp)- as we use to skip for `mzeng`?
Is it useful for me to share examples where `mzeng` / `battiato` perform better?

ace-dent:
Additional feature request (can raise as a separate issue):
For **2 colour** (1bpp) images- including grayscale- can we use different logic to simply test both palette variations?

TPS:
> For **2 colour** (1bpp) images- including grayscale- can we use different logic to simply test both palette variations?

*Generally*, below a certain deduped color-count threshold (maybe 4?), may we have _all_ palette-order permutations tested @ `-o max`?

ace-dent:
@TPS - I had considered the same, but... 

$$ 4! = 24 $$

TPS:
@ace-dent @ `-o max`, maybe ok?  Or perhaps as a brand-new (perhap debug) switch?

ace-dent:
@andrews05 - testing with latest CI build artefact @ [406b335](https://github.com/oxipng/oxipng/pull/820/commits/406b335c4df3c078063f22d79ea9dc4ebd11bd25).

Testing attached image... It looks like `100 colors` (6168 bytes) is selected over `zeng` (6140 bytes) palette? 

```bash
oxipng-pal2 -vvvo6 -f 0-9 --zopfli --zi 9000 --ziwi 300 blue_hl81.png  
```

```text
Processing: blue_hl81.png
    190x268 pixels, PNG format
    8-bit Indexed (100 colors), non-interlaced
    IDAT size = 5218 bytes
    File size = 5587 bytes
Eval: 8-bit Indexed (luma sort)     None       6143 bytes
Eval: 8-bit Indexed (luma sort)     Bigrams    6734 bytes
Eval: 8-bit Indexed (zeng sort)     None       6140 bytes
Eval: 8-bit Indexed (100 colors)    None       6168 bytes
Eval: 8-bit Indexed (100 colors)    Bigrams   >6143 bytes
Eval: 8-bit Indexed (zeng sort)     Bigrams   >6143 bytes
Transformed image to 8-bit Indexed (100 colors), non-interlaced
Trying 10 filters with zopfli, zi = 9000
Eval: 8-bit Indexed (100 colors)    None       5530 bytes
Eval: 8-bit Indexed (100 colors)    Up        >5530 bytes
Eval: 8-bit Indexed (100 colors)    Average   >5530 bytes
Eval: 8-bit Indexed (100 colors)    BigEnt    >5530 bytes
Eval: 8-bit Indexed (100 colors)    Sub       >5530 bytes
Eval: 8-bit Indexed (100 colors)    Paeth     >5530 bytes
Eval: 8-bit Indexed (100 colors)    MinSum    >5530 bytes
Eval: 8-bit Indexed (100 colors)    Bigrams   >5530 bytes
Eval: 8-bit Indexed (100 colors)    Brute      5530 bytes
Eval: 8-bit Indexed (100 colors)    Entropy   >5530 bytes
    IDAT size = 5218 bytes (0 bytes decrease)
    file size = 5587 bytes (0 bytes = 0.00% decrease)
Could not optimize further, no change written: blue_hl81.png
--------------------
Files processed: 1/1   
Input size: 5.46 KiB (5587 bytes)
Output size: 5.46 KiB (5587 bytes)
Total saved: 0 bytes (0.00%)
(1 file could not be optimized further)
```
blue_hl81

P.S. - Please let me know if the code isn't ready for wider testing yet. 🙏

andrews05:
Unfortunately palette sorting algorithms don't really work for low bpp. Theoretically perhaps an algorithm could be designed for that but it would probably be quite complex. In the meantime, brute force is probably your best bet (reminder: oxipng is not a brute force optimiser 😄).
That said, with these improvements here you may well find that more images that _could_ be low depth now actually prefer 8-bit and get compressed better.

> Testing attached image... It looks like 100 colors (6168 bytes) is selected over zeng (6140 bytes) palette?

Ah, yes this can be somewhat confusing. The description of the sort method unfortunately doesn't get retained beyond the evaluation trials, so all you can see after that is the description of the colour type. You have to check which evaluation was smallest to work out which one it actually picked.

Note that nothing has changed in this PR with regard to the processing pipeline and evaluation flow. It just replaces the sorting algorithms.

ace-dent:
Samples from comparing the PR to current release version. Note: all these files have been highly optimised before, so oxipng typically fails; the comparison is made by looking at verbose output for the first evaluation stage.

[luma better than zeng.zip](https://github.com/user-attachments/files/28630987/luma.better.than.zeng.zip)
[battiato better than zeng.zip](https://github.com/user-attachments/files/28630988/battiato.better.than.zeng.zip)
[mzeng better than zeng.zip](https://github.com/user-attachments/files/28630989/mzeng.better.than.zeng.zip)

@andrews05 - Please can you modify your benchmark process, to count number of files that are optimised better -vs- worse. The problem with a net improvement value (total bytes saved) is, in the extreme, it could be a single image that benefits massively but 99 images that are all slightly worse.

I have also given some thought to the edge-color heuristic. It's only likely to yield benefits when most rows are filter type `0`... but for complex images (where the `zeng` sorting might excel), are we hurting compression by artificially introducing this rule? I feel there is someway to calculate the maximum possible upside and adopt or disregard the edge color... just cannot think of a simple metric right now... 🤔

ace-dent:
> ... Ah, yes this can be somewhat confusing. ...

Perhaps we can clarify the first stage evaluation text? Instead of repeating `(100 colors)`, we could state `(original)` to better describe the palette sorting method?
e.g.

```text
Processing: blue_hl81.png
    190x268 pixels, PNG format
    8-bit Indexed (100 colors), non-interlaced
    IDAT size = 5218 bytes
    File size = 5587 bytes
Eval: 8-bit Indexed (luma sort)     None       6143 bytes
Eval: 8-bit Indexed (luma sort)     Bigrams    6734 bytes
Eval: 8-bit Indexed (zeng sort)     None       6140 bytes
Eval: 8-bit Indexed (original)    None       6168 bytes
Eval: 8-bit Indexed (original)    Bigrams   >6143 bytes
Eval: 8-bit Indexed (zeng sort)     Bigrams   >6143 bytes
Transformed image to 8-bit Indexed (100 colors), non-interlaced
...
```

Is that less likely to cause confusion?

TPS:
> `(original)`

Perhaps just `(unsorted)`, `(not sorted)`, or `(no sort)`?  So as to make plain what didn't happen **_&_** to match other wording?

TPS:
> Unfortunately palette sorting algorithms don't really work for low bpp.

> brute force is probably your best bet (reminder: oxipng is not a brute force optimiser 😄)

Are any of you aware of _any_ software that does exhaustive palette search for @ least the lowest bpp like we've suggested?

@ace-dent https://encode.su/threads/3879-Optimizing-palette-order-for-indexed-png-images 🙇🏾‍♂️

andrews05:
> Samples from comparing the PR to current release version. Note: all these files have been highly optimised before, so oxipng typically fails; the comparison is made by looking at verbose output for the first evaluation stage.
> 
> [luma better than zeng.zip](https://github.com/user-attachments/files/28630987/luma.better.than.zeng.zip)
> [battiato better than zeng.zip](https://github.com/user-attachments/files/28630988/battiato.better.than.zeng.zip)
> [mzeng better than zeng.zip](https://github.com/user-attachments/files/28630989/mzeng.better.than.zeng.zip)

Sorry, I'm not quite sure what you're trying to show here. I took a look at these and if you ignore the input palette then in all cases except one the luma sort evaluation is better than all other methods. The only exception was "08-ct-c6-c3.png" where ezeng actually came out best at 13325, beating battiato at 13390. (Note that "23-dt-rgb-c3.png" doesn't show a luma evaluation which just means the luma sort was the same as the input palette.)

Of course, there will always be some cases where battiato or mzeng is better than _both_ ezeng and luma, but that will also be true of any other sort method (such as popularity). It doesn't mean that all known methods are _good_ and that we should try all of them.

> I have also given some thought to the edge-color heuristic. It's only likely to yield benefits when most rows are filter type 0... but for complex images (where the zeng sorting might excel), are we hurting compression by artificially introducing this rule?

ezeng uses the popular-colour heuristic rather than edge-colour, so it won't be hurt by this rule.

> Perhaps we can clarify the first stage evaluation text? Instead of repeating (100 colors), we could state (original) to better describe the palette sorting method?

Hm, it's a good thought. Unfortunately it gets a little tricky if there's a low-depth evaluation, which also shows the colour count at the moment. The palette will usually be the luma sort but could be the original if the luma sort was the same. It would be better to try to preserve the description through the later stages, but that's also tricky with the current architecture. (I might have done this in #749, I don't quite remember.)

> Are any of you aware of any software that does exhaustive palette search for @ least the lowest bpp like we've suggested?

I'm afraid not. Note that at 4-bit, 16! = 20 trillion which is far outside the realm of practicality 😄

andrews05:
Facepalm time. After reading the original papers more carefully, I realised the new method is not the zeng method at all. I had originally understood mzeng to be a faster but not quite as effective version, as it mentioned how zeng's "performance can be improved". I now realise "performance" here actually refers to effectiveness, not speed - mzeng is _better_, not _faster_ than zeng. The new method is exactly that - something new. I'm now calling it "ezeng".

(This happened because I tried implementing zeng and was confused by the results so I asked an agent to check if I had got it correct. It suggested some “fixes” and I foolishly believed it because the results were more in line with what I was expecting.)

andrews05:
> @andrews05 - Please can you modify your benchmark process, to count number of files that are optimised better -vs- worse. The problem with a net improvement value (total bytes saved) is, in the extreme, it could be a single image that benefits massively but 99 images that are all slightly worse.

You're right that this could theoretically happen with limited test cases. But while I don't check individual image results, I test across a number of different image sets and observe average gains in all of them. The results I showed in the first post are just two examples. So while _some_ files might experience losses, these are rare and the losses are tiny.

TPS:
> I'm afraid not. Note that at 4-bit, 16! = 20 trillion which is far outside the realm of practicality 😄

Of _course_ that is (until there's some sort of generally-available quantum computing, I suppose, but probably PNG itself will be moot by then).  We only had in mind ≤4 colors, since, as mentioned above, 4!=24.

AlexTMjugador:
> (This happened because I tried implementing zeng and was confused by the results so I asked an agent to check if I had got it correct. It suggested some “fixes” and I foolishly believed it because the results were more in line with what I was expecting.)

Hmm, AI can certainly suggest interesting ideas, but it saying things or writing code alone is not a rigorous proof that the intended cost-increase minimization behaves as expected, especially for files outside your test set. How thoroughly have you verified the mathematical approach here?

ace-dent:
@andrews05 :

> Sorry, I'm not quite sure what you're trying to show here.

Simply that while `ezeng` surpasses `mzeng` (mostly), but there are occasions where `battiato` still seems useful. 
Please consider testing with attached example file... 

```text
Eval: 8-bit Indexed (battiato sort) None       66393 bytes
Eval: 8-bit Indexed (mzeng sort)    None       66935 bytes
-vs-
Eval: 8-bit Indexed (ezeng sort)    None       66548 bytes
```

So this PR with `... -o6 -vvv ...` creates a file: `63948 bytes (4.71% smaller)`.
Then running the output through the current release: `63671 bytes (0.43% smaller)`. So using the `battiato` sort saves ~277 bytes in the final output.

signs

I accept brute forcing palettes may not makes sense for most Users- and perhaps is the role for a separate program. It just seems a shame to discard potentially useful methods.

<hr>

> I'm afraid not. Note that at 4-bit, 16! = 20 trillion

I think @TPS and I agreed that 2-colour (1bpp) images should have the **2** palette variations tested. Personally I'm unsure whether 4-colour (2bpp) is still in scope, if we are not brute forcing palettes. For 2-colour *grayscale* we should also test the 1 palette variation where the order of the black and white are swapped; this can win due to the edge colour and row filter effect.

<hr>

Also just to add: Please keep up this excellent work!! 🙏🙏🙏
I'm really grateful for the progress you are making here.

andrews05:
I've just pushed a change to make the swaps a little faster and help reduce the speed loss.

> Hmm, AI can certainly suggest interesting ideas, but it saying things or writing code alone is not a rigorous proof that the intended cost-increase minimization behaves as expected, especially for files outside your test set. How thoroughly have you verified the mathematical approach here?

Yeah, good question. Despite appearances, I'm not actually an expert in entropy reduction 😆. What I can say is:
- Zeng/mzeng choose between prepend and append for each new element to add. Ezeng iterates all insertion positions and calculates the best one (this is the only difference). It's a more exhaustive process and logically it makes sense that it would give better results.
- The calculations for this seem sensible to me, though I can't say whether it's the _most_ sensible approach.
- It still inserts every element exactly once. That is, it can't break anything other than potentially giving a result that doesn't compress well.
- The AI didn't know anything of my test files, so it wasn't optimising for any specific data. It's still just a generic entropy-minimising algorithm.
- I tested over a number of image sets, with at least 500 files total, and all sets showed notable gains.

andrews05:
@ace-dent I understand your concern, but I'm really focused on pushing the average case. There will _always_ be edge cases that seem to benefit from doing some odd thing or another. But working out which odd things are actually worthwhile doing takes a lot of time and effort both implementing and extensively testing. Due to oxipng's current architecture, there are also two additional concerns:
1. Evaluation mis-predictions, which increase in chance the more evaluations are added, particularly when two or more are very similar to each other.
2. Amount of "effort" to spend on transformations and their evaluations. Currently we don't have any real way to scale this with the o level. We do disable some transformations at o2 and lower by checking the compression level and "fast" flag, but that's all. Meaning every "odd thing" that gets added will slow down mid levels, which is not desirable.

These are all things which hypothetically could be improved upon at some point. But I'm afraid I just don't have the inclination to spend a lot of time pursuing such small gains in edge cases.

TPS:
@andrews05 IIuc, you're essentially saying, practically, for people who've edge cases served this way, we'd have to use previous versions of `OxiPNG` which still include the algos we'd like to test (somewhat like @ace-dent did above), since ongoing versions will no longer have them.

I suppose that's acceptable?  We're the 1s out here wasting the cycles.

@javiergutierrezchamorro Perhaps take note for highest FO settings for PNG? 🙇🏾‍♂️

ace-dent:
@andrews05 - thanks. I appreciate where you are coming from. Would you consider hiding `battiato` behind a switch, or even adding `-o7`? This avoids keeping a legacy copy of `oxipng` installed for that purpose.

OR, in terms of speed for `ezeng`- perhaps that should be behind `-o7` (so `-o6` functions as before) ...?

Re. (1) Evaluation mis-prediction ... when your future evaluator changes land, this becomes less of an issue? 

Also note re. 'average case'... you seem to be looking at a mean average for your corpus. As I mentioned, taking an aggregate value may hide the actual number of files that are advantaged or disadvantaged by the change (like the modal average). If you can show that 400/500 files clearly benefit, then I'm sold. If its like, 100 improve, 50 are worse and 350 are the same... it's less convincing... the test file I provided was just one of about 20 I tested; it was not a contrived example.

andrews05:
Guys, every change we make that pushes the average forward always causes regression in one file or another. This is a _big win_, take it! 🙂
Enough ink has been spilled over this.

AlexTMjugador:
> Guys, every change we make that pushes the average forward always causes regression in one file or another. This is a big win, take it! 🙂

To address this sort of concerns in a more principled way, I'd suggest comparing compression-related changes using a more comprehensive set of [five-number summary](https://en.wikipedia.org/wiki/Five-number_summary) or [seven-number summary](https://en.wikipedia.org/wiki/Seven-number_summary) statistics. This would give everyone involved a clearer picture of how a change affects not just average or median cases, but also the distribution across quartiles and outliers. To make that easier, we could check in a script it that on the repository. (And if we get fancy with it, we could even render some fancy boxplots with matplotlib or something!)

Other than that, the PR looks quite sensible to me, especially given that you've also reviewed it to also be sensible. Thank you for your work! :heart:

AlexTMjugador (approved):

ace-dent:
@TPS - Just FYI: ECT does support brute forcing palettes (not necessarily in a smart way!) using `--pal_sort=120`.

TPS:
> ECT does support brute forcing palettes 
> using `--pal_sort=120`

Not quite.  Somewhere here, I remember reading that all of ECT's strategies were tried, but found wanting to mzeng, &c, OxiPNG had.  (Probably @andrews05 coding this previously?).  It's _definitely_ not brute (from encode.su threads) overall, but might devolve to that in the smallest palettes?

andrews05:
Yup, ect uses the same 120 methods I mentioned [here](https://github.com/oxipng/oxipng/pull/514#issuecomment-1572745056).

TPS:
@andrews05 Possible to use some, e.g., [SIMD filter *de*coding optimizations](https://github.com/connorskees/simd-png) for palette sorting?

andrews05:
Hm, not quite sure what you mean. Filter decoding and palette sorting are completely unrelated functions.

PR #824, opened 2026-06-11
Include version info in verbose output

### LINKED ISSUE #821: Display binary version for very verbose output
When testing the program, it can be desirable to select highly verbose output `oxipng -vvv ...`.
In this situation, it would be useful to also output the binary's version details as the first line. This makes logging test results easier when comparing versions. (Relates also to #727).

If there is concern about backwards compatibility, we should at least allow: `oxipng -vvv -V ...`.

andrews05:
Fixes #821 

Also changed the command info to use cargo env vars. The package description and the about were actually slightly different before but I don't think it matters much which text is used.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -19,10 +19,10 @@ pub fn build_command() -> Command {
     // so that they can fit on a single line in an 80 character terminal.
     // Long help descriptions are soft wrapped here at 90 characters (column 91) but this does not
     // affect output, it simply matches what is rendered when help is output to a file.
-    Command::new("oxipng")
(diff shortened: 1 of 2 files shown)


ace-dent:
Thanks @andrews05 ! Perfect!

... although I forgot we don't increment version number between releases!... 😅

TPS:
> ... although I forgot we don't increment version number between releases!... 😅

Maybe there's a finer semver &/or build # that could be exposed for this?

AlexTMjugador (approved):

PR #825, opened 2026-06-13
Remove filter restrictions for interlaced images

andrews05:
Filtering an interlaced image was previously avoiding vertical filters (up, average, paeth) for the first line of each interlacing pass. As best I can tell from the history, this was done because it wasn't zeroing out the previous line between each pass, so applying a vertical filter would have corrupted the image. Zeroing out the previous line has since been implemented, so there's no longer any reason to avoid vertical filters. Removing the restriction simplifies the code and can give a very small compression improvement for interlaced images.


AlexTMjugador (approved):
I'm always down for less code to maintain! Good catch 👍

PR #827, opened 2026-06-18
Adjust filters for levels 0 and 1

### LINKED ISSUE #775: Clarify levels 0 and 1 in the Manual
Currently, the manual says `(filter chosen heuristically)` for levels 0 and 1 but doesn't explain what that heuristic is.
 I tried reading the code (sorry if I missed something obvious, I'm not familiar with rust) and as far as I can tell, the list of filters is set cleared in the apply_preset_0 function? Where does that list get refilled?

andrews05:
The "heuristic" used for the filter on levels 0 and 1 was unclear, inconsistent, and not really appropriate. This PR removes the heuristic and changes level 0 to use None & MinSum and level 1 to use None & Bigrams. They're now slightly better on average _and_ up to 10% faster.

Closes #775


TPS:
Does this touch `man` or help output?

andrews05:
@TPS Yes, the heuristic is removed so the help just shows the filters as being 0,5 now.

src/lib.rs
```diff
@@ -389,20 +389,19 @@ fn optimize_raw(
...
-        }
+        Deflater::Libdeflater { compression } if compression < 11 => 5.min(compression),
+        _ if opts.fast_evaluation => 7,
         _ => 8,
     };
     let eval_deflater = Deflater::Libdeflater { compression };
-    // If only one filter is selected, use this for evaluations
-    let eval_filters = if opts.filters.len() == 1 {
+    let eval_filters = if opts.filters.len() <= 2 {
+        // If only one or two filters are given, use them for evaluations
```
andrews05, line 404: I had to throw out my [reasoning for not doing this](https://github.com/oxipng/oxipng/pull/582#issuecomment-1821694036) before. If someone chooses to use 0,9 then so be it 😄

andrews05:
I'm converting this to draft as I'd like #830 to land first. The performance changes may affect the choices here.

andrews05:
Following #830, I've now changed this to use None & Bigrams for level 1.

andrews05:
Closing in favour of #842

Issue #666, opened 2025-01-05
(BUG) Some PNGs refuse to work with interlacing despite consuming more space after enabling it


seacat17:
So I got this very confusing issue. I have a project and I am planning to optimize all the images (around 200 PNGs) in it with oxipng.

I want to also add interlacing to these images for faster display. I know the drawbacks and the benefits are just too good - I love interlacing and I love when the entire image is instantly displayed and then gains more detail during the loading process.

HOWEVER I have met a weird issue. You see, I've been testing oxipng for a few days already and I keep getting these weird results with some of my PNG files.

First case:
This is how it SHOULD work. I have this picture by awesome-llama that I was testing oxipng with.
Here's the source file.
![TM3D_facility_render](https://github.com/user-attachments/assets/76cb7d98-9088-47e1-a5ca-e8ddbb5ad713)
You'll probably see that it is loaded from top to bottom.

I ran it thru oxipng and here's the result.
![COMP](https://github.com/user-attachments/assets/ee6e75af-5c52-4751-99df-7568f5f949dd)
Smaller file size and proper interlacing. Pure bliss.

Second case:
This is how it SHOULD NOT work. Here's the source file.
![defpng](https://github.com/user-attachments/assets/b7471316-7c69-458b-a7f4-e12a4fd382dd)
Just some random CS2 screenshot done on Linux with MATE's "Save Screenshot" feature. Nothing new here, just a default PNG.

I use LITERALLY THE SAME SETTINGS here, so it should be just as good as the first case.
![oxipng](https://github.com/user-attachments/assets/e9ed5385-de7a-4d99-bbc5-3168daab83d5)
I know that the file size is not reduced in this case but this is not why I'm here. This picture should be interlaced but it isn't - it is still loading from top to bottom.

And to make sure that it wasn't just me using some wrong settings - I disabled interlacing and optimized the same picture again.
![oxipng_no_int](https://github.com/user-attachments/assets/3aeb4496-e233-47a1-acd1-78e4e9ef0299)
The file size is actually reduced this time because there's no interlacing drawback.

This means that oxipng actually tries to apply the interlacing but something goes wrong and it fails to.

And there's so many pictures like this one. I am going to optimize an entire digital comic book with around 200 pages that are all PNGs. I have no right to post them here but they all have literally same problem. I want to apply interlacing so the pages could display instantly even with slow connection, but I just can't. Why? I don't know.

Here's the terminal output with increased verbosity:

[...]

PR #831, opened 2026-06-27
Refactor sorting for improved performance

andrews05:
I guess #820 could have used some more time cooking. There are some big performance improvements here.
- Dramatically improve swaps performance by splitting the inner loop into three. (It still boggles me how odd things like this can make such a difference. If anyone can tell me how to keep a single loop with the same performance, I'm all ears.)
- Refactor co-occurrence matrix into a struct with a flat data array instead of a 2D one. This makes both ezeng and the swaps faster.
- Move calculation of most popular colour into the new struct, where we can determine it very quickly from the matrix.
- Simplify edge list construction by omitting zero-weight edges and retaining the weight in the output. (In typical images, the vast majority of edges are zero-weight. In quantised photographic material it might be around 50%.)
- Separate the matrix construction from the sort functions for convenience and more informative benchmarks.
- Fix matrix construction to not retain prev_val between lines, which doesn't make sense. This can affect output slightly.
- Support interlaced images in the matrix, allowing them to benefit from the advanced sorting algorithms.
- Fixed the `palette_8_should_be_palette_8` test image which already had the luma sort applied, so the luma bench result was misleading.

Master:
```
test reductions_palette_sort                        ... bench:       2,365.25 ns/iter (+/- 42.04)
test reductions_palette_sort_battiato               ... bench:   1,622,760.40 ns/iter (+/- 14,406.88)
test reductions_palette_sort_ezeng                  ... bench:   7,973,558.40 ns/iter (+/- 106,709.13)
test reductions_palette_sort_ezeng_with_swaps       ... bench:  47,167,250.00 ns/iter (+/- 128,985.46)
test reductions_palette_sort_mzeng                  ... bench:   1,631,920.80 ns/iter (+/- 6,677.21)
```
PR. Note the matrix has been split out into its own bench and its time should be added to the other results to properly compare against master above.
```
test reductions_palette_sort                        ... bench:       2,350.68 ns/iter (+/- 93.18)
test reductions_palette_sort_battiato               ... bench:      68,448.67 ns/iter (+/- 4,328.97)
test reductions_palette_sort_build_matrix           ... bench:     638,366.65 ns/iter (+/- 5,209.13)
test reductions_palette_sort_ezeng                  ... bench:   5,522,687.50 ns/iter (+/- 61,715.52)
test reductions_palette_sort_ezeng_with_swaps       ... bench:  19,365,270.80 ns/iter (+/- 59,778.46)
test reductions_palette_sort_mzeng                  ... bench:     101,971.76 ns/iter (+/- 903.48)
```
PR with updated bench image:
```
test reductions_palette_sort                        ... bench:      62,946.48 ns/iter (+/- 1,464.61)
test reductions_palette_sort_battiato               ... bench:      68,901.66 ns/iter (+/- 1,071.80)
test reductions_palette_sort_build_matrix           ... bench:     640,320.90 ns/iter (+/- 7,143.78)
test reductions_palette_sort_ezeng                  ... bench:   5,576,681.25 ns/iter (+/- 67,719.78)
test reductions_palette_sort_ezeng_with_swaps       ... bench:  36,055,941.70 ns/iter (+/- 48,153.08)
test reductions_palette_sort_mzeng                  ... bench:     100,387.50 ns/iter (+/- 1,093.16)
```


ace-dent:
Wow! Great results! 🔥

vlakoff:
**Disclaimer:** I generated the following suggestions with Claude (AI). I'm posting them as-is, for whatever it's worth — take with the appropriate grain of salt, and feel free to disregard anything that doesn't hold up.

### 1. `most_popular_color()` is still O(n²)

This point turned out to be irrelevant, so I'm collapsing its original write-up below, and I've also hidden my follow-up messages that elaborated on it.

<details>

<summary>Kept for reference — not worth pursuing</summary>

```rust
fn most_popular_color(&self) -> (usize, u32) {
    let mut best = (0, 0);
    for i in 0..self.num_colors {
        let sum: u32 = self.row(i).iter().sum();   // O(n) per color → O(n²) total
        ...
    }
}
```

For a 256-color palette that's 65,536 additions. The per-color totals could instead be accumulated **during** `build()`, at the same time each `data[prev_val * num_colors + val] += 1` happens — a `row_sums: Vec<u32>` updated in lockstep would bring this down to O(n). Given the rest of the PR is specifically about eliminating redundant recomputation, this seems like a leftover spot.

**Edit:** Correction on the complexity claim above. Accumulating `row_sums` directly in the raw scanline counting pass (as literally described) is trickier than it sounds — that pass produces asymmetric, pre-symmetrization counts, and deriving the final `/4` popular-color total from those would need careful re-derivation.

What actually works cleanly: accumulate `row_sums` in the existing symmetrization loop (`for i in 0..num_colors { for j in 0..=i { ... } }`) instead of in a separate pass afterward. Each matrix cell is already visited exactly once there, so it's a "free" accumulation that removes the second O(n²) pass currently in `most_popular_color()`.

To be precise, the arithmetic work (number of additions) is comparable either way — worth noting that's specifically because, *currently*, only one sort algorithm is active and `most_popular_color()` is only called once per matrix. Even with a single call, there's still a modest win: the accumulation happens right as each cell is written, while it's still hot in L1, instead of re-reading the whole matrix a second time afterward — so it's memory locality rather than raw op count that's being saved.

That's a real, if more modest, win — one fewer full O(n²) traversal, not the true O(pixels) result I originally implied. A genuinely O(n) `most_popular_color()` would still require accumulating during the raw counting pass, which needs someone to work out that pre-symmetrization → final-sum relationship first.

</details>

### 2. `sort_by` → `sort_unstable_by` in `weighted_edges`

```rust
edges.sort_by(|(_, w1), (_, w2)| w2.cmp(w1));
```

Stability doesn't matter here (tie order between equal weights is arbitrary anyway). `sort_unstable_by` should be a free win on what can be an O(n² log n) operation for large palettes.

### 3. Undocumented diagonal-doubling in `build()`

```rust
for j in 0..=i {
    let opposite = j * num_colors + i;
    data[row_start + j] += data[opposite];
    data[opposite] = data[row_start + j];
}
```

When `j == i`, `opposite == row_start + j`, so the diagonal cell adds to itself. This matches the legacy 2D symmetrization behavior, but it's subtle enough that a future reader might "fix" it as a bug. Worth a one-line comment noting it's intentional.

### 4. Structural duplication across `mzeng_reindex` and `ezeng_reindex`

Both repeat the same skeleton: seed `remapping` with the top two `weighted_edges`, maintain a `sums` list, find the best candidate, `swap_remove` it, update remaining sums. Only the actual insertion logic differs (prepend/append vs. optimal-position). A shared helper like:

```rust
fn greedy_chain_build(matrix: &CoOccurrenceMatrix, insert_fn: impl Fn(...) -> usize) -> Vec<usize>
```

could cut down maintenance surface — though this is a bigger refactor than what's in scope here, and only worth it if mzeng is going to stick around long-term (it's currently marked unused).

`battiato_reindex` doesn't share this skeleton, though — it iterates the edge list to merge chains rather than seeding from the top two edges and maintaining a `sums` list, so it would stay separate from any such refactor.

### 5. Index types are `usize` even though `num_colors ≤ 256`

`remapping: Vec<usize>` and `weighted_edges: Vec<(usize, usize)>` — since color indices always fit in a `u8`, switching to `u8`/`u16` could reduce memory bandwidth, especially for `weighted_edges` (up to ~32k pairs for 256 colors). This seems in the spirit of a perf-motivated PR, at the cost of a few extra `as usize` casts.

### 6. `row()` bounds-checking in hot loops

```rust
#[inline]
fn row(&self, row: usize) -> &[u32] {
    let start = row * self.num_colors;
    &self.data[start..(start + self.num_colors)]
}
```

This is called from the hottest loops (`ezeng_reindex`, `pairwise_swap_search`). Since `row < num_colors` always holds by construction, `get_unchecked` behind a documented `unsafe` block could be considered — but only if profiling actually shows the bounds checks are a real bottleneck, not preemptively.

### 7. Minor naming nit

```rust
fn row(&self, row: usize) -> &[u32] {
```

The parameter shadowing the method name (`self.row(row)`) is a bit awkward to read. `index` or `color` would be clearer, especially since it's really a color index rather than an abstract "row".

vlakoff:
For anyone interested in implementing the `most_popular_color()` optimization from point 1 above, here's a concrete sketch.

The idea: accumulate `row_sums` inside the *existing* symmetrization loop in `build()`, instead of doing a separate full traversal in `most_popular_color()`. Each matrix cell is already visited there with its final, symmetrized value, so the accumulation is essentially free — it doesn't change `build()`'s asymptotic complexity (it's already O(n²)), but it eliminates the second full O(n²) traversal that `most_popular_color()` currently does on every call.

### 1. `build()` returns `row_sums` alongside `data`

```rust
fn build(num_colors: usize, png: &PngImage) -> (Vec<u32>, Vec<u32>) {
    // ... raw scanline counting pass unchanged ...

    let mut row_sums = vec![0u32; num_colors];
    for i in 0..num_colors {
        let row_start = i * num_colors;
        for j in 0..=i {
            let opposite = j * num_colors + i;
            data[row_start + j] += data[opposite];
            data[opposite] = data[row_start + j];

            // accumulate while each cell already holds its final value
            row_sums[i] += data[row_start + j];
            if j != i {
                row_sums[j] += data[row_start + j];
            }
        }
    }

    (data, row_sums)
}
```

### 2. `CoOccurrenceMatrix` gets a new field, and `from()` is updated accordingly

```rust
pub struct CoOccurrenceMatrix {
    num_colors: usize,
    data: Vec<u32>,
    weighted_edges: Vec<(usize, usize)>,
    row_sums: Vec<u32>,   // new
}

impl CoOccurrenceMatrix {
    pub fn from(png: &PngImage) -> Option<Self> {
        // ...
        let (data, row_sums) = Self::build(num_colors, png);
        let weighted_edges = Self::weighted_edges(num_colors, &data);
        Some(Self { num_colors, data, weighted_edges, row_sums })
    }
    // ...
}
```

### 3. `most_popular_color()` just reads the precomputed sums

```rust
fn most_popular_color(&self) -> (usize, u32) {
    let mut best = (0, 0);
    for (i, &sum) in self.row_sums.iter().enumerate() {
        if sum > best.1 {
            best = (i, sum);
        }
    }
    best.1 /= 4;
    best
}
```

The `/4` factor is untouched — `row_sums[i]` is constructed to be exactly equal to what `self.row(i).iter().sum()` currently produces, just computed once during construction instead of on every call.

### Callers are unaffected

`apply_most_popular_color()` and its callers in `sorted_palette_mzeng`, `sorted_palette_ezeng`, `sorted_palette_battiato` don't need any changes — same signature, same call site.

### Suggested safety check

Before merging, it'd be worth a quick unit test asserting `row_sums[i] == matrix.row(i).iter().sum()` for all `i` on a test matrix, to lock in the equivalence with the old behavior.

The one subtle spot worth double-checking is the `j != i` diagonal handling in the accumulation — when `j == i`, `row_sums[i]` and `row_sums[j]` refer to the same slot, so the diagonal cell must only be added once (matching how `.sum()` counts it), not twice.

### Further optimization

If multiple sort algorithms (mzeng/ezeng/battiato) ever end up active at once on the same matrix, it'd be worth going one step further — have `build()` compute and return the final `(index, count)` most-popular-color result directly instead of `row_sums`, and turn `most_popular_color()` into a plain O(1) getter. That would avoid repeating even the O(n) max-search on every call. Not worth it as long as only one algorithm runs per matrix, though.

vlakoff:
That said, it's worth stepping back: even in its current form — which, per the PR description, was already moved off the O(pixels) full-image scan and onto the bounded O(n²) matrix (n ≤ 256) specifically "to determine it very quickly from the matrix" — `most_popular_color()` is likely already extremely fast in absolute terms.

65,536 additions over contiguous `u32` memory is on the order of microseconds at most on modern hardware.

Compare that to the rest of the pipeline: `build_matrix` alone takes ~1ms, `ezeng` ~8ms, and `ezeng_with_swaps` ~47ms (per the benchmarks posted in this PR and #820). That's 2-3 orders of magnitude more time spent elsewhere.

If that holds up under actual profiling, this point — despite the back-and-forth above — probably isn't worth pursuing; it'd be optimizing something that's already noise relative to the rest of the sort pipeline.

The other points from [my initial comment](https://github.com/oxipng/oxipng/pull/831#issuecomment-4862354780) may still be worth a look, though.

andrews05:
@vlakoff Thanks for the analysis, it's quite interesting!

1. **most_popular_color:** Your final comment there is absolutely correct. The calculation here in this PR is 100x faster than the previous implementation and is totally negligible. Even if it could be made faster (and I'm skeptical), it's really not worth it.
2. **sort_by → sort_unstable_by:** This is actually a pretty interesting point. Profiling shows the performance difference here is considerable. However, the comment that "stability doesn't matter here" and "tie order between equal weights is arbitrary anyway" is actually incorrect. At the time that we construct the matrix, the palette has been _previously_ sorted using our luma sort algorithm, and changing to unstable sort here can negatively impact the results. I'll do a little more experimenting but I don't think this will be a good change despite the speed improvement.
3. **diagonal-doubling:** I'm not really concerned about claude thinking this is surprising, but I just thought I'd note that despite the comment that it "matches the legacy 2D symmetrization behavior", there actually is a difference here. The so-called "legacy" code (that I did in #820) was actually incorrect because it iterated the entire matrix and ended up doubling up all the _non_-diagonal values. In practice this had no impact on the sort result but now that we also use the matrix to determine the most popular colour, that error could have been problematic. So I've fixed it here to only iterate up to (and including) `i`.
4. **Structual duplication:** A good thought but as it doesn't impact performance it's not something I'll bother with at this point.
5. **usize vs u8:** Also a good point, however the memory pales in comparison to the 256K already required for the matrix data.
6. **bounds checking:** I'd prefer to avoid unsafe, but I'll experiment a bit here to see how much difference it makes.
7. **row naming:** Meh 😄

vlakoff:
Thanks for the detailed response.

A few follow-ups on some of the points:

### 2. `sort_by` → `sort_unstable_by`

Weights are bounded `u32` counts (pixel adjacency counts). Instead of dropping the switch to `sort_unstable_by` entirely, a counting/bucket sort by weight could work.

That gives O(n) sorting while naturally preserving insertion order for ties — equivalent to a stable sort, but with the perf win. Might be worth a quick experiment, but no strong opinion if you'd rather leave it as-is.

### 3. Diagonal-doubling fix

Great catch fixing the actual bug in the legacy symmetrization. It was silent before (no visible effect on sort order), and only became relevant once the matrix started feeding `most_popular_color`.

Still seems worth adding that one-line comment on the `0..=i` bound, so it doesn't get "fixed" back to the buggy version later.

Also, since there's now a real behavioral difference from before, a small regression test on `CoOccurrenceMatrix::build`/`most_popular_color` (known input, expected output) could help. No existing test caught this the first time around, which is what makes me a bit wary here.

### 5. Index types u8/u16

Worth a second look — the memory comparison actually cuts the other way.

The matrix is `num_colors² × 4 bytes`, ~256KB at 256 colors. But `weighted_edges` is a `Vec<(usize, usize)>`, 16 bytes per pair, up to 32,640 pairs worst case — that's ~510KB, larger than the matrix itself.

`u8`/`u16` would shrink that to ~65–130KB (4–8x smaller), plus better cache locality in the loops that scan it (`battiato_reindex` walks the whole list, `ezeng`/`mzeng` scan parts of it). Similar spirit to the 2D→flat-array change that already helped elsewhere.

Worth noting this is the worst case at 256 colors; for smaller palettes it's probably negligible. But given the PR is perf-focused, might be worth a quick check on a large-palette image before ruling it out.

andrews05:
> But weighted_edges is a Vec<(usize, usize)>, 16 bytes per pair, up to 32,640 pairs worst case — that's ~510KB, larger than the matrix itself.

Oof, yes, I was only looking at the `remapping`. That's very valid and could definitely be worth changing.

I'll follow up further with this later, but thanks again!

andrews05:
Okay, I've pushed some additional changes:
1. In `ezeng_reindex` I've pulled a call to `matrix.row()` outside of a pair of loops. This was me not thinking straight when I wrote it. One of the loops was previously `matrix[ra][rp]` and since `ra` is the loop var I couldn't pull the `matrix.row(ra)` out of the loop. But I didn't consider that since the matrix is symmetrical, I can pull `matrix.row(rp)` out instead and then index that with `ra` and still get the same value.
This gives a 10% speedup on ezeng without needing any `unsafe`.
2. I've changed `weighted_edges` from `(usize, usize)` to `(u8, u8, u32)`. The u32 was previously just included temporarily for sorting but by keeping it permanently we save the cost of having to remove it. It's still a drop from 16 bytes per element to 6.
3. Rather than switching to unstable sort, I've made the weighted edges omit zero-weight edges from the list entirely. This keeps the list very short, and therefore much faster and using less memory. This _does_ affect the battiato output but we aren't using that currently and a quick test shows the difference is largely a wash (unlike unstable sort which showed a negative impact).
The result is we shave about a third off the matrix construction time.

Results updated in first post.

Thanks to @vlakoff for prompting the investigations that led to these improvements.

vlakoff:
Nice results! Two small technical notes on points 2 and 3 from above:

**2. The `(u8, u8, u32)` tuple is actually 8 bytes, not 6**

Rust doesn't pack tuples byte-by-byte — fields are aligned like in a struct, and the `u32` needs to start on a 4-byte boundary. So the layout is 4 bytes (u32) + 2 bytes (the two u8s), then the whole thing gets padded up to a multiple of 4 → **8 bytes total**, not 6.

Still a solid win — 16 → 8 bytes is an exact 2x reduction, just not quite the ~2.7x implied by "16 → 6". Flagging the real number mostly in case it matters later (e.g. if this gets extended to `#[repr(packed)]` or a manually packed encoding to reclaim those last 2 bytes).

**3. Potential panic on `edges[0]` after dropping zero-weight edges**

Both `mzeng_reindex` and `ezeng_reindex` start with:

```rust
let mut remapping = vec![edges[0].0 as usize, edges[0].1 as usize];
```

Before this commit, `weighted_edges` always contained all `C(num_colors, 2)` pairs, so it could never be empty once `num_colors > 2` — `edges[0]` was safe by construction.

After this commit, edges with weight 0 are skipped. This can actually happen on real files: a PNG's palette can legitimately declare more entries than are ever used in the pixel data (common with images produced/edited by tools that don't trim unused palette entries, or with a palette entry used only once at the very edge/corner of the image where it has no neighbor). In that case the co-occurrence matrix can end up entirely zero off-diagonal, `weighted_edges` ends up empty, and `edges[0]` panics with an out-of-bounds index — on otherwise valid, unremarkable input.

It's not a memory-safety issue — Rust's bounds checking still holds, so there's no UB or corruption — but it's still a panic that would crash oxipng on that file. Given this is a general-purpose PNG optimizer that's expected to run on arbitrary user-supplied images, this seems worth guarding against explicitly (early return / fallback when `weighted_edges.is_empty()`), ideally with a regression test using a crafted palette that triggers it.

andrews05:
2. True. The big win here actually comes from excluding the zero-weight edges. In a test image with 256 colours, the list dropped from 32k entries to 1355 entries. At 8 bytes each, that's now only 11KB of memory.
3. Also true in isolation. And if only one colour is actually present in the image, the entire exercise would be a complete waste of time. Which is why we rely on `reduced_palette()` having been run prior to running any of our sort algorithms. (Note this _cannot_ occur with a palette entry only being used in the corner - the corner does still have neighbours)

vlakoff:
Regarding the pixel adjacency logic: because `prev_val` isn’t reset between rows, the algorithm effectively treats the entire image as a single, continuous 1D sequence. This means any non-monochrome image is mathematically guaranteed to produce at least one valid transition, which eliminates the "isolated corner pixel" edge case.

However, the core concern regarding `edges[0]` and the risk of a panic still stands:

* **Implicit coupling:** The safety of `edges[0]` currently relies entirely on an *upstream* guarantee (`reduced_palette()` being called first in `mod.rs`). The reindexing functions themselves have no local defensive checks or typed invariants to enforce this.
* **Fragile `pub` API:** These functions are marked `pub`. Even if the production pipeline is safe today, any future refactoring, new caller, external crate usage, or even a benchmark could easily bypass `reduced_palette()` and trigger a panic without any compiler warnings.
* **Precedent for defensiveness:** The code already exhibits defensive patterns elsewhere (e.g., explicitly checking for out-of-bounds pixel indices with `if val >= num_colors { continue; }`). 

Relying on `mod.rs` staying exactly as it is is fragile. Adding a quick `is_empty()` guard, a `debug_assert!`, or at least a clear doc-comment documenting this strict precondition would make the code resilient against future refactorings at virtually no cost.

andrews05:
> Regarding the pixel adjacency logic: because `prev_val` isn’t reset between rows, the algorithm effectively treats the entire image as a single, continuous 1D sequence. This means any non-monochrome image is mathematically guaranteed to produce at least one valid transition, which eliminates the "isolated corner pixel" edge case.

Hm, it actually shouldn't be doing this. Since rows will ultimately be separated by filter bytes, the end of one row should not be considered adjacent to the start of the next. I've now changed it to not do this, but the logic still holds true: There is no such thing as an "isolated corner pixel" edge case, it is logically impossible.

Changes:
- `prev_val` is now reset between rows since the previous behaviour was not really appropriate. (Note this can affect the results).
- The existing defensive patterns of `if val >= num_colors { continue; }` have been removed as they are unnecessary. We get another small performance win from this (~5%).
- Added a doc comment mentioning the precondition in `CoOccurrenceMatrix::from`.

Regarding the pub API, these are only marked as pub for our internal tests. They are not actually considered public and are not present in our API docs.

andrews05:
Further update: I've added support for interlaced images in the matrix, so they can now get some big wins from advanced palette sorting too. To do this we just skip vertical adjacency between passes. A consequence of this is that an unusually constructed image could theoretically have each colour isolated on separate interlacing passes and therefore have no "transitions" and an empty edge list. So I've now added a check for such an edge case 🙂 (which, to be clear, was not previously possible).

andrews05:
Gah, this interlacing support and colour isolation issue has been bugging me and I've going back and forth on it for a while now. But I've finally decided to change to a simpler implementation that just avoids any possibility of isolation (however unlikely it may be).
Interlacing isn't exactly a priority anyway, so whether this approach is "optimal" or not isn't as important as ensuring there's no possibility of failure.

vlakoff:
By dropping the `pass == line.pass` check in `build()`, we now introduce vertical cross-pass adjacency noise into *every* interlaced image's matrix, not just the pathological ones. Since filtering doesn't apply across passes, these "adjacent" bytes aren't actually spatially adjacent in the final image, so this is real noise in the sort input for all interlaced palette images.

An alternative that avoids this entirely: keep `build()` pass-aware (i.e. the original `prev_line.filter(|l| l.pass == line.pass)` version), and instead handle the empty-edges case locally in `mzeng_reindex`/`ezeng_reindex`/`battiato_reindex` by falling back to an identity mapping (`(0..num_colors).collect()`) when `weighted_edges` is empty. That way:

- The 99.99% normal case gets the fully accurate matrix, with zero cross-pass noise.
- The pathological case (every color isolated within its own pass) just means "no advanced sort available," and skips straight to the existing order — no panic, no `None` bubbling up and discarding the whole palette-reduction attempt for that image.
- It's directly testable: construct a `CoOccurrenceMatrix` with empty `weighted_edges` and assert each reindex function returns the identity.

The one advantage of the current approach is a single point of guarantee (`build()`) instead of three small guards.

On that note, `mzeng_reindex` and `ezeng_reindex` would factor together fairly easily — this is the duplication I flagged [earlier](https://github.com/oxipng/oxipng/pull/831#issuecomment-4862354780) (point 4, the `greedy_chain_build` idea) — and a shared `is_empty()` guard would fit naturally there. `battiato_reindex` would stay separate, so it'd need its own guard rather than sharing one via that refactor. Either way, treating an empty `weighted_edges` as "no optimized sort available, fall back to identity" seems like a sound approach.

andrews05:
Yes, this is noted in my comments in the code. The "noise" really isn't a concern - we don't want to introduce extra hoops to jump through just for interlaced images since interlacing is suboptimal in the first place.
(Also, an identity mapping would not be useful, we would just return None if we weren't going to change anything.)

TPS:
_(As a side-note, I got excited when practically-deprecated-in-use interlacing support got improved here & then a bit sad that it's not going to make it in, after all.  I've no personal interest in it, & I don't even see a practical general use for it much nowadays, but it does seem like a feature_ **someone** _wants/needs that just gets no love in this optimization domain.)_

vlakoff:
Fair point. It makes sense that the "cross‑pass noise" shouldn't be concerning—though perhaps for an even stronger reason:

One detail that wasn't explicitly mentioned is that this cross‑pass boundary condition only occurs exactly 6 times per interlaced image (between the 7 Adam7 passes). On any reasonably sized image with hundreds or thousands of scanlines, adding just 6 artificial vertical adjacencies to the matrix is very minor and shouldn't noticeably affect the sorting results.

By accepting this minor noise, we avoid complexity in two places: we don't add a pass check in `build()`, thus ensuring the edge list is never empty, which removes the need for safety checks in the three reindex functions. Given this, keeping `build()` simpler seems reasonable.

(That said, I would still personally favor exactness over introducing data that is erroneous by definition, even if the real‑world impact is low and it costs a few extra lines of code.)

ace-dent:
My 2¢ -  we should remove _all_ support for interlaced PNG images. By design, it is not fit for purpose. It always creates larger images and slows file transfer. There is no place for it when a User seeks optimal compression. Happy to hear an alternative perspective.

andrews05:
> _(As a side-note, I got excited when practically-deprecated-in-use interlacing support got improved here & then a bit sad that it's not going to make it in, after all.

@TPS Oh no, we're still getting support here. Previously we only supported the basic luma sort for interlaced images. We're now adding support for the heuristic algorithms (ezeng) too.

> (That said, I would still personally favor exactness over introducing data that is erroneous by definition, even if the real‑world impact is low and it costs a few extra lines of code.)

@vlakoff I absolutely get this, which is why I wrestled with it for a while before making a call. But there are two additional points in favour of this approach: 1. It improves accuracy of the popular colour calculation (though this is incredibly minor and also just qualifies as "noise"). 2. Some quick testing actually showed an average improvement vs pass-boundary-skipping. This is not definitive of course, but does indicate that the erroneous-vs-accuracy argument doesn't necessarily stack up in practice.

> My 2¢ - we should remove _all_ support for interlaced PNG images.

@ace-dent The thing is it's really not hurting anything either. Supporting it here (in the manner I've chosen) was a trivial matter of changing one line of code. It honestly felt a bit silly _not_ supporting it in the first place.
Removing support for interlaced _output_ could save some code by dropping the function to convert to interlaced, but that's about all. Support for interlaced _input_ would obviously still be required. So there's no compelling reason to drop it at this point. (Also, we did have a report of it being used in #666)

ace-dent:
@andrews05 - fair. My concern was LOC / legacy maintenance cost. The User is using interlacing as an aesthetic choice, which actually seems a reasonable use case. The issue number does still suggest that Interlaced PNG is a little evil... `#666`... 🫣

AlexTMjugador (approved):
LGTM!

PR #837, opened 2026-07-10
Make `keep` more robust

### LINKED ISSUE #836: Fully-opaque palette tRNS is unconditionally dropped — not preserved by --keep tRNS, --strip none, --nc, or -o 0
## Summary

  For a paletted (indexed-color) PNG whose `tRNS` chunk exists but is **fully opaque** (every entry `== 255`), oxipng **always removes the `tRNS` chunk**, and there is **no flag combination that prevents
  it** — including `--keep tRNS`, `--strip none`, `--nc`, `--np`, `--nx`, `--nz`, and `-o 0`.

  This is surprising for two reasons:

  1. `--keep tRNS` and `--strip none` explicitly ask to keep the chunk, but it is silently dropped anyway.
  2. For true-color RGBA images, the corresponding alpha channel *can* be preserved with `--nc`; there is no equivalent escape hatch for the palette case, so the behavior is inconsistent between color
  types.

  The removal is pixel-lossless (an all-255 tRNS carries no visible information), so I understand *why* it's considered redundant. The problem is that some downstream consumers select a texture format
  from the PNG's **declared channel count / presence of an alpha chunk**, not from the sampled pixel values. When the `tRNS`/alpha chunk is stripped, such a consumer treats the image as having no alpha
  channel at all, and a shader sampling `.a` then reads the sampler default instead of the file's channel. In that context "losslessly redundant" and "safe to drop" are not the same thing, and an opt-out
  is needed.

  ## Environment

  - oxipng 10.1.1 (prebuilt `oxipng-10.1.1-aarch64-apple-darwin`, also reproduced against the 10.1.1 source tree)
  - macOS 15.6.1, aarch64 (Apple Silicon)

  ## Reproduction

  Create a paletted PNG that carries a `tRNS` chunk whose entries are all `255` (fully opaque), then run oxipng with any of the flag sets below.

  ```python
  # make a P-mode PNG with an all-255 tRNS chunk (Pillow)
  import numpy as np
  from PIL import Image
  rng = np.random.default_rng(42)
  rgb = rng.integers(0, 256, (32, 32, 3), dtype=np.uint8)
  im = Image.fromarray(rgb, "RGB").convert("P", palette=Image.ADAPTIVE, colors=64)
  im.info["transparency"] = bytes([255] * (len(im.getpalette()) // 3))
  im.save("palette_opaque_trns.png")   # mode=P, has tRNS
  ```

  ```bash
  for FLAGS in \
    "-o max" \
    "-o max --nc" \
    "-o max --nc --np --ng --nb" \
    "-o max --keep tRNS" \
    "-o max --nx" \
    "-o max --nx --nz" \
    "-o 0 --nx --nz --keep tRNS" ; do
    cp palette_opaque_trns.png t.png
    oxipng $FLAGS t.png >/dev/null 2>&1
    python3 -c "from PIL import Image;i=Image.open('t.png');print(repr('$FLAGS'), i.mode, 'tRNS' in i.info)"
  done
  ```

  ## Expected

  At least one flag — ideally `--keep tRNS` or `--strip none`, or a color-type lock like `--nc` — preserves the `tRNS` chunk on output.

  ## Actual

  Every combination drops it:

  ```
  '-o max'                       P False
  '-o max --nc'                  P False
  '-o max --nc --np --ng --nb'   P False
  '-o max --keep tRNS'           P False
  '-o max --nx'                  P False
  '-o max --nx --nz'             P False
  '-o 0 --nx --nz --keep tRNS'   P False
  ```

  For contrast, a **true-color RGBA** image with an all-255 alpha channel *can* be preserved:

  ```
  '-o max'         RGBA -> RGB   (alpha dropped)
  '-o max --nc'    RGBA -> RGBA  (alpha preserved)   <-- works for RGBA, no palette equivalent
  ```

  The only workaround I found for palette images is to pre-convert `P -> RGBA` before running oxipng and then pass `--nc`, which forfeits the palette's size advantage.

  ## Why flags don't help (source pointers, v10.1.1)

  The drop doesn't happen in a skippable reduction/transformation or in the `--strip`/`--keep` metadata path — it happens during palette **serialization**, so the strip-related flags never get a say:

  - On load, `tRNS` is merged into the palette entries' `.a` and stops existing as an independent chunk — `src/headers.rs:222-234` (`palette_to_rgba`, `RGBA8::new(..., 255)` then `color.a = trns`).
  - On output, the `tRNS` chunk is **rebuilt from scratch** and only written if some entry is non-opaque — `src/png/mod.rs:196`:

    ```rust
    if let Some(last_trns) = palette.iter().rposition(|px| px.a != 255) {
        let trns_data: Vec<_> = palette[0..=last_trns].iter().map(|px| px.a).collect();
        write_png_block(b"tRNS", &trns_data, &mut output);
    }
    ```

    If every entry is `255`, `rposition` returns `None`, the `if` is skipped, and **no `tRNS` chunk is emitted** — regardless of `--keep`, `--strip`, `--nc`, or `-o 0`. The size estimator uses the same
    rule (`src/png/mod.rs:310`), so this is a consistent internal model rather than a stray side effect.

  ## Request

  Either of:

  1. **Honor an explicit opt-out** so a fully-opaque palette `tRNS` (and, symmetrically, a fully-opaque RGBA alpha) can be preserved — e.g. make `--keep tRNS` / `--strip none` bypass the rebuild for
  paletted images, or extend `--nc` to lock the alpha-bearing channel/chunk for the palette case.
  2. **Document the behavior** clearly in the manual: that fully-opaque `tRNS`/alpha is always dropped as redundant and that `--keep tRNS`/`--strip none`/`--nc` do not preserve it for paletted images.

andrews05:
- Disallow protected chunks with `--keep` (the same set that we already do not allow with `--strip`).
- Add new `ChunkPreventsChanges` error, thrown if `iDOT` or `caBX` is explicitly kept.
  - The jumbf parsing to check for C2PA has been removed. The `caBX` chunk is specifically for storing a C2PA manifest, so there's no reason to check for this.
- Explicitly kept `bKGD`, `sBIT` or `hIST` chunks will now disable reductions rather than stripping them regardless.

All of this means there should no longer be any case where a chunk specified in `--keep` is still stripped anyway. Other strip modes may still strip these chunks, as documented in the help.

Closes #836.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -124,7 +124,7 @@ fn main() -> ExitCode {
                     num_not_optimized += 1;
                 }
             }
-            Err(PngError::C2PAMetadataPreventsChanges | PngError::InflatedDataTooLong(_)) => {}
+            Err(PngError::ChunkPreventsChanges(_) | PngError::InflatedDataTooLong(_)) => {}
             Err(_) => num_failed += 1,
         }
     }
@@ -543,7 +543,7 @@ fn process_file(input: &InFile, output: &OutFile, opts: &Options) -> Optimizatio
     let result = oxipng::optimize(input, output, opts);
     match &result {
         Ok(_) => {}
-        Err(e @ PngError::C2PAMetadataPreventsChanges | e @ PngError::InflatedDataTooLong(_)) => {
+        Err(e @ PngError::ChunkPreventsChanges(_) | e @ PngError::InflatedDataTooLong(_)) => {
             warn!("{input}: Skipped: {e}");
         }
         Err(e) => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -391,7 +391,7 @@ fn parse_opts_into_struct(
                 }
                 Some(match parse_chunk_name(x) {
                     Ok(name) if FORBIDDEN_CHUNKS.contains(&name) => Err(format!(
-                        "{x} chunk is controlled internally may not be explicitly kept"
+                        "{x} chunk is controlled internally and may not be explicitly kept"
                     )),
                     name => name,
                 })
--- a/src/main.rs
+++ b/src/main.rs
@@ -379,16 +379,22 @@ fn parse_opts_into_struct(
         };
     }
 
+    const FORBIDDEN_CHUNKS: [[u8; 4]; 5] = [*b"IHDR", *b"IDAT", *b"tRNS", *b"PLTE", *b"IEND"];
     if let Some(keep) = matches.get_one::<String>("keep") {
         let mut keep_display = false;
         let mut names = keep
             .split(',')
-            .filter_map(|name| {
-                if name == "display" {
+            .filter_map(|x| {
+                if x == "display" {
                     keep_display = true;
                     return None;
                 }
-                Some(parse_chunk_name(name))
+                Some(match parse_chunk_name(x) {
+                    Ok(name) if FORBIDDEN_CHUNKS.contains(&name) => Err(format!(
+                        "{x} chunk is controlled internally may not be explicitly kept"
+                    )),
+                    name => name,
+                })
             })
             .collect::<Result<IndexSet<_>, _>>()?;
         if keep_display {
@@ -403,8 +409,6 @@ fn parse_opts_into_struct(
         } else if strip == "all" {
             opts.strip = StripChunks::All;
         } else {
-            const FORBIDDEN_CHUNKS: [[u8; 4]; 5] =
-                [*b"IHDR", *b"IDAT", *b"tRNS", *b"PLTE", *b"IEND"];
             let names = strip
                 .split(',')
                 .map(|x| {
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -133,9 +133,10 @@ Strip metadata chunks, where <mode> is one of:
 
 CAUTION: 'all' will convert APNGs to standard PNGs.
 
-Please note that regardless of any options set, some chunks will necessarily be stripped \
-when invalidated by the optimization:
-    bKGD, sBIT, hIST: Stripped if the color type or bit depth changes.
+Please note that some chunks will necessarily be stripped when invalidated by the \
+optimization:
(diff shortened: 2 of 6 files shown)


ace-dent:
@andrews05 - do we need to add `iDOT` ? Are there other PNG v3 chunks to consider? 🤔

andrews05:
IDOT falls under a different category, more similar to c2pa. We don’t currently treat it the same way as c2pa (and perhaps we should) but we do notify when it is stripped.

TPS:
@andrews05 What about the rest @charlietsao [mentioned](https://github.com/oxipng/oxipng/issues/836#issue-4845091980)?
> --strip none, --nc, --np, --nx, --nz, and -o 0

andrews05:
@TPS No, there are no changes to those flags. (strip none is not a proper option, it will just strip chunks of type `none`)

@ace-dent I’ve made further changes for iDOT and others.

src/main.rs
```diff
@@ -379,16 +379,22 @@ fn parse_opts_into_struct(
...
-                if name == "display" {
+            .filter_map(|x| {
+                if x == "display" {
                     keep_display = true;
                     return None;
                 }
-                Some(parse_chunk_name(name))
+                Some(match parse_chunk_name(x) {
+                    Ok(name) if FORBIDDEN_CHUNKS.contains(&name) => Err(format!(
+                        "{x} chunk is controlled internally may not be explicitly kept"
```
AlexTMjugador: I think there is a typo here.
```suggestion
                        "{x} chunk is controlled internally and may not be explicitly kept"
```

AlexTMjugador (commented):
Just a little nit, other than that this change sounds quite reasonable to me 👍

src/main.rs
andrews05: Good catch, thanks!

AlexTMjugador (approved):

PR #838, opened 2026-07-15
Balance adjustments to palette sorting

andrews05:
One more sorting PR before the next release!
- Reinstate Battiato method at o4 and higher. A few factors have allowed this back again:
  - Sorting performance improvements from #820, including now re-using the same matrix for multiple algorithms.
  - A kludgy way to infer the o-level from known options.
  - A tweak to the popular colour rule to get slightly more benefit from battiato.
- Enable ezeng at o2 (usually instead of luma). At this level we limit swap distance to 1 for performance.

I've also bumped MSRV to 1.88 to allow let-chaining, which can help to reduce nesting in some places.

| Level | Saved | Time |
|--|--|--|
| o2 master | 14.64% | 3.1s |
| o2 PR | 16.50% | 3.3s |
| o4 master | 17.20% | 16.3s |
| o4 PR | 17.24% | 16.7s |

### DIFF
--- a/.github/workflows/oxipng.yml
+++ b/.github/workflows/oxipng.yml
@@ -167,7 +167,7 @@ jobs:
       - name: Install MSRV Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: 1.85.1
+          toolchain: 1.88.0
           cache-bin: false
           cache-shared-key: cache
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ license = "MIT"
 name = "oxipng"
 repository = "https://github.com/oxipng/oxipng"
 version = "10.1.1"
-rust-version = "1.85.1"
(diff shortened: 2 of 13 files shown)


ace-dent:
More great stuff @andrews05 🙌
Just checking if this displaces Luma ?

TPS:
> Just checking if this displaces Luma ?

Hopefully not, since [it still has its place](https://github.com/oxipng/oxipng/issues/553#issuecomment-4950054086).

andrews05:
At o2 ezeng now (mostly) displaces luma. o0-1 only use luma, and it returns at o3+ as well.
The luma transform is of course very fast and we actually still perform the transform at o2 (because it helps with determinism in ezeng), we just don't evaluate it because the gains aren't worth the evaluation cost at this level.

andrews05:
I want a little more time to work on the level balance here, I'm going to convert to draft and open a different PR in the meantime...

andrews05:
Closing in favour of #842

PR #840, opened 2026-07-20
More transformation performance improvements

andrews05:
Improve performance of some remaining lagging transformations. Unless you're enabling interlacing though, this is unlikely to translate into measurable improvements in practice.

For interlacing, we switch to bytes-only implementations with generic specialisations for different pixel sizes. Low-depth images will be expanded and reduced again if necessary, which turns out to still be much faster than the old bitvec implementation. The side benefit here is we can drop the bitvec dependency.

Master:
```
test deinterlacing_16_bits ... bench:   1,203,302.10 ns/iter (+/- 5,250.60)
test deinterlacing_1_bits  ... bench:     473,486.21 ns/iter (+/- 1,664.17)
test deinterlacing_2_bits  ... bench:     935,508.30 ns/iter (+/- 7,651.25)
test deinterlacing_4_bits  ... bench:   1,865,702.10 ns/iter (+/- 14,821.83)
test deinterlacing_8_bits  ... bench:     607,866.60 ns/iter (+/- 7,101.60)
test interlacing_16_bits   ... bench:  36,653,241.60 ns/iter (+/- 271,127.53)
test interlacing_1_bits    ... bench:     549,333.30 ns/iter (+/- 10,544.48)
test interlacing_2_bits    ... bench:   1,075,779.20 ns/iter (+/- 12,815.81)
test interlacing_4_bits    ... bench:   2,130,787.55 ns/iter (+/- 7,600.67)
test interlacing_8_bits    ... bench:  18,761,558.30 ns/iter (+/- 52,728.81)

test reductions_grayscale_8_to_palette_8            ... bench:     361,885.40 ns/iter (+/- 4,224.42)
test reductions_grayscale_alpha_to_palette_8        ... bench:     475,522.90 ns/iter (+/- 3,975.64)
test reductions_rgb_to_palette_8                    ... bench:     638,195.80 ns/iter (+/- 28,795.27)
test reductions_rgba_to_palette_8                   ... bench:     857,379.20 ns/iter (+/- 7,470.73)

test reductions_grayscale_alpha_to_grayscale_16     ... bench:     923,999.90 ns/iter (+/- 34,174.59)
test reductions_grayscale_alpha_to_grayscale_8      ... bench:     714,987.50 ns/iter (+/- 7,377.88)
test reductions_grayscale_alpha_to_grayscale_trns_8 ... bench:     714,708.40 ns/iter (+/- 10,525.79)
test reductions_rgba_to_rgb_16                      ... bench:   1,399,633.30 ns/iter (+/- 8,920.09)
test reductions_rgba_to_rgb_8                       ... bench:   1,300,487.50 ns/iter (+/- 16,369.16)
test reductions_rgba_to_rgb_trns_8                  ... bench:   1,162,769.80 ns/iter (+/- 4,642.59)

test filters_bigent  ... bench:  10,869,329.20 ns/iter (+/- 154,144.48)
```

PR:
```
test deinterlacing_16_bits ... bench:     124,966.36 ns/iter (+/- 1,009.94)
test deinterlacing_1_bits  ... bench:     307,630.91 ns/iter (+/- 6,463.10)
test deinterlacing_2_bits  ... bench:     433,753.65 ns/iter (+/- 5,293.89)
test deinterlacing_4_bits  ... bench:     466,835.40 ns/iter (+/- 22,574.88)
test deinterlacing_8_bits  ... bench:     120,005.74 ns/iter (+/- 1,756.67)
test interlacing_16_bits   ... bench:     408,762.50 ns/iter (+/- 5,364.38)
test interlacing_1_bits    ... bench:     524,241.60 ns/iter (+/- 7,712.99)
test interlacing_2_bits    ... bench:     657,570.80 ns/iter (+/- 4,787.53)
test interlacing_4_bits    ... bench:     688,583.20 ns/iter (+/- 5,711.52)
test interlacing_8_bits    ... bench:     358,910.42 ns/iter (+/- 2,147.28)

test reductions_grayscale_8_to_palette_8            ... bench:     376,310.40 ns/iter (+/- 4,916.01)
test reductions_grayscale_alpha_to_palette_8        ... bench:     495,516.70 ns/iter (+/- 6,881.72)
test reductions_rgb_to_palette_8                    ... bench:     600,275.00 ns/iter (+/- 5,767.45)
test reductions_rgba_to_palette_8                   ... bench:     715,733.30 ns/iter (+/- 7,340.38)

test reductions_grayscale_alpha_to_grayscale_16     ... bench:     219,934.40 ns/iter (+/- 2,372.37)
test reductions_grayscale_alpha_to_grayscale_8      ... bench:     485,862.55 ns/iter (+/- 3,654.26)
test reductions_grayscale_alpha_to_grayscale_trns_8 ... bench:     445,089.60 ns/iter (+/- 4,654.34)
test reductions_rgba_to_rgb_16                      ... bench:     236,690.62 ns/iter (+/- 2,453.35)
test reductions_rgba_to_rgb_8                       ... bench:     253,652.77 ns/iter (+/- 6,025.82)
test reductions_rgba_to_rgb_trns_8                  ... bench:     284,055.57 ns/iter (+/- 6,838.74)

test filters_bigent  ... bench:   6,781,095.80 ns/iter (+/- 163,795.62)
```


andrews05:
I pushed another change to improve BigEnt performance. This one is a bit of a toss-up. On the one hand, it requires 256k of memory while being unlikely to tangibly improve performance in actual usage.  On the other hand, oxipng is already memory-intensive (the co-occurrence matrix can also require up to 256k memory) and it was hard to prove that peak memory actually increased when running at o3 here. I decided the performance win is probably the better option.

TPS:
> I pushed another change to improve BigEnt performance. This one is a bit of a toss-up. On the one hand, it requires 256k of memory while being unlikely to tangibly improve performance in actual usage.

@Andrews05  Is this what you referred to the end of [this comment](https://github.com/oxipng/oxipng/pull/804#issuecomment-4147345229)?

andrews05:
Yup, that's the one.

src/filters/strategies.rs
```diff
@@ -193,13 +192,15 @@ impl StrategyEvaluator for BigramsEvaluator {
 
 // Bigram entropy, combined from Entropy and Bigrams filters
 struct BigEntEvaluator {
-    counts: FxHashMap<u16, u32>,
+    seen: [u32; 0x10000],
```
AlexTMjugador: Even though `BigEntEvaluator` instances are ultimately boxed during construction in `FilterStrategy#evaluator()`, [Rust's semantics do not guarantee that the compiler will inline the calls to elide stack initialization and construct them directly in a heap-allocated `Box`](https://users.rust-lang.org/t/allocating-array-on-heap-using-box-new-results-in-stack-overflow/129264).

Since `BigEntEvaluator` now contains an array that is at least 256 KiB in size, this could become problematic: [all current Tier 1 platforms use a default stack size of 2 MiB for Rust-spawned threads](https://doc.rust-lang.org/std/thread/index.html#stack-size), while the main thread's stack size depends on the operating system and, on Windows, [it can regularly be as small as 1 MiB](https://learn.microsoft.com/en-us/windows/win32/procthread/thread-stack-size). To avoid unnecessarily increasing the risk of stack overflows for applications using Oxipng as a library, I think we should heap-allocate this array instead.

For that purpose, I'd suggest using `vec![0; 0x10000].into()`, which [Rust has optimized very well for some time](https://godbolt.org/z/xbYKT7nxW). Another valid option is `std::iter::repeat_n(0, 0x10000).collect()`, which generates similar machine code and conceptually avoids the intermediate `Vec`, though I personally find it a bit more verbose.

src/reduction/interlace.rs
```diff
@@ -0,0 +1,187 @@
...
+        4 => deinterlace_bytes_const::<4>(png),
+        6 => deinterlace_bytes_const::<6>(png),
+        8 => deinterlace_bytes_const::<8>(png),
+        _ => unreachable!(),
+    }
+}
+
+// Delegate function with const generics for performance
+fn deinterlace_bytes_const<const BPP: usize>(png: &PngImage) -> Vec<u8> {
+    let bytes_per_pixel = BPP;
```
AlexTMjugador: Here we can use the `BPP` const generic directly, which is slightly more concise. Is there a particular reason to introduce a separate binding to rename it to `bytes_per_pixel`?

Personally, I think the meaning of `BPP` is well established enough that the renaming isn't necessary. If the concern is that the "B" could be interpreted as either bytes or bits, we could instead clarify that in the comment above the function.

AlexTMjugador (commented):
Looks good to me! Just a couple comments 👍

.github/workflows/oxipng.yml
```diff
@@ -167,7 +167,7 @@ jobs:
       - name: Install MSRV Rust toolchain
         uses: actions-rust-lang/setup-rust-toolchain@v1
         with:
-          toolchain: 1.85.1
+          toolchain: 1.88.0
```
AlexTMjugador, line 170: When preparing the next release, I think we should document this MSRV bump somewhere. I assume you did it due to the usage of newer Rust features in the source code, right?

src/filters/strategies.rs
andrews05: Hm, interesting point. Just so I'm clear, you're suggesting changing `seen: [u32; 0x10000]` to `seen: Box<[u32]>`, and then initialising it with vec into?

src/reduction/interlace.rs
andrews05: Er, it may have just been an artefact of refactors. I've removed this now.

.github/workflows/oxipng.yml
andrews05: Yeah, 1.88 allows `as_chunks` and if-let chaining. Will definitely mention it in the release notes.

src/filters/strategies.rs
AlexTMjugador: Yeah, exactly that :slightly_smiling_face:

.github/workflows/oxipng.yml
AlexTMjugador: Sounds good!

src/filters/strategies.rs
andrews05: Sounds good, done!

AlexTMjugador (approved):

PR #842, opened 2026-07-31
Rebalance

### LINKED ISSUE #755: Override default of `fast` mode
Following #753 I discovered (after a few years use!) that `--fast` is enabled by default. I admit this is a human skill issue(!)... but... 🙃

**Issue**: an 'advanced' user may use something like the following and assume `fast` is not in use:
```sh
oxipng --filters 0-9 --zc 12 INPUT.png
```

BUT... The default is level 2 `-o2` (so `fast` is enabled), as per the help:
```text
...

2   => --zc 11 -f 0,1,6,7 --fast

… Manually specifying a compression option (zc, f, etc.) will override the
optimization preset, regardless of the order you write the arguments. … 

...

--fast
          Perform a fast compression evaluation of each enabled filter, followed by a
          single main compression trial of the best result. Recommended if you have more
          filters enabled than CPU cores.
```

So:
1. Should we add a documentation note to `--fast`, e.g. "Enabled by default".
2. We have a way to enable `--fast` (despite being _on_ by default!?!?), but no obvious way to disable it! Either we should be able to pass a boolean  `--fast 0` or add a new command `--slow`. Currently the only way is to set `-o3` (or greater)... this is not intuitive!!

### LINKED ISSUE #775: Clarify levels 0 and 1 in the Manual

andrews05:
This is a notable rebalancing of the levels, which hasn't happened since v7. Hopefully the final PR before the 10.2 release.
I've tried to make the progression more consistent here by only enabling additional things as we go up, rather than trying micro-balance each level by potentially disabling things that were enabled at a lower level.

Summary:
- Level 0: Use filters 0,5
- Level 1: Use filters 0,7
- Level 2: Enable ezeng (with fewer swaps than normal)
- Level 3: Enable `fast`, add filters 1,6
- Level 4: Enable `fast`, add filters 1,6, enable battiato
- Level 5: Remove filters 2,5, change brute to 1/8, enable battiato
- Level 6: Enable battiato

Results:
| Level | 10.1.1 % | 10.1.1 time | PR % | PR time |
|--|--|--|--|--|
| 0 | 33.46% | 0.75s | 34.30% | 0.55s |
| 1 | 38.33% | 2.6s | 39.62% | 2.5s |
| 2 | 40.39% | 5.5s | 40.85% | 5.6s |
| 3 | 40.94% | 17.3s | 41.12% | 9.3s |
| 4 | 41.19% | 29.2s | 41.40% | 13.2s |
| 5 | 41.31% | 56s | 41.49% | 42s |
| 6 | 41.34% | 70s | 41.55% | 69s |

Closes #755 
Closes #775 
Supersedes #827
Supersedes #838


src/reduction/mod.rs
```diff
@@ -21,11 +21,17 @@ pub(crate) fn perform_reductions(
 ) -> Arc<PngImage> {
     let mut evaluation_added = false;
 
-    // At low compression levels, skip some transformations which are less likely to be effective
-    // This currently affects optimization presets 0-2
-    let cheap = match opts.deflater {
-        Deflater::Libdeflater { compression } => compression < 12 && opts.fast_evaluation,
-        _ => false,
+    // Try to infer the optimization level from the settings
+    // TODO: [v11] Store the level in the settings
```
AlexTMjugador, line 25: Just a note for our future selves: when using Oxipng through its API, the settings in use may not exactly match any of the predefined optimization levels. Because of that, I'd suggest keeping the logic here tied to the specific parameters that make transformations less effective, rather than to fixed optimization level values.

AlexTMjugador (approved):
Nice!

TPS:
@andrews05 So, `--fast` _isn't_ leaving [_(yet)_](https://github.com/oxipng/oxipng/issues/755#issuecomment-5146707827), after all, & it's turned on @ different levels than before? 😯  For me, that's the most surprising thing here.

PR #847, opened 2026-08-08
Add option for progress bar that also shows cumulative savings.

ADJB1212:
Adds --progress option for showing a progress bar. Useful when using -r to process all PNGs in a folder.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,7 @@ clap = { version = "4.6.0", optional = true, features = ["wrap_help"] }
 env_logger = { version = "0.11.10", optional = true, default-features = false, features = ["auto-color"] }
 image = { version = "0.25.9", optional = true, default-features = false, features = ["png"] }
 indexmap = "2.14.0"
+indicatif = { version = "0.17.11", optional = true, features = ["rayon"] }
 libdeflater = "1.25.2"
 log = "0.4.31"
 parse-size = { version = "1.1.0", optional = true }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -42,6 +42,7 @@ clap = { version = "4.6.0", optional = true, features = ["wrap_help"] }
 env_logger = { version = "0.11.10", optional = true, default-features = false, features = ["auto-color"] }
 image = { version = "0.25.9", optional = true, default-features = false, features = ["png"] }
 indexmap = "2.14.0"
+indicatif = { version = "0.17.11", optional = true, features = ["rayon"] }
```
AlexTMjugador: The latest version of the `indicatif` crate is 0.18.6, so we should use that unless we have a good reason not to.

(And no, the fact that your favorite LLM may have chosen this version isn't a good reason on its own 😉)

It is of note that the default `indicatif` 0.18.6 features pull somewhat heavyweight WASM bindings, so we should use `default-features` and opt-in to the `rayon` feature only instead. (I don't believe the `unicode-width` feature is necessary either.)

src/cli.rs
```diff
@@ -234,6 +234,21 @@ losslessly.")
...
         )
+        .arg(
+            Arg::new("progress")
+                .help("Show a progress bar with cumulative size savings")
+                .long_help("\
+Show a progress bar while processing files, tracking how many files have completed along \
+with the cumulative reduction in file size so far.
+
+Requires an interactive terminal, and has no effect when combined with '--verbose', \
+'--quiet', or '--stdout'.")
```
AlexTMjugador: Using `--progress` in combination with any of those flags goes beyond having no effect: it's not allowed, and `clap` outputs an error. Therefore, I'd reword this documentation to reflect that.

AlexTMjugador (commented):
This PR looks alright to me as a first iteration on this feature. I have a few suggestions to make it better before merging, though.

PR #856, opened 2026-08-19
Tighten CI

XhmikosR:
Please check the individual patches.

The last patch doesn't save a lot, but doesn't look like it costs a lot of time anyway.

I haven't tested the deployment workflow since it needs a release.

Let me know if you want me to adapt any of the patches.

Closes #854, closes #855 along the way.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,9 +10,7 @@ concurrency:
   group: release-${{ github.ref }}
   cancel-in-progress: true
 
-permissions:
-  actions: read
-  contents: write
+permissions: {}
 
 jobs:
   deploy:
@@ -23,6 +21,10 @@ jobs:
 
     # Prevent job from running on forks
     if: ${{ !github.event.repository.fork }}
+
+    permissions:
+      actions: read # Download the CI run artifacts
+      contents: write # Create the release
 
     strategy:
       # Execute one job at a time to avoid potential race conditions in the
@@ -42,18 +44,20 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
 
       - name: Install Rust toolchain
-        uses: actions-rust-lang/setup-rust-toolchain@v1
+        uses: actions-rust-lang/setup-rust-toolchain@166cdcfd11aee3cb47222f9ddb555ce30ddb9659 # v1.17.0
         with:
           toolchain: nightly
-          cache-bin: false
-          cache-shared-key: cache
+          # Nothing is compiled here, so the cache is dead weight and a poisoning vector
+          cache: false
 
       - name: Install cargo-deb
         if: endsWith(matrix.target, '-linux-gnu')
-        uses: taiki-e/install-action@v2.85.10
+        uses: taiki-e/install-action@82cd3e7658a6f96c86c0234aeeda1748937cb0a1 # v2.85.13
         with:
           tool: cargo-deb
 
@@ -64,7 +68,7 @@ jobs:
           jq -r '.packages[] | select(.name == "oxipng").version')" >> "$GITHUB_OUTPUT"
 
       - name: Retrieve ${{ matrix.target }} binary
-        uses: dawidd6/action-download-artifact@v21
+        uses: dawidd6/action-download-artifact@b6e2e70617bc3265edd6dab6c906732b2f1ae151 # v21
         with:
           workflow: oxipng.yml
           commit: ${{ github.sha }}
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yml
```diff
@@ -42,18 +44,20 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
AlexTMjugador, line 47: Pinning action versions to exact commits is great for security and reproducibility, but if it comes at the cost of a lack of automation and policy support from tools like Dependabot to bump those commits as updates are released, I can foresee this change backfiring over time: the pinned commit could gradually become stale because, if everything continues to work, nobody has much incentive to take on the chore of updating a somewhat opaque, error prone, long hex string manually.

I see you tweaked the Dependabot config, but as far as I can tell, it doesn't change how Dependabot handles version pinning here. Do you happen to know whether Dependabot supports this, or whether we should switch to [Renovate](https://mend.io/renovate) instead? I know for a fact that Renovate does, and I'm happy about how it works for other projects, but as Oxipng adopted Dependabot instead while not belonging to an Oxipng organization, I've never really given much consideration to making the switch happen.

On the other hand, if we're going all-in on pinning actions to commits, I don't think we should skip `taiki-e/install-action` due to it having immutable releases, since [immutable releases can be disabled for a repository](https://docs.github.com/en/rest/repos/repos?apiVersion=2026-03-10#disable-immutable-releases), and an attacker could potentially exploit this by publishing an innocuous release tag and then replacing it with a malicious one.

.github/workflows/deploy.yml
```diff
@@ -76,17 +80,20 @@ jobs:
 
       - name: Build archives
         working-directory: target
+        env:
+          VERSION: ${{ steps.oxipngMeta.outputs.version }}
+          TARGET: ${{ matrix.target }}
         run: |
-          ARCHIVE_NAME="oxipng-${{ steps.oxipngMeta.outputs.version }}-${{ matrix.target }}"
+          ARCHIVE_NAME="oxipng-$VERSION-$TARGET"
```
AlexTMjugador, line 87: The values of `matrix.target` and `steps.oxipngMeta.outputs.version` are either part of the CI workflow itself or a valid semver string respectively, so I see no risk of injection attacks here.

I suppose this was was flagged by Zizmor as a potentially dangerous thing to do? If so, I personally find the previous syntax easier to read, because you don't need to think what values are `VERSION` and `TARGET` set to. But this new syntax still beats adding specific Zizmor exceptions or configuration tweaks to me.

.github/workflows/deploy.yml
```diff
@@ -119,20 +126,27 @@ jobs:
...
+        run: |
+          # The first matrix job creates the release, the rest just add assets
+          if ! gh release view "$GITHUB_REF_NAME" > /dev/null 2>&1; then
+            gh release create "$GITHUB_REF_NAME" --verify-tag \
+              --title "v$VERSION" --notes-file RELEASE_NOTES.txt
+          fi
+
+          shopt -s nullglob
+          gh release upload "$GITHUB_REF_NAME" --clobber \
+            target/*.zip target/*.tar.gz "target/$TARGET"/debian/*.deb
```
AlexTMjugador, line 152: `softprops/action-gh-release` has been quite an okay action to use for the purpose of making a release. Are you removing it because you think it's unnecessary given that `gh release` commands can do its job instead?

.github/workflows/deploy.yml
XhmikosR: Dependabot already handles these natively. That's why I also moved the checks to monthly, to reduce noise on a weekly basis.

.github/workflows/deploy.yml
XhmikosR: Yeah, it's Zizmor and to be honest, you really want these issues flagged and not having to worry about them manually. Because it's easy to miss them and get actions pwned.

.github/workflows/deploy.yml
XhmikosR: I'm removing it because of these:

1. zizmor catches that this can be handled with native `gh`
2. `gh` can indeed handle these and you gain one less third-party action to worry about

.github/workflows/deploy.yml
AlexTMjugador: I see, that's good by me then :+1:

Still, let's pin `taiki-e/install-action` to specific commit hashes, since immutable releases are not a bulletproof solution for all sorts of supply chain attacks I can imagine.

.github/workflows/deploy.yml
AlexTMjugador: Fair enough, let's go with this then :+1:

.github/workflows/deploy.yml
AlexTMjugador: It's quite nice that Zizmor recognizes the usage of "unnecessary" actions! Let's go with this too then.

AlexTMjugador (approved):
LGTM, thanks for the PR! 🚀

